PR #127, opened 2022-05-29
feat: Add version subcommand that can tell the current build version

hirosassa:
I added `version` subcommand to print version on stdout.


yohamta:
Awesome! Thank you so much @hirosassa

hirosassa:
Thanks for the rapid response!

PR #172, opened 2022-06-14
feat: Add Dockerfile

lat0z:
To interact with the project I have create an minimal Dockerfile.
There are few options I would like to know if are considered good enough: 

- Is alpine good enough as a starting point or would be considered too minimal?
- So far it relies on the already released versions, if it is needed it can be changed to build and use the main branch content as is.
@yohamta

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,48 @@
+# syntax=docker/dockerfile:1.4
+FROM alpine:latest
+
+ARG TARGETARCH
+ARG VERSION=1.3.15 
+ARG RELEASES_URL="https://github.com/yohamta/dagu/releases"
+
+ARG USER="dagu"
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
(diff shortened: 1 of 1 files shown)


yohamta:
Thanks for the contribution!

> - Is alpine good enough as a starting point or would be considered too minimal?

I am not so sure but I think it's good enough as a starting point :)

> - So far it relies on the already released versions, if it is needed it can be changed to build and use the main branch content as is.

I think already released version is better so no need to change :)

yohamta:
@lat0z Thanks for the contribution!

> - Is alpine good enough as a starting point or would be considered too minimal?

I am not so sure but I think it's good enough as a starting point :)

> - So far it relies on the already released versions, if it is needed it can be changed to build and use the main branch content as is.

I think already released version is better so no need to change :)

yohamta:
I have confirmed that it works perfectly in my local environment 👍 Let me merge this.

PR #204, opened 2022-07-22
update serve static file and fix makefile

Arvintian:


yohamta:
@Arvintian Amazing improvement! Thanks!

PR #218, opened 2022-07-29
Fix command params issue

Arvintian:
- Use go-shellwords fix command params issue.
- Allow set web page title.


yohamta:
@Arvintian Looks amazing. Thanks!

PR #227, opened 2022-08-02
Add executor interface

Arvintian:
Let DAG‘s step do everyting!


yohamta:
@Arvintian Fantastic work. Thanks!

PR #273, opened 2022-08-16
Fix admin-web bundle.js issue

### LINKED ISSUE #272: Reduce the size of admin-web bundle js
Should fix the warning from webpack:
```
WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (6.45 MiB)
      bundle.js
```

yohamta:
Closes #272


Arvintian:
About react lazy load?

yohamta:
Thanks for the suggestion! Sounds like a good option.

I took a profile and it seems like `monaco-editor`, `mermaid` and `fontawesome` are also taking a lot of space as well as `react`. I will try to separate these from the bundle.js.
image

PR #282, opened 2022-08-18
update default dag log dir

Arvintian:
Default DAG's Log Dir
Current: `${DAG_HOME}/logs/${name}/${name}`
Update: `${DAG_HOME}/logs/dags/${name}`

Let DAG's default log dir keep consistent with admin's: `${DAG_HOME}/logs/admin`


yohamta:
Looks good to me. Thanks!

PR #317, opened 2022-09-06
fix start dag with params

Arvintian:
https://github.com/yohamta/dagu/issues/279#issuecomment-1237603177
Like shellworlds pasre string be quoted as raw-string.
```
dagu start --params "a b" demo.yaml

expect:
$1: a
$2: b

but:
$1: 'a b'
```
So we should trim the `"`.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"os"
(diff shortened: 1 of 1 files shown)


yohamta:
Thanks for fixing the bug!

Issue #307, opened 2022-09-01
Ability to set dependencies between steps with the click of a graph


yohamta:

yohamta:
This library looks better than Mermaid for the purpose.
[https://github.com/reaviz/reaflow](https://github.com/reaviz/reaflow)

PR #332, opened 2022-09-13
Migrate graph library to reaflow

yohamta:
Relate to #307


yohamta:
Reaflow turns out to be not very good UX so won't migrate for now.

PR #336, opened 2022-09-17
Fix/dag table a11y issues

RamonEspinosa:
# Description
This PR solves some accessibility issues in the DAG Table. 

## The problem
I ran an audit using [Wave](https://wave.webaim.org/) and found that there where some elements that had no text nor labels assigned. This is a problem for users who have low visibility issues and prefer to use screen readers because, provided they stumble upon one of these elements, the screen reader will skip the element, resulting in a bad user experience.

## The solution
1. Added `aria-label` attributes to input elements that had no labels.
2. Added a `VisuallyHidden` component to render text for screen readers.

### Before 
Screen Shot 2022-09-17 at 11 06 06-fullpage
Screen Shot 2022-09-17 at 11 04 11-fullpage
### After
Screen Shot 2022-09-17 at 11 05 54
Screen Shot 2022-09-17 at 11 06 53

### DIFF
--- a/admin/src/components/atoms/ActionButton.tsx
+++ b/admin/src/components/atoms/ActionButton.tsx
@@ -2,7 +2,7 @@ import { Button, IconButton } from '@mui/material';
 import React, { ReactElement } from 'react';
 
 interface ActionButtonProps {
-  children: string;
+  children: React.ReactNode;
   label: boolean;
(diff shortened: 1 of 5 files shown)


yohamta:
Thanks for the fantastic improvement! I didn't know the Wave, but it looks very handy :)

PR #365, opened 2022-11-20
update examples for docker container

yohamta:
Add more explanation about running Docker container.


cocoonkid:
@yohamta 

How would you add --add-host=database:<host-ip> to a docker containers execute command?

PR #383, opened 2023-01-11
fix bug

helloworld2019:
When a node is not be started, graph should be consided still running, however the function 'isRunning' will return true,so it should be a bug.

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -392,6 +392,8 @@ func (sc *Scheduler) isRunning(g *ExecutionGraph) bool {
 		switch node.ReadStatus() {
 		case NodeStatus_Running:
 			return true
+		case NodeStatus_None:
+			return true
 		}
 	}
 	return false


helloworld2019:
I don't know why the test program failed, and test program log don't tell me which case failed. Whatever the bug reported by me yesterday really existed.

helloworld2019:
Thanks for your reply, I am using dagu in my work. I construct a dag: TaskA finishs and then TaskB start(TaskA->TaskB).A case I encountered: TaskA finished,and TaskB were not started. I call agent.Status(), and I found scheduler's status were finished, also node of TaskB were 'not started', however status of scheduler should be still running. So I traced the code, and found switch-case in the function 'isRunning' may not be correct. And I just simply correct the code, not run the test program in my machine. 
  I will continue correct the code these days

yohamta:
Thank you for the detailed explanation. I understand the issue.
I think that the code we need to fix is the function below.

https://github.com/yohamta/dagu/blob/main/agent.go#L87
```go
// Status returns the current status of the workflow.
func (a *Agent) Status() *models.Status {
	status := models.NewStatus(
		a.DAG,
		a.graph.Nodes(),
		a.scheduler.Status(a.graph),
		os.Getpid(),
		&a.graph.StartedAt,
		&a.graph.FinishedAt,
	)
	status.RequestId = a.requestId
	status.Log = a.logFilename
	if node := a.scheduler.HandlerNode(constants.OnExit); node != nil {
		status.OnExit = models.FromNode(node)
	}
	if node := a.scheduler.HandlerNode(constants.OnSuccess); node != nil {
		status.OnSuccess = models.FromNode(node)
	}
	if node := a.scheduler.HandlerNode(constants.OnFailure); node != nil {
		status.OnFailure = models.FromNode(node)
	}
	if node := a.scheduler.HandlerNode(constants.OnCancel); node != nil {
		status.OnCancel = models.FromNode(node)
	}
	return status
}
```

It gets the workflow status from the scheduler's function below:
```go
// Status returns the status of the scheduler.
func (sc *Scheduler) Status(g *ExecutionGraph) SchedulerStatus {
	if sc.IsCanceled() && !sc.checkStatus(g, []NodeStatus{
		NodeStatus_Success, NodeStatus_Skipped,
	}) {
		return SchedulerStatus_Cancel
	}
	if g.StartedAt.IsZero() {
		return SchedulerStatus_None
	}
	if sc.isRunning(g) {
		return SchedulerStatus_Running
	}
	if sc.lastError != nil {
		return SchedulerStatus_Error
	}
	return SchedulerStatus_Success
}
```

The problem is, If the returned status is `SchedulerStatus_None`, the function of the agent treats it as `not running` even though it is actually running. I think the agent should use `SchedulerStatus_Running` if the returned status is `SchedulerStatus_None`.

Maybe we can fix it like this (pseudo code).
```go
	let schSts = a.scheduler.Status(a.graph)
	if schSts == SchedulerStatus_None && !a.graph.StartedAt.IsZero():
		schSts = SchedulerStatus_Running:

	status := models.NewStatus(
		a.DAG,
		a.graph.Nodes(),
		schSts,
		os.Getpid(),
		&a.graph.StartedAt,
		&a.graph.FinishedAt,
	)
```

yohamta:
@helloworld2019 I'd appreciate if you could update the PR as above 🙏🏻

helloworld2019:
Maybe your advice is right in functionality(sorry, i am busy now, don't have time to test).But from the perspective of code readability and clean. your advice may not the best. Actually why not add more ‘if’ and 'switch-case' in the function 'isRunning'. I will try these days. Thanks.

PR #420, opened 2023-03-22
fix: server command host & port description swapped

stefaan1o:
Those descriptions seemed to be swapped.


yohamta (approved):
@stefaan1o Thank you very much!

PR #497, opened 2023-10-27
Fix attach logs for send with auth

triole:


triole:
I guess now it's fine. I added a composeMail method that gets used in both send mail methods. Seems to work nicely.

yohamta (approved):
Thank you for the update and for submitting another pull request. I will go ahead and release the updated binary. Thanks once again for your hard work!

Issue #505, opened 2023-12-08
Implementing Token-Based Authentication


yohamta:
Current authentication method: Basic Authentication.
Proposed enhancement: Token-Based Authentication.
Motivation: To make it easier to use the API, especially for JavaScript applications running in a browser from a different domain. Refer to the discussion here for more context: [GitHub Discussion #501](https://github.com/dagu-dev/dagu/discussions/501).

To-do:

- [x] Implement the configuration option for token-based authentication (`AuthToken` field).
- [x] Create an environment variable (`DAGU_AUTH_TOKEN`) for token management.
- [x] Implement middleware for token-based authentication.
- [ ] Adjust basic auth middleware to recognize and skip requests with valid tokens.
- [ ] Test that the Web UI remains unaffected by these changes.

smekuria1:
Hello, I would like to tackle this issue.

yohamta:
Hello, that's great! thanks!

PR #508, opened 2023-12-14
Token auth

smekuria1:
Adds #505 

This is my attempt to add basic simple Auth Token Authentication.
It works in a similar way to Basic Auth but can be expanded to be able to use JWTs( I can add that but it requires adding more libraries which i didn't want to do)

To-do:
- [x] Implement the configuration option for token-based authentication (AuthToken field).
- [x] Create an environment variable (DAGU_AUTH_TOKEN) for token management.
- [x] Implement middleware for token-based authentication.
- [x] Adjust basic auth middleware to recognize and skip requests with valid tokens.
- [x] Test that the Web UI remains unaffected by these changes.
When either Basic Auth or Auth Token is turned on the WebUI doesn't work which i believe is a feature since the webui is not configured to send tokens or credentials 
- [ ] Add docs for auth Token authentication

Wrong Auth Token
image

Correct Auth Token
image


service/frontend/middleware/tokenAuth.go
```diff
@@ -0,0 +1,33 @@
...
+	"crypto/subtle"
+	"fmt"
+	"net/http"
+)
+
+// TokenAuth implements a similar middleware handler like go-chi's BasicAuth middleware but for bearer tokens
+func TokenAuth(realm string, token string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			bearer := r.Header.Get("Bearer")
```
yohamta0: I think the token should be sent in `Authorization` header not in `Bearer` header.
I.e., `Authorization: Bearer <token>`

service/frontend/middleware/global.go
```diff
@@ -12,12 +13,16 @@ func SetupGlobalMiddleware(handler http.Handler) http.Handler {
 	next = middleware.Logger(next)
 	next = middleware.Recoverer(next)
 
+	if authToken != nil {
+		next = TokenAuth("bearer", authToken.Token)(next)
+		basicAuth = nil
```
yohamta0: I think that we don't need to set `basicAuth = nil` here which disables the basic auth middleware for all requests. I think we can skip the basicAuth middleware on each HTTP request basis. When an API token is set on a request, we can skip the basic authentication for the specific request only. As I put in the comment below, let's proceed without considering how to skip basic auth middleware.

service/frontend/middleware/tokenAuth.go
```diff
@@ -0,0 +1,33 @@
...
+			}
+
+			next.ServeHTTP(w, r)
+
+		})
+	}
+}
+
+func tokenAuthFailed(w http.ResponseWriter, realm string) {
+	w.Header().Add("WWW-Authenticate", fmt.Sprintf(`Basic realm="%s"`, realm))
```
yohamta0: Nitpick: I think `Basic` should be `Bearer` here.

service/frontend/middleware/global.go
```diff
@@ -12,12 +13,16 @@ func SetupGlobalMiddleware(handler http.Handler) http.Handler {
...
+	}
+
 	if basicAuth != nil {
 		next = middleware.BasicAuth(
 			"restricted", map[string]string{basicAuth.Username: basicAuth.Password},
 		)(next)
```
yohamta0, line 23: Thought: I think we can add a meta middleware (like `skipper` middleware) to skip basic authentication for a request with api token, which allows us to check http header and skip a specific middleware if the request holds a API token in the `Authentication` header. But it requires additional work so please don't mind for this PR.

yohamta (changes requested):
Thanks a lot for submitting the PR! I have added a few comment, I appreciate if you could make some small adjustments when you have time.

smekuria1:
Hi, I have implemented the changes your requested. Thank you so much for your instructions and if there is anything i need to add or change let me know.

ps i ran into some weird git issues so i apologize for some of the unnecessary commits

yohamta:
Looks good to me. Thank you so much for the great work!

PR #511, opened 2023-12-31
Skip basic auth when auth token is set

rafiramadhana:
Resolves https://github.com/dagu-dev/dagu/issues/509

Skip basic auth when auth token is set

In addition, I also rename `basicAuth` to `authBasic` for naming consistency with `authToken`


yohamta:
LGTM, thanks a lot for the amazing PR!

PR #515, opened 2024-01-01
Fix linter error

rafiramadhana:
Fix error when running `make go-lint`

```
➜  dagu git:(main) ✗ make go-lint
service/frontend/middleware/basic_auth_test.go:14:10: Error return value of `w.Write` is not checked (errcheck)
		w.Write([]byte("OK"))
		       ^
make: *** [Makefile:83: go-lint] Error 1
```

This error affects `make build`, so I think it is quite important to fix this


yohamta:
Thank you so much!

PR #558, opened 2024-04-17
made status of dags configurable

kriyanshishah:
earlier it only show status of dags from that day, so when if I need to know latest status of dag irrespective of when did it ran last there is a variable defined in the config "latestStatusToday". If it is true, it will show latest status from that day, or it will show the latest status irrespective of when it ran


internal/persistence/jsondb/database.go
```diff
@@ -307,9 +309,14 @@ func (store *Store) pattern(dagFile string) string {
 	return filepath.Join(dir, p)
 }
 
-func (store *Store) latestToday(dagFile string, day time.Time) (string, error) {
+func (store *Store) latestToday(dagFile string, day time.Time, latestStatusToday bool) (string, error) {
```
yohamta, line 312: I think we can use the following method instead of modifying this one:

https://github.com/dagu-dev/dagu/blob/151d842411e2492dd5e52f8852833e524fe39e7c/internal/persistence/jsondb/database.go#L139-L140

We can modify the following code to use the `ReadStatusRecent` method according to the configuration:

https://github.com/dagu-dev/dagu/blob/151d842411e2492dd5e52f8852833e524fe39e7c/internal/engine/engine.go#L189-L203

internal/persistence/jsondb/database.go
```diff
@@ -307,9 +309,14 @@ func (store *Store) pattern(dagFile string) string {
 	return filepath.Join(dir, p)
 }
 
-func (store *Store) latestToday(dagFile string, day time.Time) (string, error) {
```
yohamta0, line 310: I think we can use the following method instead of modifying this one:

https://github.com/dagu-dev/dagu/blob/151d842411e2492dd5e52f8852833e524fe39e7c/internal/persistence/jsondb/database.go#L139-L140

We can modify the following code to use the `ReadStatusRecent` method according to the configuration:

https://github.com/dagu-dev/dagu/blob/151d842411e2492dd5e52f8852833e524fe39e7c/internal/engine/engine.go#L189-L203

yohamta (commented):
Thank you very much for submitting the pull request! I've added a comment; please let me know your thoughts on it.

yohamta (approved):

kriyanshishah:
thank you. i am sorry was not able to push changes

PR #585, opened 2024-06-08
Fix http query map. Add http console debug.

x2ocoder:
http executor does not include query map when configured as documented

bug: key=value is excluded from request

```
steps:
  - name: send POST request
    command: POST https://foo.bar.com
    executor:
      type: http
      config:
        query: 
          key: "value"
```

mapstructure does not read query into QueryParams.  

Rename QueryParams into Query to match yaml

Add debug resty option


yohamta:
LGTM, thank you very much!

PR #586, opened 2024-06-08
fix http timeout syntax

x2ocoder:
yaml timeout value should not have comma after integer 10


yohamta:
LGTM, thank you very much!

PR #600, opened 2024-07-05
(bug) Remove ansi escape codes from UI Log Display

zph:
Reason: Ansi escape codes are intended to manipulate a terminal and when they're printed they make it more difficult to read the intended log content.

Current output: 
image

Desired output as represented by PR: 
image

Task definition:
```
steps:
  - name: step1
    command: bash $HOME/src/dagu/tester.sh
```

```bash
# tester.sh
#!/usr/bin/env bash

set -eou pipefail

source $HOME/bin/lib/colors

echo -e "${gre}green text, ${mag}magenta text, ${coloroff}default"
echo -e "${gre}green text, ${mag}magenta text, ${coloroff}default"
echo -e "${gre}green text, ${mag}magenta text, ${coloroff}default"
echo -e "${gre}green text, ${mag}magenta text, ${coloroff}default"
echo -e "${gre}green text, ${mag}magenta text, ${coloroff}default"
echo -e "${gre}green text, ${mag}magenta text, ${coloroff}default"
....
```

$HOME/lib/colors is: https://github.com/zph/dotfiles/blob/master/home/bin/lib/colors with a number of ansii escape codes for injecting color into terminal text.

This will also manifest with terminal based progress bars which use escape codes to refresh the screen's display or move to start of line.

Thanks for reviewing this and let me know if you'd like adjustments or clarification.

### DIFF
--- a/ui/src/components/organizations/ExecutionLog.tsx
+++ b/ui/src/components/organizations/ExecutionLog.tsx
@@ -10,10 +10,17 @@ type Props = {
   log?: LogFile;
 };
 
+// Credit: https://github.com/chalk/ansi-regex/commit/02fa893d619d3da85411acc8fd4e2eea0e95a9d9 under MIT license
+const ANSI_CODES_REGEX = [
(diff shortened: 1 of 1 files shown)


yohamta (approved):
✨✨LGTM✨✨, thanks a lot!

PR #615, opened 2024-07-14
Capture stderr from Docker container

### LINKED ISSUE #611: Docker executor does not propagate stderr output
Output from stderr does not appear in step logs

Example DAG:
```yaml
steps:
  - name: 'run'
    command: ['run']
    executor:
      type: 'docker'
      config:
        image: 'ubuntu:24.04'
        host:
          AutoRemove: true
          Binds:
            - '/tmp/run:/usr/bin/run'
```
Script `/tmp/run`:
```sh
#!/bin/sh

echo '<message to stdout>'
echo '<message to stderr>' 1>&2
```

#### Expectation

Stderr output appears in step logs:
image

#### Actual

Stderr output does not appear in step logs:
image

x4204:
Fixes #611 

Stderr output from Docker container is now correctly captured by Dagu

### DIFF
--- a/internal/dag/executor/docker.go
+++ b/internal/dag/executor/docker.go
@@ -95,6 +95,22 @@ func (e *docker) Run() error {
 		}
 	}()
 
+	out, err := cli.ContainerLogs(
+		ctx, resp.ID, types.ContainerLogsOptions{
+			ShowStdout: true,
+			ShowStderr: true,
+			Follow: true,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		_, err = stdcopy.StdCopy(e.stdout, e.stdout, out)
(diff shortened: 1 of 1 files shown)


internal/dag/executor/docker.go
```diff
@@ -109,16 +125,6 @@ func (e *docker) Run() error {
 		}
 	}
 
-	out, err := cli.ContainerLogs(
-		ctx, resp.ID, types.ContainerLogsOptions{ShowStdout: true},
-	)
```
x4204, line 114: Moved this above `cli.ContainerWait` so that log streaming starts immediately after container starts and not after the container exits

yohamta (approved):
✨✨LGTM✨✨🚀, thank you very much for fixing all the docker executor issues!

PR #630, opened 2024-07-24
Update actions/cache from v3 to v4

### LINKED ISSUE #629: Update actions/cache from v3 to v4
Node.js 16 used in actions/cache@v3 is deprecated since it reached EOL last November, so you should update to use Node.js 20.

image

Kiyo510:
Fixes #629


yohamta (approved):
LGTM🚀, thanks a lot!

PR #636, opened 2024-07-29
Fix toggling DAG suspension for DAGs with custom names

rocwang:
I find that toggling DAG suspension (i.e. "live") from frontend doesn't work when the DAG has a custom name defined:

Screenshot 2024-07-29 at 8 06 11 PM

Similar to #625, the frontend fix is extracting the file basename with out the `.yaml` bit as the DAG ID, and pass that to the backend for toggling.

However, upon fruther inspection, I also found the backend rest API doesn't return the right suspension status on a DAG with custom name. Furthermore, the scheduler also assumes a DAG's "name" is always the filename when calling `client.IsSuspended()`. So I fixed those as well.

It looks like `DAG.Name` is used quite a few times when actually a `DAG.ID` should be used instead. Maybe we should consider adding a new `ID` field to the `DAG` struct, which always stores the YAML filename, while `DAG.Name` can be overwritten with a custom name from the DAG spec.

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


yohamta:
Thank you very much for fixing the problem! Yes, I totally agree that we should add the `ID` field or method to `DAG` and use it throughout the internal code.

yohamta (approved):

PR #637, opened 2024-07-29
refactor: improve the pattern of "schedule" in the JSON schema file

rocwang:
The current JSON schema of DAGs doesn't support cron expressions well on the `schedule` field. This PR improves that. The cron pattern is copied from https://github.com/MediaMath/lambda-cron/blob/master/lambda_cron/schema.json#L11.

Before:

Screenshot 2024-07-29 at 10 41 15 PM

After: No more wiggly lines.

Screenshot 2024-07-29 at 10 42 16 PM

See also https://github.com/daguflow/dagu/issues/325

### DIFF
--- a/schemas/dag.schema.json
+++ b/schemas/dag.schema.json
@@ -13,7 +13,7 @@
     },
     "schedule": {
       "type": "string",
-      "pattern": "^[0-9* ]+$",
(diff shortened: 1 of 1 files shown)


yohamta (approved):
✨✨✨LGTM✨✨✨, thank you for improving the schema file!

PR #651, opened 2024-08-07
Fix log not loading when '&' in name

Lucaslah:
Fixes an issue where if a step has the '&' character in its name, the step log will not load.


yohamta (approved):
✨✨✨LGTM✨✨✨, thank you so much for the great contribution!

PR #654, opened 2024-08-08
ISSUE[581] Add Built-in Execution Context Variables

halalala222:
Pass Build-in Execution Context Variables as command args.
example:
```
handlerOn:
  success:
    command: "echo ${DAGU_REQUEST_ID} succeed"
```
image
image

### DIFF
--- a/internal/dag/executor/command.go
+++ b/internal/dag/executor/command.go
@@ -24,9 +24,9 @@ import (
 	"sync"
 	"syscall"
 
-	"github.com/daguflow/dagu/internal/util"
-
+	"github.com/daguflow/dagu/internal/constants"
 	"github.com/daguflow/dagu/internal/dag"
+	"github.com/daguflow/dagu/internal/util"
 )
 
 type commandExecutor struct {
@@ -40,9 +40,11 @@ func newCommand(ctx context.Context, step dag.Step) (Executor, error) {
 	if len(step.Dir) > 0 && !util.FileExists(step.Dir) {
 		return nil, fmt.Errorf("directory %q does not exist", step.Dir)
 	}
+	stepSpecialExecutionLogPathKey := util.GenerateStepSpecialExecutionLogPathKey(step.NodeID)
 	cmd.Dir = step.Dir
 	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Env = append(cmd.Env, step.Variables...)
+	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", constants.StepDaguExecutionLogPathKeySuffix, os.Getenv(stepSpecialExecutionLogPathKey)))
 	step.OutputVariables.Range(func(_, value any) bool {
 		cmd.Env = append(cmd.Env, value.(string))
 		return true
--- a/internal/dag/executor/command.go
+++ b/internal/dag/executor/command.go
@@ -24,7 +24,6 @@ import (
 	"sync"
 	"syscall"
 
-	"github.com/daguflow/dagu/internal/constants"
 	"github.com/daguflow/dagu/internal/dag"
 	"github.com/daguflow/dagu/internal/util"
 )
@@ -40,11 +39,15 @@ func newCommand(ctx context.Context, step dag.Step) (Executor, error) {
 	if len(step.Dir) > 0 && !util.FileExists(step.Dir) {
 		return nil, fmt.Errorf("directory %q does not exist", step.Dir)
 	}
-	stepSpecialExecutionLogPathKey := util.GenerateStepSpecialExecutionLogPathKey(step.NodeID)
+	dagContext, err := dag.GetContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("dagContext: %v\n", dagContext)
 	cmd.Dir = step.Dir
 	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Env = append(cmd.Env, step.Variables...)
-	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", constants.StepDaguExecutionLogPathKeySuffix, os.Getenv(stepSpecialExecutionLogPathKey)))
+	cmd.Env = append(cmd.Env, dagContext.Envs...)
 	step.OutputVariables.Range(func(_, value any) bool {
 		cmd.Env = append(cmd.Env, value.(string))
 		return true
--- a/internal/dag/executor/command.go
+++ b/internal/dag/executor/command.go
@@ -43,7 +43,7 @@ func newCommand(ctx context.Context, step dag.Step) (Executor, error) {
 	if err != nil {
 		return nil, err
 	}
-	fmt.Printf("dagContext: %v\n", dagContext)
+
 	cmd.Dir = step.Dir
 	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Env = append(cmd.Env, step.Variables...)
--- a/internal/dag/scheduler/node.go
+++ b/internal/dag/scheduler/node.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/daguflow/dagu/internal/constants"
 	"io"
 	"log"
 	"os"
@@ -359,6 +360,9 @@ func (n *Node) setupLog() error {
 	n.logLock.Lock()
 	defer n.logLock.Unlock()
 	var err error
+	if err = os.Setenv(util.GenerateStepSpecialExecutionLogPathKey(n.id), n.data.State.Log); err != nil {
+		return err
+	}
 	n.logFile, err = util.OpenOrCreateFile(n.data.State.Log)
 	if err != nil {
 		n.data.State.Error = err
@@ -432,6 +436,14 @@ func (n *Node) init() {
 		return
 	}
 	n.id = getNextNodeID()
+	n.data.Step.NodeID = n.id
+
+	n.data.Step.CmdWithArgs = strings.ReplaceAll(
+		n.data.Step.CmdWithArgs,
+		constants.StepDaguExecutionLogPathKeySuffix,
+		util.GenerateStepSpecialExecutionLogPathKey(n.id),
+	)
+
 	if n.data.Step.Variables == nil {
 		n.data.Step.Variables = []string{}
 	}
--- a/internal/dag/scheduler/node.go
+++ b/internal/dag/scheduler/node.go
@@ -131,6 +131,16 @@ func (n *Node) State() NodeState {
 
 // Execute runs the command synchronously and returns error if any.
 func (n *Node) Execute(ctx context.Context) error {
+	dagContext, err := dag.GetContext(ctx)
+	if err != nil {
+		return err
+	}
+	// set node special log path environment variable
+	if err = os.Setenv(dag.GenGlobalStepLogEnvKey(n.id), n.data.State.Log); err != nil {
+		return err
+	}
+	dagContext.Envs = append(dagContext.Envs, fmt.Sprintf("%s=%s", constants.StepDaguExecutionLogPathKeySuffix, n.data.State.Log))
+	ctx = dag.WithDagContext(ctx, dagContext)
 	cmd, err := n.setupExec(ctx)
 	if err != nil {
 		return err
@@ -361,9 +371,6 @@ func (n *Node) setupLog() error {
 	n.logLock.Lock()
 	defer n.logLock.Unlock()
 	var err error
-	if err = os.Setenv(util.GenerateStepSpecialExecutionLogPathKey(n.id), n.data.State.Log); err != nil {
-		return err
-	}
 	n.logFile, err = util.OpenOrCreateFile(n.data.State.Log)
 	if err != nil {
 		n.data.State.Error = err
@@ -437,12 +444,11 @@ func (n *Node) init() {
 		return
 	}
 	n.id = getNextNodeID()
-	n.data.Step.NodeID = n.id
 
 	n.data.Step.CmdWithArgs = strings.ReplaceAll(
 		n.data.Step.CmdWithArgs,
 		constants.StepDaguExecutionLogPathKeySuffix,
-		util.GenerateStepSpecialExecutionLogPathKey(n.id),
+		dag.GenGlobalStepLogEnvKey(n.id),
 	)
 
 	if n.data.Step.Variables == nil {
--- a/internal/dag/scheduler/node.go
+++ b/internal/dag/scheduler/node.go
@@ -129,6 +129,10 @@ func (n *Node) State() NodeState {
 
 // Execute runs the command synchronously and returns error if any.
 func (n *Node) Execute(ctx context.Context) error {
+	if err := n.setUpBuildInContextEnv(ctx); err != nil {
+		return err
+	}
+
 	cmd, err := n.setupExec(ctx)
 	if err != nil {
 		return err
@@ -366,6 +370,27 @@ func (n *Node) setupLog() error {
 	}
 	n.logWriter = bufio.NewWriter(n.logFile)
 	return nil
+}
+
+func (n *Node) setUpBuildInContextEnv(ctx context.Context) error {
+	var (
+		dagCtx dag.Context
+		err    error
+	)
+
+	if dagCtx, err = dag.GetContext(ctx); err != nil {
+		return err
+	}
+
+	if err = os.Setenv(dag.ExecutionLogPathEnvKey, n.data.Log); err != nil {
+		return err
+	}
+
+	if err = os.Setenv(dag.SchedulerLogPathEnvKey, dagCtx.DaguSchedulerLogPath); err != nil {
+		return err
+	}
+
+	return os.Setenv(dag.RequestIDEnvKey, dagCtx.DaguRequestID)
 }
 
 func (n *Node) teardown() error {
--- a/internal/dag/scheduler/node.go
+++ b/internal/dag/scheduler/node.go
@@ -129,10 +129,6 @@ func (n *Node) State() NodeState {
 
 // Execute runs the command synchronously and returns error if any.
 func (n *Node) Execute(ctx context.Context) error {
-	if err := n.setUpBuildInContextEnv(ctx); err != nil {
-		return err
-	}
-
 	cmd, err := n.setupExec(ctx)
 	if err != nil {
 		return err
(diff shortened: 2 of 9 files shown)


internal/dag/scheduler/node.go
```diff
@@ -369,6 +373,27 @@ func (n *Node) setupLog() error {
...
+		return err
+	}
+
+	if err = os.Setenv(dag.SchedulerLogPathEnvKey, dagCtx.DaguSchedulerLogPath); err != nil {
+		return err
+	}
+
+	return os.Setenv(dag.RequestIDEnvKey, dagCtx.DaguRequestID)
+}
+
```
yohamta0: Shouldn't we set the environment variables only for the child process rather than the entire agent process?

yohamta (commented):
Thanks! Looking good! I thought that we should replace those system environment variables in the DAG definition to intermediate representation during DAG building stage and replace back them to the original when running the node(step), but the code change does not seem to include that part. Wouldn't it be necessary?

internal/dag/scheduler/node.go
liooooo29: >Q:  I thought that we should replace those system environment variables in the DAG definition to intermediate representation during DAG building stage and replace back them to the original when running the node(step), but the code change does not seem to include that part. Wouldn't it be necessary?

A: I attempted to implement this feature this afternoon, but I encountered a problem. All the environment variables such as ${DAG_REQUEST_ID} and ${GOPATH}, as well as other configured environment variables, are converted to an intermediate state {{DAG_REQUEST_ID}}. In the executor, {{DAG_REQUEST_ID}} is converted back into environment variables like ${DAG_REQUEST_ID}. However, in the executor command, when using echo requestId: ${DAG_REQUEST_ID}, the logs show the result as below. I found that to handle the processing of environment variables in echo, the transformation needs to be done within step.commandWithArgs.
image
Therefore, I do not perform the conversion and instead handle it by adding environment variables directly using os.SetEnv before [util.SplitCommandWithParse](https://github.com/daguflow/dagu/blob/main/internal/dag/scheduler/node.go#L162)

liooooo29: >Q: Shouldn't we set the environment variables only for the child process rather than the entire agent process?

A: I am very sorry.I misunderstood, so I will rethink and adjust my code.

yohamta0: Oh no worries! Please keep up the great work. FYI, here's the code to set the environment variables: [command.go](https://github.com/daguflow/dagu/blob/3fe07adc55ab82afcb6ba6afdcbbc0abc90b0caf/internal/dag/executor/command.go#L44-L45)

liooooo29: Thank you for the reminder. I forgot about this piece of code this afternoon. I now understand how to adjust it.

liooooo29: Hi @yohamta , When i try to use [command.go](https://github.com/daguflow/dagu/blob/3fe07adc55ab82afcb6ba6afdcbbc0abc90b0caf/internal/dag/executor/command.go#L44-L45) to test enviroment variables, i find a problem.
```
func main() {
	cmd := exec.Command("echo", "${TEST_KEY}")
	cmd.Env = append(cmd.Env, "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	cmd.Run()
}
```
output : 
image

```
func main() {
	cmd := exec.Command("bash", "-c", "echo ${TEST_KEY}")
	cmd.Env = append(cmd.Env, "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	cmd.Run()
}
```
output : 
image

```
func main() {
	cmd := exec.Command("bash", "-c", "echo", "${TEST_KEY}")
	cmd.Env = append(cmd.Env, "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	cmd.Run()
}
```
output : 
image

go version: 1.22.5

liooooo29: > Oh no worries! Please keep up the great work. FYI, here's the code to set the environment variables: [command.go](https://github.com/daguflow/dagu/blob/3fe07adc55ab82afcb6ba6afdcbbc0abc90b0caf/internal/dag/executor/command.go#L44-L45)

@yohamta It can be observed in the test I provided above 
1. `cmd := exec.Command("echo", "${TEST_KEY}")` will expand the environment variable.
2. `cmd := exec.Command("bash", "-c", "echo ${TEST_KEY}")` can expand the environment variable.
3. `cmd := exec.Command("bash", "-c", "echo", "${TEST_KEY}")`, "echo" as a arg need to hold in one args.

yohamta0: Hey @halalala222,
I took a look at the test script and noticed the cause of the behavior. We need to add the existing environment variables to the command using `os.Environ()`. This is necessary because the process needs to know the `PATH` environment variable to run the `bash` command. Could you try this updated version and see if it works?
```go
package main

import (
	"fmt"
	"os"
	"os/exec"
)

func main() {
	cmd := exec.Command("bash", "echo ${TEST_KEY}")
	cmd.Env = append(os.Environ(), "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	err := cmd.Run()
	if err != nil {
		fmt.Println("Error:", err)
	}
}
```

liooooo29: Hey @yohamta 
I try again
```
func main() {
	cmd := exec.Command("echo", "${TEST_KEY}")
	cmd.Env = append(os.Environ(), "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	if err := cmd.Run(); err != nil {
		panic(err)
	}
}
```
output : 
image
```
func main() {
	cmd := exec.Command("bash", "-c", "echo ${TEST_KEY}")
	cmd.Env = append(os.Environ(), "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	if err := cmd.Run(); err != nil {
		panic(err)
	}
}
```
output : 
image

```
func main() {
	cmd := exec.Command("bash", "-c", "echo", "${TEST_KEY}")
	cmd.Env = append(os.Environ(), "TEST_KEY=testValue")
	cmd.Stdout = os.Stdout
	if err := cmd.Run(); err != nil {
		panic(err)
	}
}
```
output : 
image

same

yohamta0: @halalala222 Ah sorry for the confusion. Now I understand the issue better. To properly expand the environment variables, we need to do it in the parent process. Here's an idea that might work:

We could replace the original environment variables with unique names across all steps in the DAG. For example:
- `DAG_SCHEDULER_LOG_PATH` could become `${STEP_ID_DAG_SCHEDULER_LOG_PATH}`
- `DAG_EXECUTION_LOG_PATH` could become `${STEP_ID_DAG_EXECUTION_LOG_PATH}`

e.g.,
- `${STEP_1_DAG_SCHEDULER_LOG_PATH}`
- `${STEP_1_DAG_EXECUTION_LOG_PATH}`

by creating different environment variable keys for each step, we can safely set these environment variables on the agent process. The process would work like this:

During the DAG building stage, we replace the system environment variables with these unique, step-specific names. When running each node (step), we replace these unique names back to their original form.

I realize that currently, there's no ID field in the step object. However, we can easily solve this by automatically assigning internal IDs to each step during the DAG building stage. We could use simple sequential numbers like 1, 2, 3, and so on. This would give us the unique identifiers we need for our environment variable naming scheme.

How does that sound?

Note:
When spawning the child process, we should make sure to pass these variables using their original keys (like `DAG_SCHEDULER_LOG_PATH` etc). This is important because the child process will expect the environment variables in their original form, not our internal, step-specific versions.

liooooo29: @yohamta  hi! Your provided approach is quite good, but I've noticed it may lead to some problems：

1. The step in multiple DAGs needs to be incremented because if there are multiple DAGs running simultaneously, `${STEP_1_DAG_SCHEDULER_LOG_PATH}` could lead to conflicts.
2. When spawning the child process, we should make sure to pass these variables using their original keys (like `DAG_SCHEDULER_LOG_PATH `etc)
There could be conflicts in environment variables if multiple steps are running. 
for example : 
A: Dag test , running step1 , pass `${STEP_1_DAG_SCHEDULER_LOG_PATH}` to `DAG_SCHEDULER_LOG_PATH`
B: Dag anotherOneTest, running step2 , pass `${STEP_2_DAG_SCHEDULER_LOG_PATH}` to `DAG_SCHEDULER_LOG_PATH`
Running the steps of the two DAGs(A and B) simultaneously could result in conflicts with the DAG_SCHEDULER_LOG_PATH.

I believe converting each step's `DAG_SCHEDULER_LOG_PATH `to internal, step-specific versions is a good approach.Perhaps we can pass variables to {STEP_1_DAG_SCHEDULER_LOG_PATH} environment variable before [uitil.SplitCommandWithParse() function](https://github.com/daguflow/dagu/blob/main/internal/util/utils.go#L92)
Upon executing the [uitil.SplitCommandWithParse() function](https://github.com/daguflow/dagu/blob/main/internal/util/utils.go#L92) can substitute the corresponding ${STEP_1_DAG_SCHEDULER_LOG_PATH} environment variable with its own specific data.

3. If we are to convert `DAG_SCHEDULER_LOG_PATH `to `${STEP_ID_DAG_SCHEDULER_LOG_PATH}`, we need to specifically determine whether it is the designated environment variable, such as `DAG_SCHEDULER_LOG_PATH`. In this case, if the scenario is only used for echoing the corresponding build context variables in the command, we can directly replace the specific ${DAG_SCHEDULER_LOG_PATH}. Are there any other considerations for different scenarios?

I'm not sure if my explanation is clear, but if you have any new ideas, we can continue the discussion. Thank you for providing such a great approach.

yohamta0: Hi @halalala222, thank you for thinking about this so thoroughly. I have a few thoughts on your points.

> The step in multiple DAGs needs to be incremented because if there are multiple DAGs running simultaneously, `${STEP_1_DAG_SCHEDULER_LOG_PATH}` could lead to conflicts.
> When spawning the child process, we should make sure to pass these variables using their original keys (like `DAG_SCHEDULER_LOG_PATH` etc)
There could be conflicts in environment variables if multiple steps are running.
for example :
A: Dag test , running step1 , pass `${STEP_1_DAG_SCHEDULER_LOG_PATH}` to `DAG_SCHEDULER_LOG_PATH`
B: Dag anotherOneTest, running step2 , pass `${STEP_2_DAG_SCHEDULER_LOG_PATH}` to `DAG_SCHEDULER_LOG_PATH`
Running the steps of the two DAGs(A and B) simultaneously could result in conflicts with the `DAG_SCHEDULER_LOG_PATH`.

You're absolutely right that if we use some global value, it would lead to conflicts. However, I think this should be local. Setting an environment variable inside a process does not affect other processes. We can just assign a new internal id field to each step, starting from 1, every time, during the DAG building step. This value would be strictly scoped within the agent process, so we don't have to worry about conflicts with other concurrent workflow executions.

Regarding your example:

> A: Dag test, running step1, pass `${STEP_1_DAG_SCHEDULER_LOG_PATH}` to `DAG_SCHEDULER_LOG_PATH`
> B: `Dag anotherOneTest, running step2, pass ${STEP_2_DAG_SCHEDULER_LOG_PATH}` to `DAG_SCHEDULER_LOG_PATH`

Running the steps of the two DAGs (A and B) simultaneously wouldn't result in conflicts with the `DAG_SCHEDULER_LOG_PATH` because each would be in its own process.

> If we are to convert DAG_SCHEDULER_LOG_PATH to ${STEP_ID_DAG_SCHEDULER_LOG_PATH}, we need to specifically determine whether it is the designated environment variable, such as DAG_SCHEDULER_LOG_PATH. In this case, if the scenario is only used for echoing the corresponding build context variables in the command, we can directly replace the specific ${DAG_SCHEDULER_LOG_PATH}. Are there any other considerations for different scenarios?

Yes, there are other scenarios to consider. For example, users might want to use the environment variable inside a Python script, where we can't replace the string. So I think just replacing the command arguments doesn't work for all use cases.

Thank you for bringing up these important considerations. Let me know if you have any thoughts on this or if there's anything else we should discuss.

However, I've been thinking that this might be a more complex implementation than we initially thought, with potential impacts in various areas of the codebase. Sorry for the issue description was not very clear in details. Given this complexity, it might be better to put this aside for now and focus on other issues, for example paging parameters for DAG list API. What do you think?

Thank you for your hard work and deep thoughts on this PR. It is very helpful for me to clarify the problem and potential solutions significantly. Your contributions are truly valuable.

liooooo29: Hi @yohamta Thank you for your reply! I think so,  it might be better to put aside this for now and focus on DAG list API pagination.

liooooo29: @yohamta hi!I will start working on this issue again!

yohamta0: Hi @halalala222, sure! Thank you very much! That would be awesome!

halalala222:
## command exec go run main.go
```
steps:
  - name: test5_steps1
    dir: /home/liooooo/draft/cmdTest
    command: go run main.go
```
```
// main.go
func main() {
	requestID := os.Getenv("DAGU_REQUEST_ID")
	executionLogPath := os.Getenv("DAGU_EXECUTION_LOG_PATH")
	schedulerLogPath := os.Getenv("DAGU_SCHEDULER_LOG_PATH")

	if requestID == "" {
		fmt.Printf("Request ID is empty\n")
	} else {
		fmt.Printf("Request ID is %s\n", requestID)
	}

	if executionLogPath == "" {
		fmt.Printf("Execution log path is empty\n")
	} else {
		fmt.Printf("Execution log path is %s\n", executionLogPath)
	}

	if schedulerLogPath == "" {
		fmt.Printf("Scheduler log path is empty\n")
	} else {
		fmt.Printf("Scheduler log path is %s\n", schedulerLogPath)
	}
}
```

### 1. Concurrent execution of multiple DAGs.
GkDOiSyDEl
test2
4dBFPEhLYI
tbaQ28yKom
test10
eTQsAhDAY0
qznZ1VtwBF
test11
rmXesN1zD4
eOw3fIOW6e
test3
4QSkCcwQVr
90bfb538-85f9-4d93-9a24-cbcbf60b4bae
test4
IFmq8FSMQM
eQPA2Ri19t
test5
ztB8XLxEkF
xlweDIccij
test6
ZuIrLc39TL
9Py1nLRSWE
test7
zbFtsZRCjz
0toEOnrQkx
test8
NfZOKydcbu
A8YEy9Qg8s
test9
9N7WMho8KN
MSbxwIj519
### Concurrent execution of multiple steps within the same DAG.
ROkovvU5dl
step1
4A65p0Plfl
step2
xJA9aJU6b6
step3
5xvBsh4MXL
step4
85D13PpsOL
step5
CfWVXPmLCY
step6
b4dec93a-9de1-4941-b5e7-46384d49966f
step7
sUB2MnZfzv
step8
B6x00DNDSZ
step9
IUbdaq6GJK
step10
PTmy02ZvjF
step11
1cf852d7-0f37-4aee-bacb-bc7d25b594aa

halalala222:
## Command echo
```
steps:
  - name: step1
    command: echo ${DAGU_REQUEST_ID} , ${DAGU_EXECUTION_LOG_PATH} , ${DAGU_SCHEDULER_LOG_PATH}
  - name: step2
    command: echo ${DAGU_REQUEST_ID} , ${DAGU_EXECUTION_LOG_PATH} , ${DAGU_SCHEDULER_LOG_PATH}
  - name: step3
    command: echo ${DAGU_REQUEST_ID} , ${DAGU_EXECUTION_LOG_PATH} , ${DAGU_SCHEDULER_LOG_PATH}
  - name: step4
    command: echo ${DAGU_REQUEST_ID} , ${DAGU_EXECUTION_LOG_PATH} , ${DAGU_SCHEDULER_LOG_PATH}
  - name: step5
    command: echo ${DAGU_REQUEST_ID} , ${DAGU_EXECUTION_LOG_PATH} , ${DAGU_SCHEDULER_LOG_PATH}
  - name: step6
    command: echo ${DAGU_REQUEST_ID} , ${DAGU_EXECUTION_LOG_PATH} , ${DAGU_SCHEDULER_LOG_PATH}
```
image
image
image
image
image
image

halalala222:
@yohamta Hi, I've updated my PR.

internal/dag/executor/command.go
```diff
@@ -40,9 +40,11 @@ func newCommand(ctx context.Context, step dag.Step) (Executor, error) {
 	if len(step.Dir) > 0 && !util.FileExists(step.Dir) {
 		return nil, fmt.Errorf("directory %q does not exist", step.Dir)
 	}
+	stepSpecialExecutionLogPathKey := util.GenerateStepSpecialExecutionLogPathKey(step.NodeID)
 	cmd.Dir = step.Dir
 	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Env = append(cmd.Env, step.Variables...)
+	cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", constants.StepDaguExecutionLogPathKeySuffix, os.Getenv(stepSpecialExecutionLogPathKey)))
```
yohamta, line 47: Can we change this to something like this?
```suggestion
	// Set environment variables from the DAG context.
	dagContext, err := dag.GetContext(ctx)
	if err != nil {
		return nil, err
	}
	cmd.Env = append(cmd.Env, dagContext.Envs...)
```

yohamta0: Can we change this to something like this?
```suggestion
	// Set environment variables from the DAG context.
	dagContext, err := dag.GetContext(ctx)
	if err != nil {
		return nil, err
	}
	cmd.Env = append(cmd.Env, dagContext.Envs...)
```

internal/dag/scheduler/node.go
```diff
@@ -359,6 +361,9 @@ func (n *Node) setupLog() error {
 	n.logLock.Lock()
 	defer n.logLock.Unlock()
 	var err error
+	if err = os.Setenv(util.GenerateStepSpecialExecutionLogPathKey(n.id), n.data.State.Log); err != nil {
+		return err
+	}
```
yohamta0: Can we use DAG context to pass down the special environment variables?
```suggestion
// Execute runs the command synchronously and returns error if any.
func (n *Node) Execute(ctx context.Context) error {
	dagContext, err := dag.GetContext(ctx)
	if err !=nil {
		return err
	}

	// set special environment variables
        globalStepLogEnvKey := dag.GenGlobalStepLogEnvKey(n.Step.ID)
        if err = os.Setenv(globalStepLogEnvKey, n.data.State.Log); err != nil {
		return err
	}
	dagContext.Envs = append(
		dagContext.Envs, 
		dag.StepLogPathEnvKey + "=" + n.data.State.Log,
	)
	ctx = dag.WithContext(ctx, dagContext)
        // ......
```

internal/util/utils.go
```diff
@@ -225,3 +228,21 @@ func AddYamlExtension(file string) string {
 	}
 	return file
 }
+
+func GenerateStepSpecialExecutionLogPathKey(stepID int) string {
```
yohamta0: Can we define this function within the `dag` package? I wonder that if it could be smaller scoped, what do you think?

internal/dag/scheduler/node.go
liooooo29: yes!!!!! Thank you!

internal/util/utils.go
liooooo29: I think so. Thank you!
I will modify my code during the day tomorrow.

liooooo29: @yohamta Hi! Hi, I've updated my PR.

yohamta: Thank you very much for the great work!! I will take a look soon!

yohamta (approved):
Looking good!👍✨✨✨🚀🚀🚀 Thank you very much!

PR #655, opened 2024-08-08
ISSUE[584] ssh executor does not support login in password

halalala222:
support login in with password while executor config not contain ssh key
1. executor config change add password
image
2. result
step log file :
image
in my machine
image


yohamta (approved):
✨✨✨LGTM✨✨✨, thank you very much!

PR #656, opened 2024-08-08
ISSUE[578] Add json bool configuration option to HTTP executor

halalala222:
HTTP executor supports writing the HTTP response in JSON format.
```
{
  "status_code": 200,
  "headers": [
    "xxx": "yyy"
  ],
  "body": {
    "some_field": "some_value"
  }
}
```
HTTP executor config
image
Result
image
image
image
image

### DIFF
--- a/internal/dag/executor/http.go
+++ b/internal/dag/executor/http.go
@@ -49,7 +49,7 @@ type httpConfig struct {
 	Json    bool              `json:"json"`
 }
 
-type HttpJSONResult struct {
+type httpJSONResult struct {
 	StatusCode int                 `json:"status_code"`
 	Headers    map[string][]string `json:"headers"`
 	Body       map[string]any      `json:"body"`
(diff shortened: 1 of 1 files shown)


internal/dag/executor/http.go
```diff
@@ -46,6 +46,13 @@ type httpConfig struct {
 	Body    string            `json:"body"`
 	Silent  bool              `json:"silent"`
 	Debug   bool              `json:"debug"`
+	Json    bool              `json:"json"`
+}
+
+type httpJSONResult struct {
+	StatusCode int                 `json:"status_code"`
+	Headers    map[string][]string `json:"headers"`
+	Body       map[string]any      `json:"body"`
```
yohamta0: Should we consider changing the `Body` type to `any` because valid JSON can be arrays, primitives, or null, not just objects?

yohamta (commented):
Looking good! Just a confirmation about the `Body` type.

internal/dag/executor/http.go
liooooo29: You are right, I didn't consider it thoroughly. Thank you for the reminder.

Issue #592, opened 2024-06-20
formatting of error text in web ui is not very noticeable


alext-extracellular:
Hi there, 
I was editing my DAGs and for a few minutes I did not know what was wrong, I thought the webui had crashed, but it was because I did not see the error message. Can I suggest some more obvious formatting, like red box, etc. to make it more obvious?

See image:

![image](https://github.com/dagu-dev/dagu/assets/142899482/e3c70a9a-35a4-4bbc-9b86-2b839dcbad31)

yohamta:
Good point. Let's make it to be more noticeable. I thought we can add an alert modal on save. Here's an example.
<img width="637" alt="image" src="https://github.com/dagu-dev/dagu/assets/1475839/38066358-7faa-4210-900d-ff4d3c2906e3">

halalala222:
@yohamta Can I try to work on this issue ?

halalala222:
@yohamta Hi! Can i try on this issue ?

yohamta:
Hi @halalala222, thank you very much for the continuous contribution! It's being really helpful. Sure, please go ahead.

PR #670, opened 2024-08-19
[ISSUE #592] formatting of error text in web ui is not very noticeable

halalala222:
1. make error text more noticeable
like :
image


yohamta (approved):
✨✨✨LGTM✨✨✨🚀, thank you very much!

PR #693, opened 2024-10-17
Add command to run server in docker compose file

KMe72:
This pull request addresses an issue where both containers were running a scheduler, leading to DAGs being scheduled twice.


yohamta:
Looking great, thank you for improving the example docker-compose file!

PR #696, opened 2024-11-04
fix: use the server timezone to parse the cron expression

jonnochoo:
There is an issue if the server is running in a timezone (eg. UTC), but the client's browser timezone is in another timezone (eg. Australia/Sydney), the cron expression incorrectly shows the Next Run incorrectly.

## Next Steps

Open on thoughts on the best way to get the server's timezone? Create an API endpoint? Open to feedback

### DIFF
--- a/ui/src/components/molecules/DAGTable.tsx
+++ b/ui/src/components/molecules/DAGTable.tsx
@@ -251,7 +251,9 @@ const defaultColumns = [
   }),
   columnHelper.accessor('Type', {
     id: 'Schedule',
-    header: 'Schedule',
+    header: getConfig().timeZone
+      ? `Schedule in ${getConfig().timeZone}`
+      : 'Schedule',
     enableSorting: true,
     cell: (props) => {
       const data = props.row.original!;
@@ -389,7 +391,15 @@ const defaultColumns = [
   }),
 ];
 
-function DAGTable({ DAGs = [], group = '', refreshFn, searchText, handleSearchTextChange, searchTag, handleSearchTagChange }: Props) {
+function DAGTable({
+  DAGs = [],
+  group = '',
+  refreshFn,
+  searchText,
+  handleSearchTextChange,
+  searchTag,
+  handleSearchTagChange,
+}: Props) {
   const [columns] = React.useState<typeof defaultColumns>(() => [
     ...defaultColumns,
   ]);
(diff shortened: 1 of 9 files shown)


yohamta:
Hi @jonnochoo, thank you for draft PR! This looks amazing. Yes, this does need to be fixed. I would suggest adding a new configuration key, `timezone`, to pass the server's timezone to the frontend. Here are the files that need modification. If possible, it would be great if you could make these changes in this PR.

https://github.com/dagu-org/dagu/blob/22fe495b06f79a9ace512e1ab255b5e3634cbafd/internal/frontend/templates/base.gohtml#L10-L19

https://github.com/dagu-org/dagu/blob/22fe495b06f79a9ace512e1ab255b5e3634cbafd/internal/frontend/server/templates.go#L62-L84

ui/src/models/index.ts
```diff
@@ -150,9 +150,16 @@ export function getNextSchedule(data: WorkflowListItem): number {
   if (!schedules || schedules.length == 0 || data.Suspended) {
     return Number.MAX_SAFE_INTEGER;
   }
-  const datesToRun = schedules.map((s) =>
-    cronParser.parseExpression(s.Expression).next()
-  );
+  const tz = getConfig().timeZone;
```
jonnochoo, line 153: For backwards compatibility tz won't be set, so it should defer to the original behaviour.

ui/src/components/molecules/DAGTable.tsx
```diff
@@ -251,7 +251,9 @@ const defaultColumns = [
   }),
   columnHelper.accessor('Type', {
     id: 'Schedule',
-    header: 'Schedule',
+    header: getConfig().timeZone
```
jonnochoo, line 254: It might be helpful to see what timezone the server is running in?

jonnochoo:
> Hi @jonnochoo, thank you for draft PR! This looks amazing. Yes, this does need to be fixed. I would suggest adding a new configuration key, `timezone`, to pass the server's timezone to the frontend. Here are the files that need modification. If possible, it would be great if you could make these changes in this PR.
> 
> https://github.com/dagu-org/dagu/blob/22fe495b06f79a9ace512e1ab255b5e3634cbafd/internal/frontend/templates/base.gohtml#L10-L19
> 
> https://github.com/dagu-org/dagu/blob/22fe495b06f79a9ace512e1ab255b5e3634cbafd/internal/frontend/server/templates.go#L62-L84

Thanks for the pointers, I have made the change and you can see in the screenshot it works.

The server time it UTC but the client/browser is in Sydney time, and the cron expression shows the correct next run

image

yohamta (approved):
LGTM ✨✨✨🚀🚀🚀✨✨✨ Thank you very much for the amazing work!

PR #707, opened 2024-11-11
fixed bug when using the CRON_TZ= cron expression

jonnochoo:
Related to https://github.com/dagu-org/dagu/issues/706, when using the `CRON_Z` in an expression

```
CRON_TZ=Asia/Tokyo 0 6 * * *
```

The `/dags`  page doesn't load, see below
image

The fixed version

image


yohamta:
Looking great! Thank you for the amazing fix!

yohamta (approved):

PR #714, opened 2024-11-16
feat: Support configurable base path for server

chrishoage:
Closes: https://github.com/dagu-org/dagu/issues/576
Closes: https://github.com/dagu-org/dagu/issues/575

This PR Adds a BasePath config that allows the go server to respond to requests at a specific  sub path. 

To test supply the BaseBath config though the environment variable `DAGU_BASE_PATH=/foo` or through the admin.yaml file 

1. Load dagu at the root `/` 
2. Should be redirected to `/foo`
3. dagu should work with out regressions
4. Restart dagu with out base path 
5. dagu should work with out regressions as before

### DIFF
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -25,6 +25,7 @@ Server Configuration
 ~~~~~~~~~~~~~~~~~~
 - ``DAGU_HOST`` (``127.0.0.1``): Server binding host
 - ``DAGU_PORT`` (``8080``): Server binding port
+- ``DAGU_BASE_PATH`` (``""``): Base path to serve the application
 - ``DAGU_TZ`` (``""``): Server timezone (default: system timezone)
 - ``DAGU_CERT_FILE``: SSL certificate file path
 - ``DAGU_KEY_FILE``: SSL key file path
@@ -59,6 +60,7 @@ Create ``admin.yaml`` in ``$HOME/.config/dagu/`` to override default settings. B
     # Server Configuration
     host: "127.0.0.1"                # Web UI hostname
     port: 8080                       # Web UI port
+    basePath: ""                    # Base path to serve the application
     tz: "Asia/Tokyo"                 # Timezone (e.g., "America/New_York")
     
     # Directory Configuration
--- a/internal/frontend/middleware/global.go
+++ b/internal/frontend/middleware/global.go
@@ -18,6 +18,7 @@ package middleware
 import (
 	"context"
 	"net/http"
+	"path"
 	"strings"
 
 	"github.com/dagu-org/dagu/internal/logger"
@@ -72,6 +73,7 @@ var (
 	authBasic      *AuthBasic
 	authToken      *AuthToken
 	appLogger      logger.Logger
+	basePath       string
 )
 
 type Options struct {
@@ -79,6 +81,7 @@ type Options struct {
 	AuthBasic *AuthBasic
 	AuthToken *AuthToken
 	Logger    logger.Logger
+	BasePath  string
 }
 
 type AuthBasic struct {
@@ -95,15 +98,21 @@ func Setup(opts *Options) {
 	authBasic = opts.AuthBasic
 	authToken = opts.AuthToken
 	appLogger = opts.Logger
+	basePath = opts.BasePath
 }
 
 func prefixChecker(next http.Handler) http.Handler {
 	return http.HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			if strings.HasPrefix(r.URL.Path, "/api") {
-				next.ServeHTTP(w, r)
+			if basePath != "" && r.URL.Path == "/" {
+				http.Redirect(w, r, basePath, http.StatusSeeOther)
+				return
+			}
+
+			if strings.HasPrefix(r.URL.Path, path.Join(basePath, "/api")) {
+				http.StripPrefix(basePath, next).ServeHTTP(w, r)
 			} else {
-				defaultHandler.ServeHTTP(w, r)
+				http.StripPrefix(basePath, defaultHandler).ServeHTTP(w, r)
 			}
 		})
 }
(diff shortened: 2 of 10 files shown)


docs/source/config.rst
```diff
@@ -25,6 +25,7 @@ Server Configuration
 ~~~~~~~~~~~~~~~~~~
 - ``DAGU_HOST`` (``127.0.0.1``): Server binding host
 - ``DAGU_PORT`` (``8080``): Server binding port
+- ``DAGU_BASE_PATH`` (``""``): Base path to serve the application
```
chrishoage, line 28: While the ticket says `BAUSE_URL` I feel `PATH` is more accurate. 

URL implies including the origin as well, and we're only supplying the path of the URL. 

I'm happy to change this if desired - it just seemed better to be accurate here. A note: `ApiBaseURL` should also be called path but that seemed out of scope for this work.

internal/frontend/middleware/global.go
```diff
@@ -95,15 +98,21 @@ func Setup(opts *Options) {
...
 
 func prefixChecker(next http.Handler) http.Handler {
 	return http.HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			if strings.HasPrefix(r.URL.Path, "/api") {
-				next.ServeHTTP(w, r)
+			if basePath != "" && r.URL.Path == "/" {
+				http.Redirect(w, r, basePath, http.StatusSeeOther)
+				return
+			}
```
chrishoage: If the root of the server is hit and we have a base path configured redirect to that base path for convenience.

internal/frontend/middleware/global.go
```diff
@@ -95,15 +98,21 @@ func Setup(opts *Options) {
...
+
+			if strings.HasPrefix(r.URL.Path, path.Join(basePath, "/api")) {
```
chrishoage: Is there a better solution to `path.Join` here? It doesn't feel robust. 

It really feels like `/api` should be a handler which is hung off of `http.StripPrefix(basePath, next)` such that any route registered under it would be handled by `/api`

I wasn't sure of the consequences of  such a change however, so I just updated the string comparison.

ui/webpack.prod.js
```diff
@@ -21,7 +21,7 @@ module.exports = merge(common, {
   output: {
     filename: 'bundle.js',
     path: path.resolve(__dirname, 'dist'),
-    publicPath: '/assets/',
+    publicPath: 'auto',
```
chrishoage, line 24: Allow [webpack to auto figure out the path](https://webpack.js.org/guides/public-path/#automatic-publicpath) for all other assets based on the the location of the entry asset served

internal/frontend/middleware/global.go
```diff
@@ -95,15 +98,21 @@ func Setup(opts *Options) {
...
+				http.StripPrefix(basePath, next).ServeHTTP(w, r)
 			} else {
-				defaultHandler.ServeHTTP(w, r)
+				http.StripPrefix(basePath, defaultHandler).ServeHTTP(w, r)
```
chrishoage: `http.StripPrefix` seemed like the cleanest way to allow all existing routes to work as before. 

This is my first time writing Go so if there's a more idiomatic way of doing this let me know!

internal/frontend/middleware/global.go
yohamta0: I've tested this locally with an nginx reverse proxy (configuration below) but encountered an issue when accessing via the proxy path `http://localhost/dagu/`. It seems the path prefix check might be the cause. The following code, however, works as expected:

```go
if strings.HasPrefix(r.URL.Path, "/api") {
    next.ServeHTTP(w, r)
} else {
    defaultHandler.ServeHTTP(w, r)
}
```

Would you mind taking a look when you have a chance?

Test setup
```nginx
server {
    listen       80;
    server_name  localhost;
    location /dagu/ {
        proxy_pass http://localhost:8080/;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
    }
}
```

~/.dagu/admin.cfg
```yaml
basePath: /dagu/
port: 8080
```

ui/webpack.prod.js
yohamta: Oh I didn't know that. Looking better, thanks!

yohamta (commented):
Overall looking amazing, thank you so much for the great PR! I've left a small question, so please take a look when you have a moment :)

docs/source/config.rst
yohamta0: Yes, 100% agreed. Thank you!

ui/webpack.prod.js
yohamta0: Looking nice, thanks!

internal/frontend/middleware/global.go
chrishoage: The issue is the trailing slash. I was only testing with `/dagu` and missed this case. 

I have made the BasePath more robust 

- Normalize the value in config.go so that all parts of the app can be sure it has the same shape
- changed `prefixChecker` to strip the prefix first, then do the strings.HasPrefix check in a HandlerFunc to not have to worry about the prefix. 

I tested with 

- no basepath set at all
- basepath set as an empty string
- basePath set as `/`
- basePath set as `/dagu`
- basePath set as `/dagu/`
- basePath set as `/dagu/foo`

I believe this should be good. 

I have pushed up the fix as a `!fixup` commit to make it easier to review. This can be "squashed" when merging or I can squash it before this is ready to merge.

internal/frontend/middleware/global.go
chrishoage: See https://github.com/dagu-org/dagu/pull/714#discussion_r1845554710 this has been improved to not rely on path.join for this prefix check

chrishoage:
@yohamta This should be good for another review, thank you!

yohamta (approved):
Looking great, thank you so much!

PR #723, opened 2024-11-21
Improve Dockerfile to reduce amount of config needed in docker-compose

### LINKED ISSUE #715: Consider more idomatic docker config
While working on my PR I made extensive changes to the Dockerfile and I'm wondering if you think I should upstream them. 

- I added an `entrypoint.sh` which configured the environment
- Using the entrypoint allow the user to specify the DOCKER_GID and add `dagu` to the docker group
- Use su-exec to execute dagu as the dagu user
- Changed the home directory and DAGU_HOME to be `/config`. This is a more idiomatic way of handling docker container configs
- removed sudo and dagu in sudoers

What this allows for is a much cleaner docker compose setup. 

- No need to run a container o run socat and forward the docker socket. Just bind to to the container
- Allows users to load extra config though `/entrypoint.d/*.sh` (this is not done in my test branch, but could very easily be added. I have a lazy ADDITIONAL_PACKAGES but I don't think this should be shipped as is) 

You can look at my changes here https://github.com/chrishoage/dagu/commit/23eae41e4dc5a3fffe227c134e7cbd2936089443 and if any of these are worth upstreaming I can produce a PR with the parts you think would be a good fit for the official docker image. 

```yaml
services:
  caddy:
    image: caddy
    container_name: caddy
    hostname: caddy
    ports:
      - "8080:80"
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile
  dagu:
    image: ch/dagu
    container_name: dagu
    hostname: dagu
    volumes:
      - ./config:/config
      - ./scripts:/scripts
      - /var/run/docker.sock:/var/run/docker.sock
    environment:
      # - ADDITIONAL_PACKAGES=docker-cli deno bash
      - PUID=1000
      - PGID=1000
      - DOCKER_GID=970
      - DAGU_TZ=America/Los_Angeles
```

```Caddyfile
http://localhost {
  handle /dagu* {
    reverse_proxy dagu:8080
  }
}
```

chrishoage:
Closes #715 

This work significantly changes the Dockerfile and the required docker-compose.yaml configuration. 

The changes include

- Moving to a more idiomatic `/config` directory using `DAGU_HOME`. This relies on the resolver.go behavior that places all files together when `appHomeEnv` is specified. 
- Introducing an `entrypoint.sh` file that does the following
  - Sets up the docker group if DOCKER_GID is provided
  - Updates the dagu user to the provided PUID and PGID if provided
  - Tries to guess if the running container is using the previous recommended setup using `/home/dagu` inside the container - if so it uses it 
  - Otherwise sets `DAGU_HOME=/config` if DAGU_HOME is not already set 
- Runs any scripts in `/etc/custom-init.d` during boot. This can be used to install things into the container - like python, node, etc
-  Finally executes dagu using `su-exec` 

In order to test use the updated documentation. To test the "legacy" home behavior use the following compose file 

```
services:
  dagu:
    image: dagu
    container_name: dagu
    hostname: dagu
    ports:
      - "8080:8080"
    environment:
      - DOCKER_GID=970
      - DAGU_BASE_PATH=/dagu
    volumes:
      - ./test-init:/etc/custom-init.d
      - /var/run/docker.sock:/var/run/docker.sock
      - ./test-config:/home/dagu/.config/dagu
      - ./test-data:/home/dagu/.local/share
```

Where `test-config` and `test-data` map to the previous docker volume mounts. 

The "guessing" is best effort and it could be fooled, however I tried to make it as robust as I could. 

This is a pretty large change, so please feel free to try to test this as thoroughly as possible. I'm happy to make more improvements or changes before merging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ WORKDIR /app
 COPY ui/ ./
 
 RUN rm -rf node_modules; \
-    yarn install --frozen-lockfile --non-interactive; \
-    yarn build
+  yarn install --frozen-lockfile --non-interactive; \
+  yarn build
 
 # Stage 2: Go Builder
 FROM --platform=$TARGETPLATFORM golang:1.22-alpine as go-builder
@@ -32,35 +32,28 @@ ARG USER="dagu"
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 
-# Create user and set permissions
-RUN apk update; \
-    apk add --no-cache sudo tzdata; \
-    addgroup -g ${USER_GID} ${USER}; \
-    adduser ${USER} -h /home/${USER} -u ${USER_UID} -G ${USER} -D -s /bin/ash; \
-    echo ${USER} ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/${USER}; \
-    chmod 0440 /etc/sudoers.d/${USER}; \
-    mkdir -p .config/dagu/dags; \
-    chown -R ${USER}:${USER} /home/${USER};
-
 COPY --from=go-builder /app/bin/dagu /usr/local/bin/
+COPY ./entrypoint.sh /entrypoint.sh
 
-USER ${USER}
-WORKDIR /home/${USER}
+# Create user and set permissions
+RUN apk update && \
+  apk add --no-cache su-exec shadow tzdata && \
+  addgroup -g ${USER_GID} ${USER} && \
+  adduser ${USER} -h /config -u ${USER_UID} -G ${USER} -D -s /bin/ash && \
+  chown -R ${USER}:${USER} /config && \
+  chmod +x /entrypoint.sh;
 
-# Add the hello_world.yaml file
-COPY --chown=${USER}:${USER} <<EOF .config/dagu/dags/hello_world.yaml
-schedule: "* * * * *"
-steps:
-  - name: hello world
-    command: sh
-    script: |
-      echo "Hello, world!"
-EOF
+WORKDIR /config
+
 
 ENV DAGU_HOST=0.0.0.0
 ENV DAGU_PORT=8080
 ENV DAGU_TZ="Etc/UTC"
+ENV PUID=${USER_UID}
+ENV PGID=${USER_GID}
+ENV DOCKER_GID=-1
 
 EXPOSE 8080
 
+ENTRYPOINT ["/entrypoint.sh"]
 CMD ["dagu", "start-all"]
--- a/docs/source/docker-compose.rst
+++ b/docs/source/docker-compose.rst
@@ -1,48 +1,28 @@
+.. _Using Docker Compose:
 Using Docker Compose
 ===================================
 
-To automate DAG executions based on cron expressions, it is necessary to run both the ui server and scheduler process. Here is an example `docker-compose.yml` setup for running Dagu using Docker Compose.
+Here is an example `docker-compose.yml` setup for running Dagu using Docker Compose.
 
 .. code-block:: yaml
 
-  version: "3.9"
-  services:
-      # init container updates permission
-      init:
-          image: "ghcr.io/dagu-org/dagu:latest"
-          user: root
-          volumes:
-              - dagu_config:/home/dagu/.config/dagu
-              - dagu_data:/home/dagu/.local/share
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -32,35 +32,28 @@ ARG USER="dagu"
...
 
-# Add the hello_world.yaml file
-COPY --chown=${USER}:${USER} <<EOF .config/dagu/dags/hello_world.yaml
-schedule: "* * * * *"
-steps:
-  - name: hello world
-    command: sh
-    script: |
-      echo "Hello, world!"
-EOF
```
chrishoage, line 58: This could never work as when a user binds over `/home/dagu/.config/dagu` it will mount over this file at this location. 

If there is a desire to have an automatically populated hello-world dag, it would have to happen during runtime on the first boot.

entrypoint.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# If DAGU_HOME is not set, try to guess if the legacy /home directory is being
+# used. If so set the HOME to /home/dagu. Otherwise force the /config directory
+# as DAGU_HOME
+if [ -z "$DAGU_HOME" ]; then
+  if [ -d /home/dagu/.config/dagu ]; then
+    echo "WARNING: Using legacy /home/dagu directory. Please consider moving to /config"
+    usermod -d /home/dagu dagu
+    chown $PUID:$PGID -R /home/dagu
+  else
```
chrishoage, line 39: This check isn't the best, but it should cover _most_ cases. 

We mirror what the "init" part of the original docker-compose stack did by chowning the directory.

docs/source/docker-compose.rst
```diff
@@ -1,48 +1,28 @@
...
+            environment:
+            - DAGU_PORT=8080 # optional. default is 8080
+            - DAGU_TZ=Asia/Tokyo
+            - DAGU_BASE_PATH=/dagu # optional. default is /
+            - PUID=1000 # optional. default is 1000
+            - PGID=1000 # optional. default is 1000
+            - DOCKER_GID=999 # optional. default is -1 and it will be ignored
+            volumes:
+            - dagu_config:/config
+            - /var/run/docker.sock:/var/run/docker.sock # optional. required for docker in docker
```
chrishoage, line 25: I wasn't sure if this was worth including here. It maybe the comment should mention the `socat` alternative in the docs?

Dockerfile
```diff
@@ -7,8 +7,8 @@ WORKDIR /app
 COPY ui/ ./
 
 RUN rm -rf node_modules; \
-    yarn install --frozen-lockfile --non-interactive; \
-    yarn build
+  yarn install --frozen-lockfile --non-interactive; \
+  yarn build
```
chrishoage, line 11: My editor kept formatting this - happy to revert if it is undesired.

Dockerfile
```diff
@@ -32,35 +32,28 @@ ARG USER="dagu"
...
 COPY --from=go-builder /app/bin/dagu /usr/local/bin/
+COPY ./entrypoint.sh /entrypoint.sh
+
+# Create user and set permissions
+RUN apk update && \
+  apk add --no-cache su-exec shadow tzdata && \
+  addgroup -g ${USER_GID} ${USER} && \
+  adduser ${USER} -h /config -u ${USER_UID} -G ${USER} -D -s /bin/ash && \
+  chown -R ${USER}:${USER} /config && \
+  chmod +x /entrypoint.sh;
```
chrishoage, line 44: Instead of `sudo` we rely on `su-exec` and allowing the user to run scripts in `/etc/custom-init.d` 

Ideally `sudo` would not be added to docker containers, since it is easy to get root if you need it anyway (through docker exec)

chrishoage (commented):
Like I said in the body of the PR - this is a really big change, so it's okay if it takes some time to review it!

docs/source/docker-compose.rst
yohamta0: Thank you for the great suggestion! I believe this would be a valuable addition to include here. I think mentioning socat would be nice, but I also like the current focus on the docker-compose configuration. So, let's leave it as it is for now.

yohamta (approved):
Everything looks great! Thank you so much for improving the Dockerfile, docker-compose.yaml, and the documentation! These look much cleaner now :)

PR #728, opened 2024-11-29
chore: add support for devcontainers

jonnochoo:
I am developing on Windows, so it makes it easier to develop using devcontainers for this project.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
jonnochoo, line 1: this converts line endings to LF (useful for when on Windows)

yohamta (approved):
✨✨✨LGTM✨✨✨🚀 Thanks!

PR #729, opened 2024-11-29
add support for default page

jonnochoo:
Related to https://github.com/dagu-org/dagu/issues/721

This allows the max items returned on the dashboard page to be adjusted according to the number of DAGs.

It feels a little hacky, so open to further thoughts? 

What if we have an new API to aggregate the results? (to provide the values to the highlighted in green)
image

Or we filter the results based on time? (in red, sorry for the bad drawings) eg only show those that ran "today"

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -32,33 +32,34 @@ import (
 
 // Config represents the configuration for the server.
 type Config struct {
-	Host               string         // Server host
-	Port               int            // Server port
-	DAGs               string         // Location of DAG files
-	Executable         string         // Executable path
-	WorkDir            string         // Default working directory
-	IsBasicAuth        bool           // Enable basic auth
-	BasicAuthUsername  string         // Basic auth username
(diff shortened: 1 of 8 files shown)


yohamta:
Looking great, thank you!

> It feels a little hacky, so open to further thoughts?
> What if we have an new API to aggregate the results? (to provide the values to the highlighted in green)

Sounds good! However, the original reason for limiting the results was to mitigate performance issues, so I think it’s better to keep the limit for now. We can consider adding an aggregate API after we’ve improved the storage layer.

> Or we filter the results based on time? (in red, sorry for the bad drawings) eg only show those that ran "today"

Yes, we actually already have this option. It shows only today’s status when the corresponding setting is enabled.
https://github.com/dagu-org/dagu/blob/81f6057416bb3d0863c2c2a837aabe4b0f228d26/internal/config/config.go#L55

yohamta (approved):

PR #777, opened 2025-01-08
digraph: Allow a map for steps instead of a list

yohamta:
Example:
```yaml
steps:
  step1:
    command: echo hello
  step2:
    command: echo world
  step3:
    command: echo hello world
    depends:
      - step1
      - step2
```


ghansham:
Jusrlt wanted to know why switched from list to map?

PR #791, opened 2025-01-13
Fixes itemized list in Contribution Guide

kennethpjdyer:
## Description

Fixes a rendering issue in the docs.

The Contribution Guide has an itemized list under the [Code Structure](https://dagu.readthedocs.io/en/latest/contrib.html#code-structure) heading. The final item (which covers `internal`) is rendering incorrectly.  This is due to the lack of an empty line between the item and the first sub-item, causing Docutils to parse the sub-items as part of the paragraph rather than as distinct block-level nodes.


yohamta:
Thanks for improving the documentation!

PR #855, opened 2025-02-22
fix: show legacy path warnings

arky:
Show a legacy configuration page warning if users are using ``.dagu``.

Adds a note in configuration docs.


arky:
Closing this PR as this needs refactoring of core by @yohamta

PR #862, opened 2025-02-24
docs: note block in execute the DAG of QuickStart

Lewiscowles1986:
See https://github.com/dagu-org/dagu/issues/841

Just a note-block for now.

I noticed the other `note:` was not a note block, but managed to find an example in rest.rst of how this project was setup.

When testing I did use python 3.11.9 instead of 3.7.16 (which I think was old in 2018 😄 )

I can add changing the python version to this, but it might be better to have in it's own PR and issue so others can discuss.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -55,6 +55,9 @@ Go to the ``SPEC`` Tab and hit the ``Edit`` button. Copy & Paste the following Y
 4. Execute the DAG
 -------------------
 
+.. note::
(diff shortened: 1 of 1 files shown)


Lewiscowles1986:
Apologies, found a unexpected silent failure to highlight docker-compose.yml

Screenshot 2025-02-24 at 19-45-45 Quick Start Guide — Dagu documentation

yottahmd (approved):
Oh, I had no idea the note block existed. Looks awesome! Yeah, let’s update the Python version in a separate PR!

PR #865, opened 2025-02-25
docs: bump python version for docs to 3.11.9

Lewiscowles1986:
WiP: bumping python version (checking it works with pipelines and others)


Lewiscowles1986:
See #862 comments about this. Open to making changes to help others.

Lewiscowles1986:
Do you think it would be a good idea for me to trigger building docs in pipelines so that we can match the python versions? I think 3.7 will still work, it's just very old and I don't know if it will compile or if it would cause dev friction on my machine.

Lewiscowles1986:
I could bump to 3.12 to bring in-line with [.readthedocs.yaml](https://github.com/dagu-org/dagu/blob/3e665baf632f2c7e5b6be85d01a6fcce07973a26/.readthedocs.yaml)

yottahmd (approved):
Thanks a lot for upgrading the Python version! 🚀🚀🚀 I really appreciate it!

PR #874, opened 2025-03-10
docs: fixed incorrect command syntax in slack notification example

david-waterworth:
Closes https://github.com/dagu-org/dagu/issues/873#issuecomment-2708754362


yottahmd (approved):
Looking great! Thank you so much for improving the document.

Issue #875, opened 2025-03-10
Docker executor: support `platform` config option


bremyozo:
Docker server API Version 1.41+ has support to specify the architecture which should be used to run containers, e.g.: linux/amd64. This will allow to run multiplatform/multiarch containers

This applies to podman too after #815

yottahmd:
Sounds great to me. Let's add support for the `platform` config option. Would you be interested in implementing this? Here's the file to update:
https://github.com/dagu-org/dagu/blob/main/internal/digraph/executor/docker.go

PR #898, opened 2025-03-29
feat: allow specify container name and image platform

### LINKED ISSUE #853: Allow supplying container name when creating a container with docker executor run
Hello. This functions also take a name for the newly created container. Could you add an option to supply that option ?
https://github.com/dagu-org/dagu/blob/7bbb13df18f2e1a3e39415c6a841c84e51a82f39/internal/digraph/executor/docker.go#L145

Thank you very much for this project :D

vnghia:
Closes #853 #875


internal/digraph/executor/docker.go
```diff
@@ -108,14 +109,17 @@ func (e *docker) Run(ctx context.Context) error {
...
+		if e.containerName == "" {
+			return fmt.Errorf("containerName must be set in exec mode")
+		}
 		return e.execInContainer(ctx, cli, args)
 	}
 
 	// New container creation logic
 	if e.pull {
-		reader, err := cli.ImagePull(ctx, e.image, image.PullOptions{})
+		reader, err := cli.ImagePull(ctx, e.image, image.PullOptions{Platform: e.platform})
```
vnghia: We could make this a little more robust by checking if the locally available image targets the correct platform or not and attempt to pull the correct one. WDYT @yottahmd ?

yohamta0: Yes, I think this is a great idea. Could we do that?

vnghia:
Hi @yottahmd, could you take a look at this one ? This is my very first experience with go so could you tell where should I write tests and also if I follow the good practice in go

internal/digraph/executor/docker.go
```diff
@@ -371,16 +411,36 @@ func newDocker(
 		}
 	}
 
-	pull := true
-	if p, ok := execCfg.Config["pull"]; ok {
+	pull := ""
```
yohamta0: I'm not entirely sure, but could we use `bool` or `*bool` here instead?
If we go with `string`, it might break compatibility with older versions of Dagu

internal/digraph/executor/docker.go
```diff
@@ -108,28 +111,65 @@ func (e *docker) Run(ctx context.Context) error {
...
 		return e.execInContainer(ctx, cli, args)
 	}
 
-	// New container creation logic
-	if e.pull {
-		reader, err := cli.ImagePull(ctx, e.image, image.PullOptions{})
+	// If platform is not provided, imageReference will be set to the input image name
+	// else, it will be set to the locally available imageId whose platform matches the input platform
+	// or an empty string to signify that we don't have the right image locally.
+	imageReference := ""
```
yohamta0: Why don't we initialize the `imageReference` with `e.image`?

internal/digraph/executor/docker.go
```diff
@@ -108,28 +111,65 @@ func (e *docker) Run(ctx context.Context) error {
...
+
+	// Only inspect image if pull mode is not always
+	if e.pull != "always" {
```
yohamta0: Could we extract this into a separate method? It’s starting to get a bit complicated.

yottahmd (commented):
@vnghia Sorry for the delay in reviewing this.
It looks great overall! I've left a few comments and questions, please take a look when you have a moment.

vnghia:
Thanks for your review @yottahmd.

Well what I have in mind is a little bit complicated so I will document my idea below.

---

For container name:
- If we have `image` and `containerName` => create a new container with `image` and `containerName`.
- If we only have `image` => create a new container with `image` and random name.
- If we only have `containerName` => exec into the container whose name is `containerName`.
- If we don't have neither of them => error

---

For image pull, basically I am trying to mimic this [option](https://docs.docker.com/reference/cli/docker/container/run/#pull).

> - missing (default): Pull the image if it was not found in the image cache, or use the cached image otherwise.
> - never: Do not pull the image, even if it's missing, and produce an error if the image does not exist in the image cache.
> - always: Always perform a pull before creating the container.

In current Dagu implementation, `pull` true corresponds to `always` and `pull` false corresponds to `never` and it is missing the `missing` one (XD) which is the sensible default IMO.

To implement this, I change `pull` to a string (We could just read `pull` as a boolean or a string and map it to the above values to avoid a breaking change). On a high-level, it will:
- set an empty `imageReference`.
- if pull is not `always`:
  - if `platform` is provided => list all the cached images with the same name and check if we have the correct platform cached or not. If yes, set `imageReference` to that image id.
  - if `platform` is not provided => set `imageReference` to the supplied `image`.
- then if pull is `always` or `imageReference` is empty (this means that the image is not yet available in the local cache) => pull the `image` with an optional `platform` and set `imageReference` to `image`.
- if `imageReference` is still empty => error
- if `imageReference` is not empty => create a new container with `imageReference`.

However, while trying to do this, I found it was hard to implement correctly and may slow down Dagu because there are a lot more things to do. So I guess it's up to you to see if this feature is a good addition to Dagu or we should just keep the current behavior.

yottahmd:
Thank you, @vnghia for such the detailed explanation! I think it's a valuable addition.

Here’s how I’m thinking of handling the pull value:
- User can provide a bool or string (`always` `never`, `newer`, or `missing`)
- If the input is truthy (use `strconv.ParseBool`), treat it as `always`
- If it's is `nil`, treat it as `missing`
- If it's falsy (use `strconv.ParseBool`), treat it as `never`
- Otherwise, it returns an error
- Add a type `PullPolicy` and constants (e.g., `type PullPolicy string`)

I was referencing the [Podman's document](https://docs.podman.io/en/latest/markdown/podman-run.1.html).

You could handle this using a [type switch](https://go.dev/tour/methods/16)

What do you think?

vnghia:
I think we could do it. I will continue to work on this in the next comming weeks. Thank you!

yottahmd:
Thanks a lot! We could also add a `docker_test.go` in this same directory for things like the `PullPolicy` parsing method.
Feel free to add the [integration tests](https://github.com/dagu-org/dagu/blob/3bd0d3f948c6dd3e28e55a6ecaa0ddbf10806dde/internal/integration/dag_test.go#L20) if you’re up for it. Here is the directory to add DAG files for integration tests: https://github.com/dagu-org/dagu/tree/main/internal/testdata/integration

vnghia:
@yottahmd I don't add the integration test because it requires a lot of setting up to test the pull policy and platform properly (registry authentication to avoid rate-limited pulling, docker image isolation for each test, multi-platform supported runner). We can return to this later

internal/digraph/executor/docker.go
```diff
@@ -45,12 +49,39 @@ steps:
...
+func parsePullPolicy(s string) PullPolicy {
+	switch s {
+	case "always":
+		return Always
+	case "never":
+		return Never
+	default:
+		return Missing
+	}
+}
```
yohamta0: Can we make this a map to simplify?
```suggestion
var pullPolicyMap = map[string]PullPolicy{
	"always":  Always,
	"missing": Missing,
	"never":   Never,
}
```

internal/digraph/executor/docker.go
```diff
@@ -45,12 +49,39 @@ steps:
 ```
 */
 
+type PullPolicy int
+
+const (
+	Always PullPolicy = iota
+	Never
+	Missing
```
yohamta0: Can we add prefixes?
```suggestion
	PullPolicyAlways PullPolicy = iota
	PullPolicyNever
	PullPolicyMissing
```

internal/digraph/executor/docker.go
```diff
@@ -373,16 +463,51 @@ func newDocker(
...
+			}
+
+			boolPull, err := strconv.ParseBool(value)
+			if err != nil {
+				pull = parsePullPolicy(value)
+			} else {
+				pull = boolToPullPolicy(boolPull)
+			}
+		}
+	}
```
yohamta0: Can we make this into a separate function something like this?
```go
func (e *docker) parsePullPolicy(ctx context.Context, raw any) (PullPolicy, error) {
	switch v := raw.(type) {
	case string:
		value, err := digraph.EvalString(ctx, v)
		if err != nil {
			return Missing, fmt.Errorf("failed to evaluate pull policy: %w", err)
		}

		// Try to parse the string as a pull policy
		p, ok := pullPolicyMap[value]
		if ok {
			return p, nil
		}

		// If the string is not a valid pull policy, try to parse it as a boolean
		b, err := strconv.ParseBool(value)
		if err != nil {
			return Missing, fmt.Errorf("failed to evaluate pull policy: %w", err)
		}
		return boolToPullPolicy(b), nil

	case bool:
		return boolToPullPolicy(v), nil

	default:
		return Missing, fmt.Errorf("invalid pull policy type: %T", raw)
	}
}

```

yottahmd (commented):
Looking great! I've left a few minor comments. Could you take a look and address them when you get a chance?

vnghia:
Thank you for your suggestion. Please review again @yottahmd !

vnghia:
I also created #909 that should ease the implementation of testing and also the `newer` policy.

yottahmd (approved):
LGTM🚀🚀🚀 Thank you very much for the amazing work!

Issue #792, opened 2025-01-14
Support for Retry on Custom Exit Codes


yottahmd:
It would be helpful to have a feature that allows retrying a task based on custom exit codes. This feature would specify certain non-zero exit codes where retries should occur, while other non-zero exit codes would cause the task to fail immediately.

Example configuration:
```yaml
steps:
  - name: retryable task
    command: main.sh
    retryPolicy:
      limit: 3
      intervalSec: 5
      exitCode: [1, 2]  # Retries if the exit code is 1 or 2; otherwise, the task fails.
```

yottahmd:
It's now implemented in v1.16.8. Thank you @kriyanshii !

PR #902, opened 2025-04-17
added support for custom exit codes on retry #792

kriyanshii:
# Add Custom Exit Code Support for Retry Policy

## Description
This PR adds support for specifying custom exit codes that should trigger task retries. Previously, the retry policy would retry on any non-zero exit code. Now users can specify exactly which exit codes should trigger retries, while other non-zero exit codes will cause the task to fail immediately.

## Changes
1. Added `exitCode` field to the retry policy schema in `schemas/dag.schema.json`
2. Updated `RetryPolicy` struct in `internal/digraph/step.go` to include `ExitCodes` field
3. Added `ExitCode` field to `retryPolicyDef` struct in `internal/digraph/spec.go`
4. Updated builder in `internal/digraph/builder.go` to handle the new `exitCode` field
5. Enhanced exit code handling in `internal/digraph/scheduler/scheduler.go`:
   - Added robust exit code extraction from different error types
   - Added support for parsing exit codes from error strings
   - Added handling for signal terminations
   - Added detailed debug logging for better troubleshooting
   - Improved retry policy decision logic

## Example Usage
```yaml
steps:
  - name: retryable task
    command: main.sh
    retryPolicy:
      limit: 3
      intervalSec: 5
      exitCode: [1, 2]  # Retries if exit code is 1 or 2; otherwise fails immediately
```

## Backward Compatibility
- If no `exitCode` is specified, the behavior remains unchanged (retries on any non-zero exit code)
- All existing retry policy configurations will continue to work as before

## Testing
- Added comprehensive error handling for different types of process terminations
- Added detailed debug logging to track exit code determination and retry decisions
- Maintained backward compatibility with existing retry policy configurations

## Benefits
- More precise control over which failures should trigger retries
- Better error handling and logging for troubleshooting
- Maintains backward compatibility with existing configurations

### DIFF
--- a/internal/digraph/scheduler/scheduler.go
+++ b/internal/digraph/scheduler/scheduler.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"runtime/debug"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


internal/digraph/scheduler/scheduler.go
```diff
@@ -193,12 +196,71 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, done c
...
+									exitCodeFound = true
+									logger.Debug(ctx, "Process terminated by signal", "error", errStr)
+								}
+							}
+
+							if !exitCodeFound {
+								logger.Debug(ctx, "Could not determine exit code", "error", execErr, "errorType", fmt.Sprintf("%T", execErr))
+								// Default to exit code 1 if we can't determine the actual code
+								exitCode = 1
+							}
```
yohamta0: Can we define a separate method for this?

internal/digraph/scheduler/scheduler.go
```diff
@@ -193,12 +196,71 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, done c
...
+										shouldRetry = true
+										break
+									}
+								}
+								logger.Debug(ctx, "Checking retry policy", "exitCode", exitCode, "allowedCodes", node.retryPolicy.ExitCodes, "shouldRetry", shouldRetry)
+							} else {
+								// If no exit codes specified, retry for any non-zero exit code
+								shouldRetry = exitCode != 0
+								logger.Debug(ctx, "Using default retry policy", "exitCode", exitCode, "shouldRetry", shouldRetry)
+							}
```
yohamta0: Can we define a `ShouldRetry(exitCode int)` method on `RetryPolicy` struct to handle this?

internal/digraph/scheduler/scheduler.go
```diff
@@ -193,12 +196,71 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, done c
...
+								logger.Info(ctx, "Step execution failed. Retrying...", "step", node.data.Name(), "error", execErr, "retry", node.data.GetRetryCount(), "exitCode", exitCode)
+								time.Sleep(node.retryPolicy.Interval)
+								node.data.SetRetriedAt(time.Now())
+								node.data.SetStatus(NodeStatusNone)
+							} else {
+								// finish the node with error
+								node.data.SetStatus(NodeStatusError)
+								node.data.MarkError(execErr)
+								sc.setLastError(execErr)
+							}
```
yohamta0: I thought we could simplify this. What do you think?
```go
	if shouldRetry {
		// retry
		node.data.IncRetryCount()
		logger.Info(ctx, "Step execution failed. Retrying...", "step", node.data.Name(), "error", execErr, "retry", node.data.GetRetryCount(), "exitCode", exitCode)
		time.Sleep(node.retryPolicy.Interval)
		node.data.SetRetriedAt(time.Now())
		node.data.SetStatus(NodeStatusNone)
		break
	} 
	fallthrough
```

yottahmd (commented):
Thanks, it looks great! I've left a few comments.

internal/digraph/scheduler/scheduler.go
kriyanshii: Yess sure.

internal/digraph/scheduler/scheduler.go
kriyanshii: Will try it

kriyanshii:
@yottahmd resolved all the comments.

internal/digraph/scheduler/scheduler.go
```diff
@@ -587,3 +587,60 @@ func (sc *Scheduler) isSucceed(g *ExecutionGraph) bool {
...
+	shouldRetry = node.retryPolicy.ShouldRetry(exitCode)
+	logger.Debug(ctx, "Checking retry policy", "exitCode", exitCode, "allowedCodes", node.retryPolicy.ExitCodes, "shouldRetry", shouldRetry)
+
+	if !shouldRetry {
+		// finish the node with error
+		node.data.SetStatus(NodeStatusError)
+		node.data.MarkError(execErr)
+		sc.setLastError(execErr)
+		return false
+	}
```
yottahmd, line 637: Would it be possible to reuse the same logic in the switch statement below? In some cases, a step should be marked as successful even if it fails.

https://github.com/dagu-org/dagu/blob/2ab2fab4636f69f0dc1e7c0d913673f1f143de0f/internal/digraph/scheduler/scheduler.go#L203-L214

Perhaps we could check the return value from the `handleNodeRetry` fucntion and use fallthrough when the return value is false. What do you think?

yohamta0, line 637: @kriyanshii 
Would it be possible to reuse the same logic in the switch statement below? In some cases, a step should be marked as successful even if it fails.

https://github.com/dagu-org/dagu/blob/2ab2fab4636f69f0dc1e7c0d913673f1f143de0f/internal/digraph/scheduler/scheduler.go#L203-L214

Perhaps we could check the return value from the `handleNodeRetry` fucntion and use fallthrough when the return value is false. What do you think?

Something like this:
```diff
case node.retryPolicy.Limit > node.data.GetRetryCount():
	if sc.handleNodeRetry(ctx, node, execErr) {
		continue ExecRepeat
	}
+	fallthrough
```

internal/digraph/scheduler/scheduler.go
```diff
@@ -587,3 +587,60 @@ func (sc *Scheduler) isSucceed(g *ExecutionGraph) bool {
...
+			exitCode = -1
+			exitCodeFound = true
+			logger.Debug(ctx, "Process terminated by signal", "error", errStr)
+		}
+	}
+
+	if !exitCodeFound {
+		logger.Debug(ctx, "Could not determine exit code", "error", execErr, "errorType", fmt.Sprintf("%T", execErr))
+		// Default to exit code 1 if we can't determine the actual code
+		exitCode = 1
```
yohamta0, line 625: I thought we can initialize the exitCode with `1` at declaration. WDYT?

internal/digraph/scheduler/scheduler.go
kriyanshii: Yes, that's a good observation! We can reuse the success marking logic by using fallthrough when handleNodeRetry returns false. This would make the code more DRY (Don't Repeat Yourself) and ensure consistent behavior.

yottahmd (commented):
It looks great! I've added some nit comments.

yottahmd:
@kriyanshii Would you mind updating the documents to reflect this change? Thanks a lot!

kriyanshii:
hey can you check it? i have updated docs. do i need to update anything else?

yottahmd (approved):
Thanks for updating the document and JSON schema! LGTM 🚀🚀🚀

Issue #905, opened 2025-04-19
Incomplete text in dag rendering component


ghansham:
The latest mermaid version is impressive showing animated steps when dag execution is in progress but our observation is that at times, the text within the step box is not completely rendered. It does not happen always. May be has something to do with the browser zooming.

Another point the new dag rendering makes rendering of large dags (which have lot of parallel processes running, whose rendering is very wide) a bit obscure. The older mermaid version was much better. May be the zooming facility recently committed will be of some help.

We can share some snapshots for same.

yottahmd:
Yes, I noticed that too. Let me remove the icon and revert the styling to the old version!

ghansham:
You can close this issue

PR #906, opened 2025-04-19
[#905] DAG visualization updates

yottahmd:
Addressing: #905 

**Changes**
- Removed the animation icons on graph to avoid the incomplete text issue.
- Fix an issue where graph position and zooming are reset when graph is updated while running the DAG.

**Updated version of the graph visualization**
image


yottahmd:
@ghansham What do you think?

ghansham:
I would say this looks way better. 👍

Issue #939, opened 2025-05-14
Enhance `RepeatPolicy` field


yottahmd:
Document about repeat policy:
https://dagu.readthedocs.io/ja/latest/yaml_format.html#repeat-steps

We would like to enhance this functionality as follows:
- Add condition to repeatPolicy until a certain condition is satisfied
  Add `condition` setting to the step

```
step:
- name: step1
  command: xxx
  repeatPolicy:
    condition: "foo"
    expected: "bar"
    intervalSec: 30
```

ghansham:
Custom exir code documentation can also be added in retry clause

yottahmd:
Thanks! Updated the design.

thefishhat:
Does `ExitCode` here mean repeat **until** command exit code matches, or repeat **while** command exit code matches?

yottahmd:
Hi @thefishhat, it continues to repeat **while** the command's exit code matches.

thefishhat:
Thanks @yottahmd ! Two more questions about the design:
* Should `3. Command (new)` condition also be covered in backticks, since it is also a shell command - similar to 2. This makes the spec more consistent and the code is easier to implement.
*  For `1. Simple string matching (new)` I assumed that the output of command is matched against `expected`. But what is `condition` used for in this case? Right now in my draft implementation it remains ignored for this use case.

yottahmd:
@thefishhat Thank you for raising these thoughtful points!

* I agree the current wording can be confusing. I was planning to align the new behavior with the existing [Precondition](https://dagu.readthedocs.io/en/latest/schema.html#id6) field:

  1. When **both** `condition` and `expected` are set, treat the pair as a *text-match* test.
  2. When **only** `condition` is set, treat it as a *command* (no back-tick expansion). Keep repeating the step as long as the command exits with `0`.

  The current `Precondition` implementation is here. What do you think?
  [https://github.com/dagu-org/dagu/blob/51102e5d3c80cf66b7c8771d3b1df4c47c4df657/internal/digraph/scheduler/condition.go#L60](https://github.com/dagu-org/dagu/blob/51102e5d3c80cf66b7c8771d3b1df4c47c4df657/internal/digraph/scheduler/condition.go#L60)

* I was thinking the `condition` will usually reference an environment variable. So it needs to be used in evaluation. Example:

```yaml
steps:
  - name: repeat
    command: "python main.py"
    output: "RESULT"
    repeatPolicy:
      condition: "$RESULT"
[...]

PR #953, opened 2025-05-24
feat: enhance repeat policy with condition, expected output and exit code checks

thefishhat:
# What does this PR do?
This PR enhances the `RepeatPolicy` field to allow a step to be repeated until a certain condition is satisfied.

# Why is this needed?
Addresses issue #939.

# What changed?
* Adds new advanced repeat logic to workflow steps, allowing them to repeat until a shell condition is met, output matches an expected value, or a specific exit code is returned.
* Updates the configuration, scheduler logic, documentation, tests, and schema to support these new repeat conditions.

# How was this tested?
* Written automated tests
* Manually, using the following DAGs:
1. shell command condition
```steps:
  - name: repeat-until-string-match
    command: echo foo
    output: RESULT
    repeatPolicy:
      condition: "$RESULT"
      expected: "foo"
      intervalSec: 1
```
  2. string match condition
```
steps:
  - name: repeat-until-string-match
      command: echo foo
      output: RESULT
      repeatPolicy:
        condition: "$RESULT"
        expected: "foo"
        intervalSec: 30
```

3. while command returns exit code
```
steps:
  - name: repeat-while-exitcode-matches
      command: test -f /tmp/flag
      repeatPolicy:
        exitCode: [0]
        intervalSec: 5
```

4. while condition returns status code 0
```
steps:
  - name: repeat-until-condition-exits-non-zero
      command: echo "checking"
      repeatPolicy:
        condition: "test -f /tmp/flag"
        intervalSec: 1
```

### DIFF
--- a/internal/digraph/scheduler/scheduler_test.go
+++ b/internal/digraph/scheduler/scheduler_test.go
@@ -794,8 +794,16 @@ func TestScheduler(t *testing.T) {
 
 		// This step will repeat until the file contains 'ready'
 		file := filepath.Join(os.TempDir(), fmt.Sprintf("repeat_test_%s.txt", uuid.Must(uuid.NewV7()).String()))
-		_ = os.Remove(file)
-		defer os.Remove(file)
+		err := os.Remove(file)
+		if err != nil && !os.IsNotExist(err) {
+			require.NoError(t, err)
+		}
+		defer func() {
+			err := os.Remove(file)
+			if err != nil && !os.IsNotExist(err) {
+				require.NoError(t, err)
+			}
+		}()
 		graph := sc.newGraph(t,
 			newStep("1",
 				withCommand(fmt.Sprintf("cat %s || true", file)),
@@ -826,8 +834,16 @@ func TestScheduler(t *testing.T) {
 		sc := setup(t)
 		// This step will repeat until the file exists
 		file := filepath.Join(os.TempDir(), fmt.Sprintf("repeat_exit0_%s", uuid.Must(uuid.NewV7()).String()))
-		_ = os.Remove(file)
-		defer os.Remove(file)
+		err := os.Remove(file)
+		if err != nil && !os.IsNotExist(err) {
+			require.NoError(t, err)
+		}
+		defer func() {
+			err := os.Remove(file)
+			if err != nil && !os.IsNotExist(err) {
+				require.NoError(t, err)
+			}
+		}()
 		graph := sc.newGraph(t,
 			newStep("1",
 				withCommand("echo hello"),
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -262,8 +262,8 @@ build-ui:
 	@echo "${COLOR_GREEN}Building UI...${COLOR_RESET}"
 	@cd ui; \
 		pnpm install; \
-		NODE_OPTIONS="--max-old-space-size=8192" pnpm webpack --config webpack.dev.js; \
-		pnpm webpack --config webpack.prod.js
+		NODE_OPTIONS="--max-old-space-size=8192" pnpm webpack --config webpack.dev.js --progress --color; \
+		pnpm webpack --config webpack.prod.js --progress --color
```
thefishhat, line 266: nit: added this unrelated Makefile change because my system is slow and I needed to see it's not blocked while building

docs/source/examples.rst
```diff
@@ -267,3 +267,42 @@ Customizing Signal Handling on Stop
...
+    - name: repeat-legacy
+      command: echo 'hello'
+      repeatPolicy:
+        repeat: true
+        intervalSec: 60
+
+- ``condition``: Command or expression to evaluate after each run.
+- ``expected``: Value or regex to match the output of ``condition``.
+- ``exitCode``: Integer or list of integers; repeat if the last command exits with one of these codes.
+- ``repeat``/``intervalSec``: Legacy mode; repeat unconditionally at the given interval.
```
thefishhat: q: do we want this to be referred to as `Legacy mode`?

yohamta0: Great question. Personally, I don’t think there’s a need to mark `intervalSec` as Legacy since it can’t be achieved using condition. When `condition` is specified, I believe `intervalSec` should still be considered. 

Regarding `repeat: true`, I believe that `condition: true` can accomplish the same thing, so we can treat it as `Legacy` while mentioning this point as a note. What do you think?

yottahmd:
@thefishhat Oh sorry, could you take a look at the CI failure when you have a chance?

thefishhat:
@yottahmd Apologies for that! I've fixed the linting errors. I think the workflow should pass now. The only uncertainty is that I still get failures running tests locally.
```
internal/cmd TestServerCommand/StartServer
internal/cmd TestServerCommand
internal/cmd TestStartAllCommand/StartAll
internal/cmd TestStartAllCommand
internal/integration TestServer_BasePath
internal/integration TestSkippedPreconditions
internal/integration TestComplexDependencies
internal/integration TestProgressingNode
internal/integration TestServer_RemoteNode
internal/integration TestServer_RemoteNode/#00
internal/integration TestServer_RemoteNode//dagu
internal/integration TestDAGExecution
```

They seem to be unrelated to this change (and they also fail for me on `next` branch as well), so it'll probably work now 👍 .

yottahmd:
LGTM! 🚀🚀🚀 Thank you so much!

Issue #970, opened 2025-05-31
race condition while using retry functionality


kriyanshii:
While running this DAG it seems race condition happens to open the logfile.

YAML to reproduce the bug.
```
  steps:
    - name: api call
      command: sh main.sh 
      retryPolicy:
        limit: 3
        intervalSec: 1
        exitCode: [0, 1]  # Retry on rate limit and service unavailable errors

```

![Image](https://github.com/user-attachments/assets/a7f40c6e-92d9-44a2-8dfd-c23621063f0b)

ghansham:
@yottahmd I like the part that error column is showing a message and not simply the exit code. Can we get the error message generated by a step in this column. For a non zero exit code returned by a step, it always shows the error code.

yohamta0:
I think that's a good idea! I'm wondering what we should do if the error log is large. Perhaps we could try retrieving a maximum of N bytes, and if it exceeds that limit, simply display 'Error log too large to display.'

ghansham:
we can try displaying last line (or last N bytes as you recommend) of the stderr stream. And the link to open the stderr stream can be moved here from its current position. And mention this in the documentation that dagu suggests user to write error messages to be written to stderr stream. Or we can include in dag definition or admin.yaml an entry to indicate which stream to be used to capture error messages (stdout or stderr). That should be enough. 
The final entry in error column may look like `err_code (last line of error message <errstream_link>)`

kriyanshii:
back to this issue. my laptop died so was afk for weeks. now back with the bang....................

yohamta0:
Thank you very much! Apologies for delayed comment, but the issue of race condition might have been resolved at some point now. I'm sorry for the PR unmerged.

PR #971, opened 2025-05-31
solving #970 and updating the doc regarding wrong mention of exitCode

kriyanshii:
# Fix: Improve Log File Handling During Retries

## Problem
When a DAG step is retried, the log files were not being properly handled, leading to "file already closed" errors. This occurred because:
1. The same log file was being reused across retries
2. File handles were being closed after the first attempt
3. Subsequent retries couldn't write to the closed file

## Solution
Modified the log file handling in `internal/digraph/scheduler/node.go` to:
1. Create unique log files for each retry attempt
2. Append retry count to log filenames (e.g., `.retry1`, `.retry2`)
3. Ensure proper file handle management during retries

### Changes
- Updated `Setup` method to include retry count in log filenames
- Added retry suffix format: `.retryN` where N is the retry count
- Maintained backward compatibility for non-retry cases

## Testing
- Verified that retries now create separate log files
- Confirmed no "file already closed" errors during retries
- Tested with multiple retry attempts
- Ensured backward compatibility for non-retry cases

## Impact
- Improved reliability of retry mechanism
- Better debugging capabilities with separate logs for each retry
- Clearer history of retry attempts
- No changes to existing functionality for non-retry cases

## Related Issues
Closes #[970]

### DIFF
--- a/docs/reference/yaml.md
+++ b/docs/reference/yaml.md
@@ -121,6 +121,7 @@ schedule:
 |-------|------|-------------|---------|
 | `mailOn` | object | Email notification triggers | - |
 | `errorMail` | object | Error email configuration | - |
+| `infoMail` | object | Info email configuration | - |
 | `smtp` | object | SMTP server configuration | - |
 
 ```yaml
@@ -133,6 +134,12 @@ errorMail:
   to: oncall@example.com
   prefix: "[ALERT]"
   attachLogs: true
+  
+infoMail:
+  from: notifications@example.com
+  to: team@example.com
+  prefix: "[INFO]"
+  attachLogs: false
   
 smtp:
   host: smtp.gmail.com
--- a/docs/writing-workflows/examples.md
+++ b/docs/writing-workflows/examples.md
@@ -950,6 +950,195 @@ Send email alerts on events.
 
 </div>
 
+## Operations & Production
+
+<div class="examples-grid">
+
+<div class="example-card">
+
+### History Retention
+
+```yaml
+name: data-archiver
+histRetentionDays: 30    # Keep 30 days of history
+schedule: "0 0 * * *"     # Daily at midnight
+steps:
+  - name: archive-old-data
+    command: ./archive.sh
+  - name: cleanup-temp
+    command: rm -rf /tmp/archive/*
+```
+
+Control how long execution history is retained.
+
+<a href="/reference/yaml#data-fields" class="learn-more">Learn more →</a>
+
+</div>
+
+<div class="example-card">
+
+### Output Size Management
+
+```yaml
+name: log-processor
+maxOutputSize: 10485760   # 10MB max output per step
+steps:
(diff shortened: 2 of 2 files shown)


ghansham:
Retry should be happening only after last retry is complete. I am not sure if it is a race condition. If each retry step is closing log file, we should try to reopen it in append mode if it exists (in subsequent retries) or create it (in first retry). @yottahmd What was original behavior before this bug got in?

You can also attempt this solution:

if n.GetRetryCount() > 0 and exitCode is in the exitCodesList {
//don't close the log file at following line:
https://github.com/dagu-org/dagu/blob/9c2c1633a935133dd172fcacc8696fcca862b53f/internal/digraph/scheduler/node.go#L441
}

We need to pass on the exitCode returned from last retry to tearDown function. And this bug exists in 1.16.x

yohamta0:
Hi @ghansham, @kriyanshii — thanks for investigating this! 🙏

I think @ghansham is right. It seems like a bug that the log file gets closed on each retry. We should ensure that it stays open and doesn’t get closed with each retry
(It was the original behavior).

yohamta0:
I think we can just remove the code below. @kriyanshii Can you please update the PR?
https://github.com/dagu-org/dagu/blob/4dfbec53d0093333bce71cb3033a3541ef15dda5/internal/digraph/scheduler/scheduler.go#L252-L255

ghansham:
Very cool

kriyanshii:
@yottahmd I have updated the PR. Please review.

internal/digraph/scheduler/node.go
```diff
@@ -400,7 +400,14 @@ func (n *Node) Setup(ctx context.Context, logDir string, dagRunID string) error
 	safeName := fileutil.SafeName(n.Name())
 	timestamp := startedAt.Format("20060102.15:04:05.000")
 	postfix := stringutil.TruncString(dagRunID, 8)
-	logFilename := fmt.Sprintf("%s.%s.%s", safeName, timestamp, postfix)
+
+	// Add retry count to filename if this is a retry
+	retrySuffix := ""
+	if n.GetRetryCount() > 0 {
+		retrySuffix = fmt.Sprintf(".retry%d", n.GetRetryCount())
+	}
```
yohamta0: Question: Why do we need to create separate log files? I'm asking because if we create different log files, the older ones might become orphaned and won't be removed when Dagu deletes the DAG-run data.

yottahmd (commented):
Looks great! Thank you for updating the documentation as well. I’ve added a question about log file.

ghansham:
It's better to keep a single log file but at the start of every retry we should print some message mentioning retry no. X (Start Time: ......)

yohamta0:
Apologies for leaving the PR open. Thank you for all your hard work on this issue —I really appreciate it, even though it hasn’t been merged. If any other issues catch your eye, please feel free to pick them up!

Issue #982, opened 2025-06-04
[1.17.0-beta-fb] Reuse log file prefix to compress DAG-run file size for larger DAGs


yottahmd:
Feedback from @ghansham

PR #983, opened 2025-06-05
[#982] feat: reduce log file size with path placeholders in status JSON

yottahmd:
## Summary

This PR introduces a mechanism to reduce the size of status JSON files by replacing absolute log directory
paths with a placeholder (`__INTERNAL_LOG_DIR__`) during serialization. This significantly reduces file sizes
when DAG runs have many nodes, as the log directory path is typically repeated many times.

### Core Implementation
- Added custom `MarshalJSON` method to `DAGRunStatus` that replaces log paths with placeholders
- Modified `StatusFromJSON` to restore the actual paths when deserializing
- Added helper functions for clean path extraction and replacement

### Refactoring
- Renamed and reorganized node creation functions for better clarity.
- Removed duplicate teardown defer in scheduler
- Improved logging for DAG run starts with retry-specific information

### Testing
- Added comprehensive tests for MarshalJSON and StatusFromJSON
- Added test to verify original data is not modified during marshaling
- Added round-trip test to ensure data integrity
- Tests cover edge cases: nil nodes, handler nodes, missing log paths

## Benefits

1. **Reduced Storage**: Status files are significantly smaller, especially for DAGs with many steps
2. **Backward Compatible**: Existing status files without placeholders still work correctly

## Example

Before:
```json
{
  "nodes": [
    {"stdout": "/var/log/dagu/my-dag/step1.stdout.log"},
    {"stdout": "/var/log/dagu/my-dag/step2.stdout.log"},
    {"stdout": "/var/log/dagu/my-dag/step3.stdout.log"}
  ]
}

After:
{
  "nodes": [
    {"stdout": "__INTERNAL_LOG_DIR__/step1.stdout.log"},
    {"stdout": "__INTERNAL_LOG_DIR__/step2.stdout.log"},
    {"stdout": "__INTERNAL_LOG_DIR__/step3.stdout.log"}
  ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,6 @@ Dagu’s design emphasizes minimal external dependencies: it operates solely as 
   - [Via GitHub Releases Page](#via-github-releases-page)
   - [Via Homebrew (macOS)](#via-homebrew-macos)
   - [Via Docker (stable)](#via-docker-stable)
-  - [Via Docker (next preview)](#via-docker-next-preview)
   - [Quick Start](#quick-start)
 - [Building from Source](#building-from-source)
   - [Prerequisites](#prerequisites)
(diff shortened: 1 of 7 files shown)


ghansham:
Rather than adding custom marshaljson, can't we have an environment variable pointing to logdir and replace that when to be used.

yohamta0:
@Thanks for the comment! I've considered the approach, but it might be difficult to correctly evaluate the log file path outside of DAG-run context (for example, in API handler).

ghansham:
Oh.. thanks. I thought we need to evaluate actual path only when we are going to open file handler for stdout and stderr from the node structure.

PR #1006, opened 2025-06-11
feat: add resource management for DAGs and steps

yottahmd:
**Overview:**
- Adds CPU and memory resource management capabilities to Dagu
- Supports both DAG-level and step-level resource specifications
- Implements platform-specific enforcement (Linux cgroups v2, Unix rlimits)

**Resource Specification**
DAGs and steps can now specify resource requirements:
```yaml
resources:
  requests:
    cpu: "0.5"      # 0.5 CPU cores
    memory: "512Mi" # 512 MiB memory
  limits:
    cpu: "2"        # Max 2 CPU cores
    memory: "2Gi"   # Max 2 GiB memory
```

**Platform Support**
- Linux: Full support via cgroups v2 (preferred) or v1
- macOS/Unix: Limited support via rlimits (limits only)
- Windows: Not supported


yohamta0:
It looks like the spec still needs some polish, so I’ll close this at the moment.

Issue #1009, opened 2025-06-12
feat: Support a new `PartialSuccess` status


yottahmd:
Some DAGs finish their primary work yet still hit non-critical errors (e.g., skipped by `continueOn: failed`). At the moment we must label these runs as either `Success` or `Failed`, which loses important nuance. We would like to add a distinct `PartialSuccess` state to the run-status enum and surface it across the CLI, API, and Web UI.

Feedback by @ghansham ❤️

ghansham:
I think we can close this one

yohamta0:
Yes! Thank you. I'm closing this issue as it's already implemented.

PR #1011, opened 2025-06-12
feat: add Partial Success status to DAG runs

thefishhat:
## What does this PR do?
It implements a new `StatusPartialSuccess` enum value for DAG executions to handle scenarios where some nodes failed but were allowed to continue via continueOn.failure settings. The status indicates that primary work was completed despite some non-critical failures.

## Why is this needed?
Addresses issue #1009 . Previously, DAG runs would show as `Failed` even when critical steps succeeded and only optional/non-critical steps failed with `continueOn.failure: true`. This made it difficult to distinguish between complete failures and successful runs with minor issues, reducing visibility into workflow health and success metrics.

## What changed?
* Added `StatusPartialSuccess = 6` to scheduler `Status` enum with "partial success" string representation
* Implemented `isPartialSuccess()` method in scheduler to detect when DAG has both successful/skipped nodes AND error nodes that were allowed to continue
* Final status Detection: Updated main [Status()](vscode-file://vscode-app/c:/Users/Armand/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) method to check for PartialSuccess before returning StatusError, fixing the evaluation order
* Event Handlers: Updated event handler logic to treat PartialSuccess as success (triggers onSuccess handlers)
* API Schema: Updated v2 API YAML to include PartialSuccess status
* In the UI added `PartialSuccess` status chip component with orange styling
* DAG Runs UI: Added `PartialSuccess` option to status filter dropdown in `pages/dag-runs/index.tsx`
* Dashboard Metrics: Updated dashboard to include `PartialSuccess` count with orange CheckCircle icon

## How was this tested?
* Created test DAG with failing step that has continueOn.failure: true followed by successful steps
```yaml
name: Test Partial Success
description: Testing PartialSuccess status with continueOn.failure
steps:
  - name: step1
    command: echo "Step 1 successful"
    
  - name: step2_fail_but_continue
    command: exit 1
    continueOn:
      failure: true
    depends:
      - step1
      
  - name: step3
    command: echo "Step 3 successful despite step2 failure"
    depends:
      - step2_fail_but_continue
```

## Screenshots

image

### DIFF
--- a/internal/digraph/scheduler/scheduler.go
+++ b/internal/digraph/scheduler/scheduler.go
@@ -716,57 +716,49 @@ func (sc *Scheduler) isSucceed(g *ExecutionGraph) bool {
 	return true
 }
 
-// isPartialSuccess checks if the DAG completed with some failures that were allowed to continue.
-// This represents scenarios where primary work was completed but non-critical errors occurred.
+// isPartialSuccess checks if the DAG completed with some failures or skips that were allowed to continue.
+// This represents scenarios where primary work was completed but non-critical errors occurred or optional steps were skipped.
 func (sc *Scheduler) isPartialSuccess(g *ExecutionGraph) bool {
 	sc.mu.RLock()
 	defer sc.mu.RUnlock()
 
 	hasErrors := false
+	hasSkippedWithContinue := false
 	hasSuccessOrSkipped := false
-
-	logger.Debug(context.Background(), "Checking partial success conditions", "totalNodes", len(g.nodes))
 
 	for _, node := range g.nodes {
 		nodeStatus := node.State().Status
-		logger.Debug(context.Background(), "Node status check",
-			"nodeName", node.Name(),
-			"status", nodeStatus.String(),
-			"shouldContinue", node.shouldContinue(context.Background()))
 
 		switch nodeStatus {
-		case NodeStatusSuccess, NodeStatusSkipped:
+		case NodeStatusSuccess:
 			hasSuccessOrSkipped = true
-			logger.Debug(context.Background(), "Found success/skipped node", "nodeName", node.Name())
+		case NodeStatusSkipped:
+			hasSuccessOrSkipped = true
+			// Check if this skip was intentional (with continueOn.skipped = true)
+			if node.shouldContinue(context.Background()) {
+				hasSkippedWithContinue = true
+			}
 		case NodeStatusError:
 			// Check if this error node was allowed to continue
 			if node.shouldContinue(context.Background()) {
 				hasErrors = true
-				logger.Debug(context.Background(), "Found error node with continueOn", "nodeName", node.Name())
 			} else {
 				// If any error node was not allowed to continue, this is not partial success
-				logger.Debug(context.Background(), "Found error node without continueOn, not partial success", "nodeName", node.Name())
 				return false
 			}
 		case NodeStatusCancel:
 			// Cancelled nodes prevent partial success
-			logger.Debug(context.Background(), "Found cancelled node, not partial success", "nodeName", node.Name())
 			return false
 		case NodeStatusRunning, NodeStatusNone:
 			// DAG is not finished yet
-			logger.Debug(context.Background(), "DAG not finished, not partial success", "nodeName", node.Name(), "status", nodeStatus.String())
 			return false
 		}
 	}
 
-	result := hasSuccessOrSkipped && hasErrors
-	logger.Debug(context.Background(), "Partial success check result",
-		"hasSuccessOrSkipped", hasSuccessOrSkipped,
-		"hasErrors", hasErrors,
-		"result", result)
-
-	// Partial success requires both successful/skipped nodes AND error nodes that were allowed to continue
-	return result
+	// Partial success occurs when:
+	// 1. There are successful/skipped nodes AND error nodes that were allowed to continue, OR
+	// 2. There are nodes that were skipped with continueOn.skipped = true (indicating optional work was not completed)
+	return (hasSuccessOrSkipped && hasErrors) || hasSkippedWithContinue
 }
 
 func (sc *Scheduler) isTimeout(startedAt time.Time) bool {
--- a/internal/digraph/scheduler/scheduler.go
+++ b/internal/digraph/scheduler/scheduler.go
@@ -283,7 +283,12 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, progre
 								node.SetStatus(NodeStatusSuccess)
 							} else {
 								node.MarkError(execErr)
-								sc.setLastError(execErr)
+								// Only set lastError if this failure should not be allowed to continue
+								// If shouldContinue() is true, this error is allowed and should contribute
+								// to partial success rather than overall failure
+								if !node.shouldContinue(ctx) {
+									sc.setLastError(execErr)
+								}
 							}
 						}
 					}
--- a/internal/digraph/scheduler/scheduler.go
+++ b/internal/digraph/scheduler/scheduler.go
@@ -716,49 +716,47 @@ func (sc *Scheduler) isSucceed(g *ExecutionGraph) bool {
 	return true
 }
 
-// isPartialSuccess checks if the DAG completed with some failures or skips that were allowed to continue.
-// This represents scenarios where primary work was completed but non-critical errors occurred or optional steps were skipped.
+// isPartialSuccess checks if the DAG completed with some failures that were allowed to continue.
+// This represents scenarios where execution continued despite failures due to continueOn conditions.
 func (sc *Scheduler) isPartialSuccess(g *ExecutionGraph) bool {
 	sc.mu.RLock()
 	defer sc.mu.RUnlock()
 
-	hasErrors := false
-	hasSkippedWithContinue := false
-	hasSuccessOrSkipped := false
+	hasFailuresWithContinueOn := false
+	hasSuccessfulNodes := false
 
+	// First pass: check if any failed node is NOT allowed to continue
+	// If so, this is an error, not partial success
 	for _, node := range g.nodes {
-		nodeStatus := node.State().Status
-
-		switch nodeStatus {
-		case NodeStatusSuccess:
-			hasSuccessOrSkipped = true
-		case NodeStatusSkipped:
-			hasSuccessOrSkipped = true
-			// Check if this skip was intentional (with continueOn.skipped = true)
-			if node.shouldContinue(context.Background()) {
-				hasSkippedWithContinue = true
-			}
-		case NodeStatusError:
-			// Check if this error node was allowed to continue
-			if node.shouldContinue(context.Background()) {
-				hasErrors = true
-			} else {
-				// If any error node was not allowed to continue, this is not partial success
+		if node.State().Status == NodeStatusError {
+			if !node.shouldContinue(context.Background()) {
+				// Found a failed node that was NOT allowed to continue
+				// This disqualifies the DAG from being partial success
 				return false
 			}
-		case NodeStatusCancel:
-			// Cancelled nodes prevent partial success
-			return false
-		case NodeStatusRunning, NodeStatusNone:
-			// DAG is not finished yet
-			return false
 		}
 	}
 
-	// Partial success occurs when:
-	// 1. There are successful/skipped nodes AND error nodes that were allowed to continue, OR
-	// 2. There are nodes that were skipped with continueOn.skipped = true (indicating optional work was not completed)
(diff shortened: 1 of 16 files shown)


internal/digraph/scheduler/scheduler.go
```diff
@@ -284,7 +287,12 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, progre
 								node.SetStatus(NodeStatusSuccess)
 							} else {
 								node.MarkError(execErr)
-								sc.setLastError(execErr)
+								// Only set lastError if this failure should not be allowed to continue
+								// If shouldContinue() is true, this error is allowed and should contribute
+								// to partial success rather than overall failure
+								if !node.shouldContinue(ctx) {
+									sc.setLastError(execErr)
+								}
```
thefishhat, line 291: Not entirely sure about this one. @yottahmd - should last error be recorded for ignored errors? I had to change it to fix the scheduler tests.

ghansham:
Can we have another screenshot for skipped step and partial success.

thefishhat:
> Can we have another snapshot for skipped step and partial success. 

What do you mean by snapshot here? Another screenshot?

ghansham:
> > Can we have another snapshot for skipped step and partial success.
> 
> What do you mean by snapshot here? Another screenshot?

Yes. Sorry for the wrong word selection

internal/digraph/scheduler/scheduler.go
yohamta0: This looks good to me! Thanks for your careful attention to detail.

yohamta0:
Hi @thefishhat, could you please rebase your branch onto `main`? We’re no longer using `next`. Apologies for the confusion.

thefishhat:
> Can we have another screenshot for skipped step and partial success.

image

Should this have said `Partial Success`? I might have misunderstood the issue requirement. My thinking is `Partial Success` means overall success in the presence of errors.

This is the DAG definition:
```
name: skipped_test
steps:
  - name: step1
    command: echo "Step 1 success"
  - name: step2
    command: echo "This won't run"
    depends:
      - step1
    preconditions:
      - condition: "false"  # This will always fail
        expected: "true"
    continueOn:
      skipped: true  # Allow continuation when skipped
  - name: step3
    command: echo "Step 3 success"
    depends:
      - step2
```

thefishhat:
> Hi @thefishhat, could you please rebase your branch onto `main`? We’re no longer using `next`. Apologies for the confusion.

Apologies! I must have missed the message. I've rebased onto `main` now.

ghansham:
> > Can we have another screenshot for skipped step and partial success.
> 
> image
> 
> Should this have said `Partial Success`? I might have misunderstood the issue requirement. My thinking is `Partial Success` means overall success in the presence of errors.
> 
> This is the DAG definition:
> 
> ```
> name: skipped_test
> steps:
>   - name: step1
>     command: echo "Step 1 success"
>   - name: step2
>     command: echo "This won't run"
>     depends:
>       - step1
>     preconditions:
>       - condition: "false"  # This will always fail
>         expected: "true"
>     continueOn:
>       skipped: true  # Allow continuation when skipped
>   - name: step3
>     command: echo "Step 3 success"
>     depends:
>       - step2
> ```

Yup. You are absolutely right. This is also a case of partial success.

thefishhat:
Sorry, @ghansham I might have not been clear. My point was that intuitively to me this doesn't seem like a "Partial Success" case because there are no errors present. Could you clarify if you are saying that we should treat skipped nodes as errors in terms of defining overall success?

internal/digraph/scheduler/scheduler.go
```diff
@@ -698,6 +716,59 @@ func (sc *Scheduler) isSucceed(g *ExecutionGraph) bool {
...
+			logger.Debug(context.Background(), "DAG not finished, not partial success", "nodeName", node.Name(), "status", nodeStatus.String())
+			return false
+		}
+	}
+
+	result := hasSuccessOrSkipped && hasErrors
+	logger.Debug(context.Background(), "Partial success check result",
+		"hasSuccessOrSkipped", hasSuccessOrSkipped,
+		"hasErrors", hasErrors,
+		"result", result)
```
thefishhat: nit: What is general consensus on logs for Dagu? Is it better to be as verbose as possible or is it too much clutter for customers?

ghansham:
I consider this scenario as partial success:
A Step (say step N)  in the DAG failed but continueOn is true, making this step kind of optional. As a result, the later steps that depended on the success of the step N got skipped.

Basic idea is that some optional steps (continueOn) in sequence were not executed and some steps could not get executed because they depended on optional step success.

Let me know if still there is doubt.

thefishhat:
> I consider this scenario as partial success: A Step (say step N) in the DAG failed but continueOn is true, making this step kind of optional. As a result, the later steps that depended on the success of the step N got skipped.
> 
> Basic idea is that some optional steps (continueOn) in sequence were not executed and some steps could not get executed because they depended on optional step success.
> 
> Let me know if still there is doubt.

But in the scenario above, `step3` that depends on a skipped optional `step2` still gets executed. Currently a skipped step seems to not be considered a failure, and does not result in cascading failures for dependent steps. Are we sure we want skipped tests to count towards a partial success?

thefishhat:
I've pushed up a latest commit that counts skipped nodes towards the `Partial Success` status. I'll leave it up to you to decide if that is the desired implementation. If you change your mind, we can always revert the last commit. Here's how the skipped node DAG looks like now:

image

For this DAG:
```yaml
name: skipped_test
steps:
  - name: step1
    command: echo "Step 1 success"
  - name: step2
    command: echo "This won't run"
    depends:
      - step1
    preconditions:
      - condition: "false"  # This will always fail
        expected: "true"
    continueOn:
      skipped: true  # Allow continuation when skipped
  - name: step3
    command: echo "Step 3 success"
    depends:
      - step2
```

ghansham:
This is the real world example, I am showing the part of the dag:

Point_Blur_Dec272024_202600-1


See how the steps will look in yaml corresponding to the encircled part of the dag (mandatory steps can be added):
```
steps:
.
.
.
.
.
  - name: optional_task
    command: "echo 72; exit 72"
    output: opt_result
    continueOn:
      exitCode: [72]
  - name: skippable_task1
    command: task1.sh
    depends: optional_task
    preconditions:
      - condition: "${opt_result}"
        expected: "0" 
  - name: skippable_task2
    command: task2.sh
    depends: skippable_task1
    preconditions:
      - condition: "${opt_result}"
        expected: "0" 
```

The optional task will fail when a non zero exit code other than 72 is returned. If 72 is returned, it will continue. Skippable tasks will be skipped if optional task returns any value other than 0. We are using output capturing here. This will become even more sophisticated when following issue gets resolved:

https://github.com/dagu-org/dagu/issues/986

It will help us get rid of capturing output.

I think continueOn is mandatory. Skipped is optional. It may or may not be there.

internal/digraph/scheduler/scheduler.go
yohamta0: Debug logs are mainly for Dagu developers and aren’t shown by default, so I don’t think being a bit verbose is a problem. There’s no strong consensus on this, though. Happy to hear any suggestions you have!

yohamta0:
This is a really helpful discussion!
I feel it might make sense to mark it as `partial success` if one or more nodes have a `failed` status, but the DAG as a whole still finishes. What do you think? @ghansham @thefishhat

ghansham:
> This is a really helpful discussion! I feel it might make sense to mark it as `partial success` if one or more nodes have a `failed` status, but the DAG as a whole still finishes. What do you think? @ghansham @thefishhat

@yottahmd, Agree. It makes sense. Out of curiosity, is there any possibility without the use of continueOn clause that dag may get succesful even after a step fails?

thefishhat:
> It makes sense. Out of curiosity, is there any possibility without the use of continueOn clause that dag may get succesful even after a step fails?

If we define step failure as "failed" node status, and not "skipped" status, then I can't see a way for a DAG to result in a success state without `continueOn`.

To clarify, the current logic to determine partial success is:

- if all nodes which have `failed` status, were marked with `continueOn: failure`, then the DAG run is a partial success.
- if there are no error nodes, but there are skipped nodes, if the skipped nodes were skipped intentionally (with `continueOn: skipped`) then that is also partial success

What is not considered partial success:

- DAG contains any errors that did not have `continueOn`
- DAG contains skipped nodes with `continueOn` but also contain failed nodes with no `continueOn`
- DAG contains skipped nodes that did not have `continueOn` set for skipped

Does this match our currebr thinking? Should all skipped nodes count towards partial success regardless of `continueOn`? Or should they just count towards Success as before, because they are not errors?

If we answer these two questions above, I think the PR will be done.

yohamta0:
Thank you very much for the thorough analysis. I really appreciate it.

To be honest, I don't think we should mark DAG with skipped nodes as a partial success. For example, this could also happen when a precondition check isn't met, which I don't necessarily see as a failure.
> if there are no error nodes, but there are skipped nodes, if the skipped nodes were skipped intentionally (with continueOn: skipped) then that is also partial success

Regarding cases that aren’t partial successes, I agree with you. In my view, skipped nodes without any failures shouldn’t count as partial success.

To clarify, I think a DAG should be marked as partial success when:
- The DAG contains one or more nodes in a failure state, but execution continued because of a matching `continueOn` condition.

What do you think? @thefishhat @ghansham

thefishhat:
That is also my thinking, but perhaps @ghansham might provide an alternative view as per his example above.

yohamta0:
If a node exits with a code like 72, it will be marked as a failure. In this case, the DAG status should be set to `PartialSuccess`. Without the `markSuccess: true` option, nodes will remain in a failed state even if the `continueOn` setting allows the DAG to proceed.

**Example:**
```yaml
steps:
  - name: optional_task
    command: "echo 72; exit 72"
    output: opt_result
    continueOn:
      exitCode: [72]
  - name: step
    command: echo "$opt_result"
```

Result:
image

**With `markSuccess: true`:**

```yaml
steps:
  - name: optional_task
    command: "echo 72; exit 72"
    output: opt_result
    continueOn:
      markSuccess: true
      exitCode: [72]
  - name: step
    command: echo "$opt_result"
```

image

In this case, all nodes are successful, so the final DAG status is success.

yohamta0:
@ghansham When you get a moment, could you share your thoughts with us?

ghansham:
The concept is simple. Most of the nodes in a dag are successful. Only a few nodes failed with `continueOn` true (based on exitCodes in my quoted example, other ways are also possible). Nodes that depended on these failed nodes got skipped.
So our criteria should be if some node failed with `continueOn` returning true.

I agree with @yottahmd that skipping of nodes can happen due to precondition which is not in our control and can't be a criteria to mark dag partial success. And moreover pre conditions can be used to implement `switch case` logic at step level in the dag where a set of nodes are going to be executed and others skipped. Check this case:

https://dagu.dev/examples/#conditional-steps

The above is `success` and not `partial_success`. So precondition based skipped can't be used as criteria to mark a dag `partial_success`.
The above example used to exist in dagu readthedocs earlier.

Regarding `markSuccess`, if a node is marked `success` irrespective of its actual status, it is like users wants to keep dag status `success` and not `partial_success`. This new feature actually wants to give user an extra information that some of the nodes were not successful which was earlier not possible. Before this feature, users had to manually check if the optional nodes (with `continueOn` set to true based on certain conditions) failed or were successful.

```
steps:
  - name: optional_task
    command: "echo 72; exit 72"
    output: opt_result
    continueOn:
      exitCode: [72]
  - name: step
    command: echo "$opt_result"
```

The final status should be `partial_success` in the above case

To keep things simple and relevant to real world use cases, we can ensure that atleast one node status was `success`. Otherwise it can't be termed `partial_success` 😄 because not even a single node status was `success` 😄.  What do you think @thefishhat  and @yottahmd ?

yohamta0:
Thank you for the detailed insights! This is really helpful. I believe we're aligned on this. Shall we proceed with this approach? @thefishhat

Condition for marking a DAG as Partial Success:
- The DAG contains one or more nodes in a failure state, but execution continued because of a matching continueOn condition.

ghansham:
From my side that is sufficient

thefishhat:
Done! I reverted the implementation back to the initial state where only errors contribute to partial success, while also readjusting to account for the `markSuccess` override. Please have another look now, @yottahmd !

ghansham:
I am grateful to @thefishhat for getting it through🙏

ghansham:
Great work done by @thefishhat.❤️
Thanks to @yottahmd for accepting this feature.

PR #1050, opened 2025-06-22
feat: enhance repeat policy with while until

thefishhat:
Still in draft because PR has Todos.

Todo:
* ensure tests for all repeat behaviors (both implicit and explicit)
* add validation for when `while`/`until` is used to always be accompanied by a `condition` or `exitCode`

Behaviour:
1. **Mode is omitted** (backwards compatibility)
* condition and expected are set: mode defaults to `until` 
* condition/exitcode are set: default to `while` 
2. **Mode is `while`:** 
* \+ condition & expected: repeats as long as the condition matches the expected value
* \+ condition: repeats as long as the condition evaluates successfully
* \+ exit code: repeats as long as step exit code matches any in the list
* neither: repeats as long as the step execution succeeds (no execErr)
3. **Mode is `until`**: vice versa
4. **Mode is `true`**: executes as a `while` with no `condition` or `exitCode` (indefinite loop, repeats as long as the step succeeds).

### DIFF
--- a/internal/frontend/api/v2/transformer.go
+++ b/internal/frontend/api/v2/transformer.go
@@ -30,7 +30,7 @@ func toStep(obj digraph.Step) api.Step {
 	}
 
 	repeatPolicy := api.RepeatPolicy{
-		Repeat:   ptrOf(obj.RepeatPolicy.Repeat),
+		Repeat:   ptrOf(obj.RepeatPolicy.Repeat != ""),
 		Interval: ptrOf(int(obj.RepeatPolicy.Interval.Seconds())),
 	}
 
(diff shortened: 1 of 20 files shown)


yohamta0:
@thefishhat The code looks fantastic!

Regarding todo:
> ensure tests for all repeat behaviors (both implicit and explicit)

Yes, agreed.

> add validation for when while/until is used to always be accompanied by a condition or exitCode

Yes, let's include this.

The behaviour you described looks correct. Thank you for organising all of these complex details.

thefishhat:
@yottahmd ready for review 🙏

internal/frontend/api/v2/transformer.go
```diff
@@ -30,7 +30,7 @@ func toStep(obj digraph.Step) api.Step {
 	}
 
 	repeatPolicy := api.RepeatPolicy{
-		Repeat:   ptrOf(obj.RepeatPolicy.Repeat),
```
yohamta0, line 33: The response type for the Repeat field should be updated to an repeat mode enum, but let's handle that in a separate PR.

Issue #1051, opened 2025-06-22
[Bug] Problem with Repeat policy and exitCode


jeremydelattre59:
Hi,

You need to drop the test.sh file into /tmp and make it executable. For now, it will return 1.
The Manuel.yaml file should be dropped into dags.

When I launch the workflow, the first execution shows the "running" for the first minute. For the next one, it falls into failed but it continues to run (the duration continues to increase). Finally, when I modify the test.sh file to replace 1 with 0, the execution goes to completed but the last step remains canceled and the overall execution remains in failed.

[test.sh.txt](https://github.com/user-attachments/files/20851444/test.sh.txt)

[Manuel.yaml.txt](https://github.com/user-attachments/files/20851445/Manuel.yaml.txt)

Regards,
Jeremy

thefishhat:
I managed to reproduce this. Having a look 👀

thefishhat:
I found the problem. The repeat logic policy is correct, but the logic responsible for defining workflow & node success was not accounting for the possibility of having repeat policy on non-zero exit codes.

I'm working on a fix here: https://github.com/dagu-org/dagu/pull/1052

yohamta0:
Hi @jeremydelattre59, thank you for reporting this issue. It has been resolved in the latest version - thanks to @thefishhat! Let us know if you run into any other issues.

PR #1052, opened 2025-06-22
fix: repeat policy non zero exit codes

thefishhat:
## What does this PR do?

* Introduce case for repeat policy on non-zero exit codes to workflow/node success definition logic

## Why is this needed?
Addresses issue #1051 . Previously although the repeat process was working correctly, nodes were marked as failed due to their non-zero exit code early, not allowing to account for exit codes repeat policy.

### DIFF
--- a/internal/digraph/scheduler/scheduler.go
+++ b/internal/digraph/scheduler/scheduler.go
@@ -257,6 +257,10 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, progre
 				for setupSucceed && !sc.isCanceled() {
 					execErr := sc.execNode(ctx, node)
 					if execErr != nil {
+						lastExit := node.State().ExitCode
+						step := node.Step()
+						shouldRepeatOnExitCode := len(step.RepeatPolicy.ExitCode) > 0 && slices.Contains(step.RepeatPolicy.ExitCode, lastExit)
+
 						status := node.State().Status
 						switch {
 						case status == NodeStatusSuccess || status == NodeStatusCancel:
@@ -274,6 +278,11 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, progre
 							if sc.handleNodeRetry(ctx, node, execErr) {
 								continue ExecRepeat
 							}
+
+						case shouldRepeatOnExitCode:
+							// The step failed with an exit code that is configured to trigger a repeat.
+							// Don't mark the node as failed, as it will be repeated.
+							// The repeat logic later in the loop will handle the repetition.
 
 						default:
 							// finish the node
--- a/internal/digraph/scheduler/scheduler.go
+++ b/internal/digraph/scheduler/scheduler.go
@@ -257,10 +257,6 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, progre
 				for setupSucceed && !sc.isCanceled() {
 					execErr := sc.execNode(ctx, node)
 					if execErr != nil {
-						lastExit := node.State().ExitCode
-						step := node.Step()
-						shouldRepeatOnExitCode := len(step.RepeatPolicy.ExitCode) > 0 && slices.Contains(step.RepeatPolicy.ExitCode, lastExit)
-
 						status := node.State().Status
 						switch {
 						case status == NodeStatusSuccess || status == NodeStatusCancel:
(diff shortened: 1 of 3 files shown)


yohamta0:
@thefishhat Thanks a ton for tackling the bug in the repeat policy!

I tested the changes on my local environment and encountered an issue when running a DAG with the following configuration. Specifically, when `repeatPolicy.limit` is set and the node's status becomes running, the scheduler does not terminate and enters an infinite loop.
```yaml
- name: Check Stop RAID Scrubbing
  description: Vérifie si tous les disques RAID ne sont plus en cours de nettoyage
  #command: /share/CE_CACHEDEV1_DATA/Script/Utilitaires/scrubbing_running.sh
  command: ~/Downloads/test.sh
  continueOn:
    exitCode: [0]
    markSuccess: true
  repeatPolicy:
    repeat: true
    intervalSec: 10
    exitCode: [1]
    limit: 3
  depends:
    - Initialize
- name: Done
  command: echo "Stop Manuel"
```

What do you think about this change? It looks like it fixes the problem on my end and passes the tests.
```diff
@@ -340,6 +332,11 @@ func (sc *Scheduler) Schedule(ctx context.Context, graph *ExecutionGraph, progre
                                        }
 
                                        if shouldRepeat && !sc.isCanceled() {
+                                               node.SetStatus(NodeStatusRunning) // reset status to running for the repeat
+                                               if sc.lastError == node.Error() {
+                                                       sc.setLastError(nil) // clear last error if it was the same as node error
+                                               }
+
                                                logger.Info(ctx, "Step will be repeated", "step", node.Name(), "interval", step.RepeatPolicy.Interval)
```
Let me know what you think :)

thefishhat:
This is not perfectly clear to me, but I committed your changes. I do have a few questions:

* `node.SetStatus(NodeStatusRunning)` - is this needed? At this point of the code it should already be in a Running state, as the error does not get persisted on the node at line 282 in `case shouldRepeatOnExitCode:`, no?
* How did you discover that you need to clear the `lastError`? I can't seem to understand how the flow changes when `lastError` is set vs. when it is cleared. What is the infinite loop caused by?

yohamta0:
@thefishhat Apologies for the insufficient explanation.

Let's consider the following DAG as a test case. With the current code, the process is as follows:

```yaml
steps:
  - name: step1
    command: exit 1
    continueOn:
      failure: true
    repeatPolicy:
      repeat: true
      exitCode: [1]
      limit: 2
```

**Execution Trace:**

1.  `step1` is identified as the node to be executed.
2.  `step1` is executed.
      * The node's status is kept as `running` due to the `case shouldRepeatOnExitCode:` logic.
3.  `step1` is executed for a second time.
      * The status remains `running` for the same reason.
4.  The program then exits the node execution goroutine (`go func(ctx context.Context, node *Node) {`) due to the conditional branch at this location:
    [https://github.com/dagu-org/dagu/blob/09c44400150590ba3e8c298303ad40850c1e4c77/internal/digraph/scheduler/scheduler.go\#L358-L361](https://github.com/dagu-org/dagu/blob/09c44400150590ba3e8c298303ad40850c1e4c77/internal/digraph/scheduler/scheduler.go#L358-L361)
5.  Because the node's status is left as `running`, the scheduler can no longer retrieve the next node here, causing it to enter an infinite loop and never terminate:
    [https://github.com/dagu-org/dagu/blob/09c44400150590ba3e8c298303ad40850c1e4c77/internal/digraph/scheduler/scheduler.go\#L162-L164](https://github.com/dagu-org/dagu/blob/09c44400150590ba3e8c298303ad40850c1e4c77/internal/digraph/scheduler/scheduler.go#L162-L164)

Therefore, I propose removing the `case` block entirely. This ensures that if the node has an `Error` status when the `shouldRepeat` condition is evaluated, the node execution loop can exit safely without causing the scheduler to hang.

```diff
-						case shouldRepeatOnExitCode:
-							// The step failed with an exit code that is configured to trigger a repeat.
-							// Don't mark the node as failed, as it will be repeated.
-							// The repeat logic later in the loop will handle the repetition.
```

Regarding the clearing of `lastError`: this is necessary because a persistent `lastError` would cause the entire schedule to be marked with an `Error` status. My expectation is that if a step ultimately succeeds via the `repeat` mechanism, the overall status should be `Success`.

However, I acknowledge that this fix may be more of a temporary patch. A more robust, long-term solution should probably be considered through refactoring.

Please let me know if any part of this explanation is unclear. I'm totally open to suggestions for other approach as well :)

thefishhat:
Thanks for the thorough explanations! I've commit the change just like how you explained it to me. Let's get this fix in before it impacts too many customers!

I'll next look into refactoring the scheduler with this evolved understanding, as it has become quite unwieldy! This will help future extensions, such as the `while/until` repeat method that I have in draft right now.

yohamta0:
@thefishhat LGTM 🚀🚀🚀

> I'll next look into refactoring the scheduler with this evolved understanding, as it has become quite unwieldy! This will help future extensions, such as the while/until repeat method that I have in draft right now.

Yes, the scheduler definitely needs a refactor. Let's make it happen. Thanks again!

ghansham:
May be need separate code files for handling different clauses like retry, etc.  Design patterns can be explored

Issue #1053, opened 2025-06-22
When using ssh links, all output is captured as an error log


NebulaCoding1029:
```yaml
steps:
  - name: docker-pull
    executor:
      type: ssh
      config:
        user: xxx
        ip: xxx
        key: /xxx/key
    command:
      - cat /logs/tmp/pull_info.log 2>&1
    output: PULL_OUTPUT
```

Before the version upgrade, there was no problem, but when I updated the docker image, now all output is directed to ERR, even if it is a simple cat.

It's a great project, thank you to the development team！

yohamta0:
Hi @NebulaCoding1029, thank you for the kind words and for reporting the issue! The problem has been fixed in the latest version. Please give it a try!

NebulaCoding1029:
The log output of the SSH executor is now in line with the intended effect！
ありがとう、スーパーハムスター ^_^

PR #1057, opened 2025-06-23
fix: separate stdout and stderr streams in SSH executor

yottahmd:
**Overview**
The SSH executor output was being incorrectly captured as error logs. This fix separates stdout and stderr streams in the SSH executor, matching expected behavior.

Issue: #1053
Feedback-from: @NebulaCoding1029

**Changes**
- Added `stderr` field to `sshExec` struct
- Initialized stderr field in constructor
- Added tests for stdout/stderr separation

### DIFF
--- a/internal/digraph/executor/http_test.go
+++ b/internal/digraph/executor/http_test.go
@@ -1,0 +1,153 @@
+package executor
+
+import (
+	"context"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


NebulaCoding1029:
Thank you very much for contributing!!
If it's convenient, please update the docker image……
ありがとうございます！！！

PR #1088, opened 2025-07-03
fix: jq executor output formatting issues

yottahmd:
**Overview**
Users need jq executor to properly format output values. When using queries that return multiple values or special formatting like `@tsv`, the output was being incorrectly concatenated or escaped.

As a solution, fixed the jq executor to add newlines between outputs and preserve special characters in raw mode.

Reported-by: @yurivish

**Changes**
- Fixed jq executor to add newlines between output values in both raw and non-raw modes
- Fixed raw mode to preserve special characters (tabs, newlines) when jq outputs strings
- Added comprehensive integration tests for multiple outputs, string outputs, and TSV formatting

**Example 1 - Multiple values**
```yaml
steps:
  - name: extract-values
    executor: 
      type: jq
      config:
        raw: true
    script: |
      { "data": [1, 2, 3] }
    command: '.data[]'
    output: RESULT
```
Before: `RESULT` would be `123`
After: `RESULT` is now:
```
1
2
3
```

**Example 2 - TSV output**
```yaml
steps:
  - name: extract-tsv
    executor: 
      type: jq
      config:
        raw: true
    script: |
      { "data": [1, 2, 3] }
    command: '.data[] | [., 100 * .] | @tsv'
    output: RESULT
```
Before: `RESULT` would be `1\\t100\\n2\\t200\\n3\\t300` (escaped tabs)
After: `RESULT` is now properly formatted TSV with real tabs:
```
1	100
2	200
3	300
```

### DIFF
--- a/internal/digraph/executor/http_test.go
+++ b/internal/digraph/executor/http_test.go
@@ -200,7 +200,7 @@ func TestHTTPExecutor_CrossPlatform(t *testing.T) {
 						"headers":  r.Header,
 						"url":      r.URL.String(),
 					}
-					
+
 					w.Header().Set("Content-Type", "application/json")
 					w.WriteHeader(nethttp.StatusOK)
 					_ = json.NewEncoder(w).Encode(response)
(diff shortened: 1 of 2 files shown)


yurivish:
Thank you for the diagnosis and extremely quick fix!

PR #1126, opened 2025-07-24
removed harcoded value of 10000 in models.NewPaginator

kriyanshii:
hardcoding value 10000 generates issue and only shows top 200 dags while listing.

https://github.com/dagu-org/dagu/blob/111594fdd6186db2643f65661cabcc751f944702/internal/frontend/api/v2/dags.go#L320

changed this value of 10000 with 
`maxInt := int(^uint(0) >> 1)`


yohamta0:
Thank you so much for catching this! Let me merge this and refactor to address the root issue.

PR #1127, opened 2025-07-24
chore: update dependencies

reneleonhardt:
### Chores
- Update dependencies
- Let dependabot update dependencies


reneleonhardt:
Everywhere is Python 3.11 required, can that be upgraded to 3.12?
3.14 will be released soon.

https://github.com/dagu-org/dagu/blob/main/docs/features/distributed-execution.md?plain=1#L114
```markdown
--worker.labels python=3.11,cuda=11.8,torch=2.0
```
What about CUDA and torch?

yohamta0:
I don't think we need to update the versions in these examples. It might be better to simply remove this one, since it's not very practical.
```
# Software environment
--worker.labels python=3.11,cuda=11.8,torch=2.0
```
Thanks for upgrading those dependencies!

yohamta0:
@chrishoage Could you check linter errors, please?

reneleonhardt:
Fixed new golint rules and removed one doc line as requested.

reneleonhardt:
If you can, check locally, especially npm updates can be tricky.

Gosh, golint could use some AI to be smarter, why not suggest crypto/rand right from the start 😞

yohamta0:
I will test this on my end tomorrow. Thank you!

reneleonhardt:
Could you double-check your codecov configuration?
I didn't need to touch internal/fileutil/cache.go for those updates, the report is wrong.
```
Attention: Patch coverage is 0% with 5 lines in your changes missing coverage.
```

Issue #1120, opened 2025-07-22
Feature request: provide a native health-check for the scheduler service


jonasban:
As I mentioned here in this [issue](https://github.com/dagu-org/dagu/issues/1103#issuecomment-3103791934) I run the web UI and the scheduler as separate services in docker compose. 

Right now I am testing if the scheduler is running by creating a simple DAG that writes a file with a timestamp every minute it runs. Docker healtcheck then reads this file every few minutes and checks how "stale" that file is. If the timestamp written to the file is older than 3 minutes when the healthcheck is performed it is assumed that scheduler is dead and is then restarted. 

I think we can agree that this hacky workaround is not the optimal way to check the health of the scheduler. Maybe the scheduler could open a dummy socket that could be checked for existence instead? This is just a suggestion, you guys know your product better and can suggest more optimal solutions.

I think this is something multiple people would gladly use, what do you think?

Cheers!

yohamta0:
I 100% agree that we need a way to perform health checks on the scheduler. I'll go ahead and add a health check API endpoint, similar to the one we have for the server process.

Sarvesh-11:
I would love to take this, can i work on this?

yohamta0:
Sorry @Sarvesh-11, this issue has been addressed already. I'll close this issue when the next minor version (1.18) is released.

jonasban:
@yottahmd do you have any guesstimates when `1.18` is going to be released? 🤞

yohamta0:
@jonasban Sorry it’s taken a while. I'm aiming to release the next version by the end of this weekend.

jonasban:
> [@jonasban](https://github.com/jonasban) Sorry it’s taken a while. I'm aiming to release the next version by the end of this weekend.

Dagu's development is moving fast, no worries!

yohamta0:
This feature has been implemented in PR #1129 and released in v1.18.0. The scheduler now provides a health check endpoint at http://localhost:8090/health that is automatically enabled when running 'dagu scheduler'. Thank you for suggesting this improvement\!

PR #1129, opened 2025-07-25
Add health check endpoint for scheduler

yottahmd:
**Overview:**
Users need to monitor scheduler health status for production deployments. They currently have no way to check if the scheduler is running properly.

This PR adds a simple HTTP health check endpoint to the scheduler that monitoring tools can use.

Issue: #1120

**Changes:**
- Add `scheduler.port` configuration field (default: 8090)
- Implement `/health` endpoint that returns `{"status": "healthy"}`
- Health server only runs with `dagu scheduler` command, not `dagu start-all`
- Support `DAGU_SCHEDULER_PORT` environment variable
- Can disable health check by setting port to 0

**Example:**
```yaml
# config.yaml
scheduler:
  port: 8090  # Health check port (default)
```

```bash
# Start scheduler with health check
dagu scheduler

# Check health
curl http://localhost:8090/health
# Returns: {"status": "healthy"}
```

### DIFF
--- a/design/20_implemented/scheduler-health-check-requirements.md
+++ b/design/20_implemented/scheduler-health-check-requirements.md
@@ -1,0 +1,174 @@
+# Scheduler Health Check Design
+
+## Overview
+
+This design document outlines the implementation of an HTTP health check endpoint for the Dagu scheduler component. The health check will enable external monitoring systems (e.g., Kubernetes, monitoring tools) to verify that the scheduler is running and healthy.
+
+## Requirements
+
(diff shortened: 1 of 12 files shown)


jonasban:
Hello! Great to see you working on this! I have a question about the health check web server. Do I understand correctly that `scheduler.port` functionality will be local to the instance running the scheduler? In other words, will the user be able to check whether the UI and scheduler processes are healthy without depending on one another?

jonasban:
I checked the design doc and I see that the answer to my question is `yes`. Ignore my comment above :pray:

Issue #1136, opened 2025-07-26
fix: duration display shows 'undefinedd undefinedh' for invalid dates


yottahmd:
## Problem
When a DAG run has invalid or undefined date values, the duration display shows malformed text like "undefinedd undefinedh 1m" instead of handling the error gracefully.

## Root Cause
The custom duration format function in `ui/src/lib/dayjs.ts` doesn't properly handle cases where duration calculations result in NaN or undefined values. When `dayjs.duration()` receives invalid input, the `.asDays()`, `.hours()`, etc. methods may return NaN, which when converted to string becomes "NaN" or in some cases JavaScript returns undefined.

## Location
The issue occurs in:
- `ui/src/lib/dayjs.ts` - Custom format function (lines 22-51)
- Used by `ui/src/features/dags/components/dag-list/DAGTable.tsx` with format `'d[d] h[h] m[m]'`

## Expected Behavior
When duration calculation fails due to invalid dates, the display should show:
- A dash "-" or 
- A meaningful error indicator
- Skip showing days/hours if they are 0 or invalid

## Steps to Reproduce
1. Have a DAG run with corrupted or invalid date values
2. View the DAG list or run details
3. The duration column shows "undefinedd undefinedh Xm"

## Suggested Fix
Add validation in the custom format function to check if values are valid numbers before converting to string:
- Check for NaN/undefined values
- Only include units that have valid positive values
- Return a fallback value like "-" for completely invalid durations

Reported-by: yurivish

ghansham:
Thanks a lot for fixing this one.

Does that mean that date strings being used are not proper? Just for your reference, we are using migrated history where we such kind of issues. I will dig further to find root cause.

yohamta0:
This has been fixed in commit b88c2490 and released in v1.18.0. The duration display now properly handles invalid dates. Thank you @yurivish for finding this issue!

PR #1137, opened 2025-07-26
fix: duration display shows 'undefinedd undefinedh' for invalid dates

yottahmd:
**Overview:**
Users encountered a bug where DAG run durations displayed as "undefinedd undefinedh 1m" instead of showing the correct duration format. This occurred when the custom dayjs duration format function received certain date values.

The root cause was that the custom format function in `dayjs.ts` wasn't properly handling cases where duration methods returned undefined values, leading to malformed output strings.

Issue: #1136
Feedback-from: @yurivish

**Changes:**
- Replace custom dayjs format function with manual duration formatting in DAGTable.tsx
- Add date validation using `dayjs.isValid()` before calculating durations
- Implement cleaner duration display that only shows non-zero time units
- Remove debugging console.log statements

**Why:**
The custom format function was fragile and didn't handle edge cases properly. The new implementation is more explicit, easier to understand, and handles invalid dates gracefully.

**Example:**
Before: "undefinedd undefinedh 1m"
After: "1m 40s" (only showing relevant units)

**Notes:**
- TypeScript and lint checks pass
- The fix is localized to the DAGTable component where the issue manifests
- Invalid dates now result in no duration display rather than showing malformed text

### DIFF
--- a/ui/src/features/dags/components/dag-list/DAGTable.tsx
+++ b/ui/src/features/dags/components/dag-list/DAGTable.tsx
@@ -379,17 +379,34 @@ const defaultColumns = [
       let durationContent: React.ReactNode = null;
 
       if (finishedAt && finishedAt !== '-') {
-        const durationMs = dayjs(finishedAt).diff(dayjs(startedAt));
-        // Choose format based on duration length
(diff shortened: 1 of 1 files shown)


ghansham:
Thanks for fixing this issue

PR #1138, opened 2025-07-26
fixing next run for timezones

### LINKED ISSUE #1083: Web UI can no longer parse the CRON_TZ prefix
Amazing work on the new 1.17 version looks fantastic!

We upgraded and the schedule no longer can show the next run when using the `CRONTZ` prefix for the cron expression

Image

Sarvesh-11:
Fixes the next run where cron exp has CRON_TZ.

Refs #1083 
Fixes #1083

Testing proof
Screenshot 2025-07-26 at 5 20 15 PM

### DIFF
--- a/ui/src/features/dags/components/dag-list/DAGTable.tsx
+++ b/ui/src/features/dags/components/dag-list/DAGTable.tsx
@@ -146,6 +146,18 @@ declare module '@tanstack/react-table' {
 
 const columnHelper = createColumnHelper<Data>();
 
+function getTzAndExp(exp: string) {
+  const parts = exp.trim().split(/\s+/);
+
+  if (parts[0]?.startsWith("CRON_TZ=")) {
+    const timezone = parts[0]?.split("=")[1];
+    const cronExpr = parts?.slice(1).join(" ");
+    return [timezone, cronExpr];
+  } else {
+    return [parts.join(" ")];
+  }
+}
+
 function getNextSchedule(
   data: components['schemas']['DAGFile']
 ): CronDate | undefined {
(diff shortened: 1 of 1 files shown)


Sarvesh-11:
@yottahmd please review.

yohamta0:
Thank you, @Sarvesh-11 !
I have a question. How will the timezone part be used to determine the next schedule? Currently, it seems that the timezone is not taken into account. Would it be possible to consider the timezone when calculating the next scheduled time?

Sarvesh-11:
@yottahmd nice question.
Actually there were two option i had, for next run i can calculate the time of next run

1. according to the local timezone from where cron is triggered.
2. take time zone from cron expresssion.

Its better suited that user gets the time from where its executed(local time). anyways, for next run it is delta in time, timezone will not play a role in calculating delta.

Sarvesh-11:
i need to deep dive in this, now i am having some doubt about this. 

To answer you question.
Yes, can we use timezone, but need to figure out whats best from user POV.

yohamta0:
@Sarvesh-11 Could you please check the original issue #1083?

For example, there might be a case where a DAG is scheduled with `CRON_TZ=Australia/Sydney 0 9 * * *` (UTC+10), while the user's browser is in the UTC+0 timezone.
Could you confirm whether the time until the next run is being calculated correctly in a similar scenario?

Sarvesh-11:
@yottahmd i verified the case you mentioning. 
suppose we specified Australia/Sydney timezone, then cron parser will convert that time to UTC according to tz and then interval.next() will return UTC timestamp.

yottahmd (approved):
Looking great! Thank you!

Issue #1102, opened 2025-07-11
feat: support multiple email recipients in `errorMail.to` field


yottahmd:
Users need to send failure notifications to multiple email recipients. Currently `errorMail.to` only accepts a single string, requiring workarounds like distribution lists.

**Current behavior:**
- `errorMail.to` only accepts a single email string
- Using an array causes YAML error: "Incorrect type. Expected 'string'"

**Proposed solution:**
Accept both string and array formats for `errorMail.to`:

```yaml
# Single recipient (keep working)
errorMail:
  to: "admin@example.com"

# Multiple recipients (new)
errorMail:
  to: 
    - "admin1@example.com"
    - "admin2@example.com"
```

Note:
- inforMail.to and mail executor should also be modified as well.

feedback-by: @FlorianBEME

Sarvesh-11:
@yottahmd you working on this? i am thinking to work on this weekend.

yohamta0:
@Sarvesh-11 No, not yet. I'll assign this to you. Thank you!

Sarvesh-11:
thank you

yohamta0:
@Sarvesh-11 I understand you may have a lot on your plate. Let me know if you'd be okay with me taking over this issue.

ghansham:
We can add cc also here and even bcc 😄

yohamta0:
Good idea! 👍

Sarvesh-11:
hey @yottahmd sorry man, i was away from last week, didn't even got a chance to look at the issue. I see you started some work on it, you can work on it.

i will try to be more responsive in future.

yohamta0:
No worries at all @Sarvesh-11 ! The change for the `To` field has been already done. How about adding `cc` field or `bcc` to the mail executor? You can look at the PR #1125 to get idea of what files need to be modified. Ask me anything if you have any questions!

ghansham:
In the mail executor, for the failed dag notification, in the attachment we can give an optional choice to compress via zip, gzip, bzip2, etc the log files before sending. The reason is user step logs may be big enough to go as attachment. Just an idea. We can discuss it if required.

We can add an optional field:
```yaml
attachment_opts: 
     - combine: true #filename will be relevant if combine is true
     - filename: single_combined.tar #possible extensions: tar, zip, .tar.bz2, .tar.gz, .tar.xz, rar
```

Sarvesh-11:
@yottahmd i have started making changes for cc and bcc. FYI

[...]

PR #1139, opened 2025-07-26
[WIP] adding cc and bcc field

Sarvesh-11:
Refs #1102

### DIFF
--- a/internal/agent/reporter.go
+++ b/internal/agent/reporter.go
@@ -20,7 +20,7 @@ type Sender interface {
 }
 
 // SenderFn is a function type for sending reports.
-type SenderFn func(ctx context.Context, from string, to []string, subject, body string, attachments []string) error
+type SenderFn func(ctx context.Context, from string, to []string, cc []string, bcc []string, subject, body string, attachments []string) error
 
 // reporter is responsible for reporting the status of the scheduler
 // to the user.
@@ -41,10 +41,12 @@ func (r *reporter) reportStep(
 	if nodeStatus == status.NodeError && node.NodeData().Step.MailOnError && dag.ErrorMail != nil {
 		fromAddress := dag.ErrorMail.From
 		toAddresses := dag.ErrorMail.To
+		ccAddresses := dag.ErrorMail.Cc
+		bccAddresses := dag.ErrorMail.Bcc
 		subject := fmt.Sprintf("%s %s (%s)", dag.ErrorMail.Prefix, dag.Name, dagStatus.Status)
 		html := renderHTMLWithDAGInfo(dagStatus)
 		attachments := addAttachments(dag.ErrorMail.AttachLogs, dagStatus.Nodes)
-		return r.senderFn(ctx, fromAddress, toAddresses, subject, html, attachments)
+		return r.senderFn(ctx, fromAddress, toAddresses, ccAddresses, bccAddresses, subject, html, attachments)
 	}
 	return nil
 }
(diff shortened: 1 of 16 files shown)


Sarvesh-11:
@yottahmd please review.
let me know if needed any update.

still a work in progress, will test tomorrow for info and errorMail.

internal/mailer/mailer.go
```diff
@@ -136,13 +156,16 @@ func (m *Mailer) sendWithAuth(
 	}
 	resultChan := make(chan result, 1)
 
+	to = append(to, cc...)
+	to = append(to, bcc...)
+
```
yohamta0, line 161: It seems that it merges `cc` and `bcc` into `to` addresses. Could you check if this is correct implementation?

internal/mailer/mailer.go
```diff
@@ -157,9 +180,10 @@ func (m *Mailer) sendWithAuth(
 }
 
 func (*Mailer) composeHeader(
-	to []string, from string, subject string,
+	to []string, cc []string, from string, subject string,
```
yohamta0, line 183: Question: should we add bcc here as well or it's not needed?

internal/digraph/executor/mail.go
```diff
@@ -93,11 +96,49 @@ func (e *mail) Run(ctx context.Context) error {
...
+		bccAddresses = v
+	case []any:
+		for _, addr := range v {
+			if str, ok := addr.(string); ok && str != "" {
+				bccAddresses = append(bccAddresses, str)
+			}
+		}
+	default:
+		return fmt.Errorf("invalid type for 'bcc' field: expected string or array, got %T", v)
+	}
```
yohamta0, line 133: This code seems a bit repetitive. Could we consider to define an utility function to parse these address fields, something like below?
```go
func parseAddressField(value any) ([]string, error)
```

internal/agent/reporter.go
```diff
@@ -20,7 +20,7 @@ type Sender interface {
 }
 
 // SenderFn is a function type for sending reports.
-type SenderFn func(ctx context.Context, from string, to []string, subject, body string, attachments []string) error
+type SenderFn func(ctx context.Context, from string, to []string, cc []string, bcc []string, subject, body string, attachments []string) error
```
yohamta0, line 23: nitpick: this can be a bit simplified:
```suggestion
type SenderFn func(ctx context.Context, from string, to, cc, bcc []string, subject, body string, attachments []string) error
```

internal/digraph/builder.go
```diff
@@ -779,6 +779,46 @@ func buildMailConfig(def mailConfigDef) (*MailConfig, error) {
...
+	case []any:
+		// Multiple recipients
+		for _, addr := range v {
+			if str, ok := addr.(string); ok && str != "" {
+				bccAddresses = append(bccAddresses, str)
+			}
+		}
+	default:
+		return nil, fmt.Errorf("invalid type for 'to' field: expected string or array, got %T", v)
+	}
```
yohamta0, line 820: This code seems a bit repetitive. Could we consider to simplify this by defining an utility method?

yottahmd (commented):
Looking good! I've left some minor comments in the first pass.

internal/mailer/mailer.go
Sarvesh-11: as far as bcc is concerned i.e blind carbon copy, we will not be adding that in header, its not blind anymore.

yohamta0: Ah that’s true. Thank you for your clarification.

yohamta0:
Sorry, please let me close this at the moment as it became a bit outdated. Feel free to open a new PR if you're still interested!

PR #1150, opened 2025-07-29
chore: upgrade node to 22

reneleonhardt:
### Chores
- Upgrade node to 22

### DIFF
--- a/.github/workflows/manual-release.yaml
+++ b/.github/workflows/manual-release.yaml
@@ -30,7 +30,7 @@ jobs:
       - name: Set up NodeJS
         uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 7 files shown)


reneleonhardt:
You're welcome!

It feels a bit odd to merge a PR before CI is green, can you configure it to run automatically?

And as far as I can see, node isn't part of CI yet, even npm-publish-test.yaml doesn't build and test npm.
I recommend to test it in CI if any relevant files have been changed (like package.json).

yohamta0:
Yes, great point! The CI currently only checks Go code, so it didn’t run. I agree that we should run `npm-publish-test.yaml` and add CI for the frontend code when relevant files are updated.

Issue #1130, opened 2025-07-25
Feature request: make Dagu core components highly available


jonasban:
Dagu is unique in its choice of using the file system as a database to store the metadata unlike other schedulers out there. This gives it a competitive edge in distributed containerization environments (think K8s, Docker Swarm, Nomad etc.) that utilize multiple nodes and rely on shared storage (E.g. CephFS, GlusterS etc.) as one no longer needs to go through the hassle and pain of running and maintaining a relational DB service in such setup. Clusters typically consist of odd number of master nodes and multiple worker nodes that perform the computation.  In order for Dagu to be highly available it needs to support running multiple UI and scheduler service instances on each of the master nodes and be able to run workflows on both master and compute nodes.

I used the following Compose file to test out Dagu's HA ability:
```
services:
  ui-1:
    image: dagu:latest
    environment:
      DAGU_HOME: "/home/dagu"
    volumes:
      - ./config.yaml:/home/dagu/config.yaml
      - ./dags:/home/dagu/dags
      - ./data:/home/dagu/data
      - ./logs:/home/dagu/logs
      - ./suspend:/home/dagu/suspend
    ports:
      - 8080:8080
    command: ["server"]

  ui-2:
    image: dagu:latest
    environment:
      DAGU_HOME: "/home/dagu"
    volumes:
      - ./config.yaml:/home/dagu/config.yaml
      - ./dags:/home/dagu/dags
      - ./data:/home/dagu/data
      - ./logs:/home/dagu/logs
      - ./suspend:/home/dagu/suspend
    ports:
      - 8081:8080
    command: ["server"]

  scheduler-1:
    image: dagu:latest
    environment:
      DAGU_HOME: "/home/dagu"
    volumes:
      - ./config.yaml:/home/dagu/config.yaml
      - ./dags:/home/dagu/dags
      - ./data:/home/dagu/data
      - ./logs:/home/dagu/logs
      - ./suspend:/home/dagu/suspend
    command: ["scheduler"]

  scheduler-2:
    image: dagu:latest
    environment:
      DAGU_HOME: "/home/dagu"
    volumes:
      - ./config.yaml:/home/dagu/config.yaml
      - ./dags:/home/dagu/dags
      - ./data:/home/dagu/data
      - ./logs:/home/dagu/logs
      - ./suspend:/home/dagu/suspend
    command: ["scheduler"]
```

Sample DAG used:
```
name: data-pipeline
description: Process some data
 
[...]

PR #1163, opened 2025-08-05
feat: add zombie DAG run detection to scheduler

yottahmd:
Implements automatic detection and cleanup of zombie DAG runs - processes marked as running but whose underlying process is no longer alive.

## Changes
- Add `ZombieDetector` that periodically checks running DAG runs
- Configure via `zombieDetectionInterval` (default 45s, 0 to disable)  
- Support environment variable `DAGU_SCHEDULER_ZOMBIE_DETECTION_INTERVAL`
- Update zombie DAG runs to Error status with appropriate node error messages
- Add panic recovery and comprehensive unit tests

Related to #1130

Reported-by: jonasban

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,10 @@ For a detailed list of changes, bug fixes, and new features, please refer to the
 
 | Environment Variable | Default | Description |
 |---------------------|---------|-------------|
+| `DAGU_SCHEDULER_PORT` | `8090` | Health check server port |
+| `DAGU_SCHEDULER_LOCK_STALE_THRESHOLD` | `30s` | Scheduler lock stale threshold |
+| `DAGU_SCHEDULER_LOCK_RETRY_INTERVAL` | `5s` | Lock retry interval |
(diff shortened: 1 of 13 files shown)


jonasban:
Jesus, I have not seen an open source project move so fast. I am truly blown away. :heart:

jonasban:
One question: how expensive is this `ZombieDetector` process? 45s seems a bit long to wait and I would set it to a value within 2-5s to get faster feedback in the UI. Would this be a bad idea?

yohamta0:
@jonasban It probably won’t be too expensive if there are fewer DAG-runs like 10 for file-based storage, but that can vary depending on the hardware. The heartbeat timeout is currently 45s, so that’s how long it would take to catch a zombie run anyway. I think under normal circumstances, zombies are relatively rare, so I’m not sure it’s worth make it something like 2–5s. What do you think?

Still, it would be nice to make them configurable if user needs more control.

PR #1176, opened 2025-08-10
Add DAG configuration to lock parameters and run ID

kriyanshii:
# Summary: Add runConfig Support for DAG Execution Control

## Overview
This PR adds support for the `runConfig` feature that allows DAG authors to control user interactions when starting DAG runs through the web UI.

### Backend Changes
1. **Fixed API Response**: Updated `internal/frontend/api/v2/transformer.go` to properly include the `runConfig` field in DAG API responses
2. **Type Handling**: Fixed type conversion between internal `digraph.RunConfig` and API response types
3. **Boolean Value Preservation**: Ensured that `false` values in `runConfig` are properly included in API responses

### Frontend Changes
1. **Enhanced StartDAGModal**: Updated `ui/src/features/dags/components/dag-execution/StartDAGModal.tsx` to handle `runConfig` settings
2. **Conditional UI**: Added logic to disable parameter editing and run ID input fields based on `runConfig` values
3. **User Feedback**: Added helpful messages when editing is disabled

## New Configuration Options

DAG authors can now add a `runConfig` section to their YAML files:

```yaml
name: my-workflow
description: "Example workflow with restricted editing"

runConfig:
  allowEditParams: false  # Disable parameter editing
  allowEditRunId: false   # Disable custom run ID input

params:
  - DATE: "2024-01-01"

steps:
  - name: process
    command: ./process.sh ${DATE}
```

## Configuration Options

- **`allowEditParams`** (boolean, default: `true`): Controls whether users can edit parameters when starting the DAG
- **`allowEditRunId`** (boolean, default: `true`): Controls whether users can specify custom run IDs

## User Experience

- When `allowEditParams: false`, parameter input fields are disabled and show a message "Parameter editing is disabled for this DAG"
- When `allowEditRunId: false`, the run ID input field is disabled and shows a message "Custom run IDs are disabled for this DAG"
- When `runConfig` is not specified, all editing is allowed by default (backward compatible)

## Testing

- Created test DAG files to verify `runConfig` functionality
- Tested both `true` and `false` values for each configuration option
- Verified backward compatibility when `runConfig` is not specified

## Breaking Changes

None. This feature is fully backward compatible - existing DAGs without `runConfig` will continue to work as before with all editing enabled by default.


yohamta0:
Thank you! Let me merge the latest main.

yohamta0:
I feel that using false to disable editing is a little tricky to handle default values.
How about we flip the logic like this instead? Let me handle the change.
```yaml
runConfig:
  disableParamEdit: true
  disableRunIdEdit: true
```

PR #1273, opened 2025-09-28
feat: add params JSON Schema validation support

thefishhat:
## What does this PR do?
This PR adds JSON Schema validation support for params.

## Why is this needed?
It addresses issue https://github.com/dagu-org/dagu/issues/1182, and https://github.com/dagu-org/dagu/issues/1252.

## What changed?
- Added **JSON Schema validation** for DAG parameters - Parameters can now be validated against JSON Schema files (local or remote URLs) using [Google's JSON Schema validator](https://github.com/google/jsonschema-go)
- Introduced reserved keyword `schema` - New parameter format uses `schema` field to reference validation schemas and had to be reserved
- Maintained backward compatibility - Existing parameter formats (positional and named) continue to work unchanged alongside the new schema validation feature (with the `schema` keyword caveat above).


internal/digraph/params.go
```diff
@@ -65,8 +81,141 @@ func buildParams(ctx BuildContext, spec *definition, dag *DAG) error {
...
+	paramMap := make(map[string]any)
+	for _, pair := range paramPairs {
+		// Try to parse as JSON first, fall back to string
+		var value any
+		if err := json.Unmarshal([]byte(pair.Value), &value); err != nil {
+			// If JSON parsing fails, use as string
+			value = pair.Value
+		}
+		paramMap[pair.Name] = value
+	}
```
yohamta0, line 100: Do you think it would make sense if we validate just `map[string]string` when dealing with `[]paramPair`?
My thinking is that if we have types like `number` or `array`, then maybe we should treat the entire `params` value  as one JSON object (something we haven’t added support for yet). So we could remove `json.Unmarshal` and `json.Marshal` conversion here. What do you think?

internal/digraph/params.go
```diff
@@ -65,8 +81,141 @@ func buildParams(ctx BuildContext, spec *definition, dag *DAG) error {
...
+
+	data, err = io.ReadAll(resp.Body)
+	return data, err
+}
+
+// loadSchemaFromFile loads a JSON schema from a file path.
+func loadSchemaFromFile(filePath string) ([]byte, error) {
+	// Validate file path to prevent directory traversal attacks
+	filePath = filepath.Clean(filePath)
```
yohamta0: Could we use `fileutil.ResolvePath` instead?

internal/digraph/params.go
```diff
@@ -160,9 +309,23 @@ func parseParamValue(ctx BuildContext, input any) ([]paramPair, error) {
...
 	case map[string]any:
-		return parseMapParams(ctx, []any{v})
+		schemaRef := extractSchemaReference(v)
+		if schemaRef == "" {
+			return parseMapParams(ctx, []any{v})
+		}
+
+		values, ok := v["values"]
+		if !ok {
+			return []paramPair{}, nil // Schema-only mode, no values to validate
```
yottahmd, line 325: Super nit:
```suggestion
			return nil, nil // Schema-only mode, no values to validate
```

docs/writing-workflows/parameters.md
```diff
@@ -14,10 +14,85 @@ params:
 # Positional parameters (accessed as $1, $2, ...)
 params: first second third
 
+# JSON Schema validation
+params:
+  schema: "./schemas/params.json"  # Local file or remote URL
+  values:
```
yohamta0: Would it be helpful to include documentation explaining the precedence between the schema’s default values and the specified `values`?

yottahmd (commented):
Great work, thanks a lot! I left some comment on the first pass.

internal/digraph/params.go
thefishhat: I'm not sure I understand the problem fully. Without JSON deserialization we can't validate against a JSON schema, because of the complex types, like you said (`number`, `array`, etc.). They are then serialized back into the form they were inserted in, as a string. 

By adding more "support" for this, I assume you mean doing the deserialization on the user's behalf. I think that's a separate problem and we can get to it later. We just pass the params along to the user, who can then deal with deserialization if they want/need to. I don't think we need any more support for JSON objects than we already have for that to work. 

Without this we wouldn't be able to support the current test cases which include integer and boolean JSON schema validation. WDYT?

internal/digraph/params.go
thefishhat: Ah, missed it! Good catch on that one!

docs/writing-workflows/parameters.md
thefishhat: Definitely! I forgot to update the docs after adding the previous commit. Thanks!

thefishhat:
Pushed up a new commit addressing the concerns above. PTAL, @yottahmd ! 🙏

internal/digraph/params.go
```diff
@@ -65,8 +81,134 @@ func buildParams(ctx BuildContext, spec *definition, dag *DAG) error {
...
+	return data, err
+}
+
+// loadSchemaFromFile loads a JSON schema from a file path.
+func loadSchemaFromFile(filePath string) ([]byte, error) {
+	// Resolve the path (handles relative paths, env vars, tilde expansion)
+	resolvedPath, err := fileutil.ResolvePath(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve schema file path: %w", err)
+	}
```
yohamta0, line 204: Should we check if the file actually exists? Also, in cases where a relative path is provided, we might need to take multiple search paths into account:

1. From the current directory  
2. [From the value specified in the DAG’s `workingDir` field](https://github.com/thefishhat/dagu/blob/9dfe0194faa294e1feccf0cf94ddfd68ee34aa1e/internal/digraph/dag.go#L43)
3. [From the location where the DAG file was originally loaded](https://github.com/thefishhat/dagu/blob/9dfe0194faa294e1feccf0cf94ddfd68ee34aa1e/internal/digraph/dag.go#L46)

You can also find [a similar piece of code here.](https://github.com/thefishhat/dagu/blob/9dfe0194faa294e1feccf0cf94ddfd68ee34aa1e/internal/persistence/filedag/store.go#L427)

We could either add an argument to the function or introduce a more general resolver. What are your thoughts?

yottahmd (commented):
Thank you for the update! I just left one comment about relative path resolution.
EDIT: Let's handle the relative path resolution in a separate PR.

internal/digraph/params.go
yohamta0: Thank you for the explanation! I think it makes sense 👍

yottahmd (approved):
Great work, LGTM 🚀🚀🚀

ghansham:
Too good

PR #1308, opened 2025-10-06
feat: Add validation to prevent using script field with SSH executor

### LINKED ISSUE #1306: feat: Add validation to prevent using `script` field with SSH executor
Let's add validation to prevent using `script` field with `ssh` executor to avoid confusion. In the future we can implement this to work. Later on, we could look into implementing this functionality properly.

Examples for invalid case:
```yaml
ssh:
  xxx # ssh settings
steps:
  - script: |
      echo "some task"
```
or
```yaml
steps:
  - executor:
      type: ssh
      config:
        xxxx
    script: |
      echo "some task"
```

Valid case: Only the `command` field is supported at the moment:
```yaml
ssh:
  xxx # ssh settings
steps:
  - command "some task"
  - "some task"
```

AdityaTel89:
- Implemented extensible executor validator registry pattern
- Added SSH executor validation to reject script field
- Added comprehensive unit tests with 100% passing rate
- Maintains backward compatibility with command field

Fixes #1306

## Why This Change?

The `script` field is not currently supported with SSH executor and causes confusion. This validation provides clear error messages directing users to use the `command` field instead.

## Design Philosophy

The implementation uses a **registry pattern** to make validation adaptable to other executor types (as requested by maintainers). This makes it easy to add executor-specific validations in the future without modifying core validation logic.


internal/digraph/executor_validator.go
```diff
@@ -0,0 +1,73 @@
...
+			"script field is not supported with SSH executor. " +
+				"Use 'command' field instead. " +
+				"See: https://github.com/dagu-org/dagu/issues/1306",
+		)
+	}
+	return nil
+}
+
+// DockerExecutorValidator validates Docker executor configurations
+type DockerExecutorValidator struct{}
```
yohamta0: What do you think about implementing the validator interface directly within each executor?  
This approach would remove the need for a separate struct and the `RegisterExecutorValidator` function.  
We could then use type assertions to check whether an executor implements the validator interface (see: https://go.dev/ref/spec#Type_assertions).

To complete this refactor, it may make sense to relocate the executor registry to the `digraph` package:  
https://github.com/dagu-org/dagu/blob/71c5d0f6366e079d370caf865420199ec806ebda/internal/digraph/executor/executor.go#L54-L72

internal/digraph/executor_validator.go
```diff
@@ -0,0 +1,73 @@
+package digraph
+
+import "fmt"
+
+// ExecutorValidator validates executor-specific configurations
+type ExecutorValidator interface {
```
yohamta0: How about renaming it to `StepValidator`?
```suggestion
type StepValidator interface {
```

yottahmd (commented):
Looking great, thanks! I have left two comments. Please take a look when you have a chance.

internal/digraph/executor_validator.go
AdityaTel89: Thanks for the feedback! I've implemented your suggested approach:

 Changes Made:
1. Removed separate validator structs - No more `SSHExecutorValidator`, `DockerExecutorValidator`, etc.
2. StepValidator interface in executor package** - Added to `internal/digraph/executor/executor.go`
3. Direct implementation - SSH executor now implements `StepValidator` directly in `ssh.go`
4. Type assertion pattern - Validation uses `executor.(StepValidator)` to check if validation is implemented

 Implementation:
 `// In ssh.go
func (e *sshExec) ValidateStep(step *digraph.Step) error {
if step.Script != "" {
return fmt.Errorf("script field is not supported with SSH executor...")
}
return nil
}`

The validation is now integrated directly into the executor, eliminating the need for a separate registry!

Regarding moving the executor registry to the digraph package - I can see the benefit for centralization. However, I ran into an import cycle issue when trying to call executor validation from `builder.go`. Currently, I'm doing a simple check in the builder:

`if step.ExecutorConfig.Type == "ssh" && step.Script != "" {
return error...
}`


Would you prefer I pursue moving the executor registry as a follow-up PR, or is the current implementation sufficient for this issue?

internal/digraph/executor_validator.go
AdityaTel89: I've renamed `ExecutorValidator` to `StepValidator` in `internal/digraph/executor/executor.go`. You're absolutely right - the interface validates steps, not executors themselves, so the name is now more accurate.

The interface is now:
`type StepValidator interface {
ValidateStep(step *digraph.Step) error
}`


Also, I've removed the old `internal/digraph/executor_validator.go` file as it's no longer needed with the refactored approach.

internal/digraph/executor/executor.go
```diff
@@ -29,6 +29,12 @@ type NodeStatusDeterminer interface {
 	DetermineNodeStatus(ctx context.Context) (status.NodeStatus, error)
 }
 
+// StepValidator is an interface for executors that validate step configurations.
+// Executors can optionally implement this interface to enforce executor-specific constraints.
+type StepValidator interface {
```
yohamta0, line 34: It may make sense to place this in `digraph` package where it will be used.

yohamta0:
@AdityaTel89 Would you like to work on a follow-up refactoring for enabling to use executor's `ValidateStep` functions resolving the import cycle issue? I'd appreciate it if you can. Thanks!

PR #1371, opened 2025-10-29
feat(spec): resolve relative jsonschema path

thefishhat:
## What does this PR do?
This PR adds JSON Schema validation support for params.

## Why is this needed?
It addresses issue https://github.com/dagu-org/dagu/issues/1291.

## What changed?
- Updated `buildParams` to resolve declared schema paths by checking the DAG’s cwd/`WorkingDir`/DAG file directory

I couldn't find any docs to update, the previous ones have been deleted I think. r? @yottahmd

### DIFF
--- a/internal/core/spec/params.go
+++ b/internal/core/spec/params.go
@@ -64,14 +64,16 @@ func buildParams(ctx BuildContext, spec *definition, dag *core.DAG) error {
 	}
 
 	// Validate the parameters against a resolved schema (if declared)
-	if resolvedSchema, err := resolveSchemaFromParams(spec.Params, spec.WorkingDir, dag.Location); err != nil {
-		return fmt.Errorf("failed to get JSON schema: %w", err)
-	} else if resolvedSchema != nil {
-		updatedPairs, err := validateParams(paramPairs, resolvedSchema)
-		if err != nil {
-			return err
+	if !ctx.opts.Has(BuildFlagSkipSchemaValidation) {
+		if resolvedSchema, err := resolveSchemaFromParams(spec.Params, spec.WorkingDir, dag.Location); err != nil {
+			return fmt.Errorf("failed to get JSON schema: %w", err)
+		} else if resolvedSchema != nil {
+			updatedPairs, err := validateParams(paramPairs, resolvedSchema)
+			if err != nil {
+				return err
+			}
+			paramPairs = updatedPairs
 		}
-		paramPairs = updatedPairs
 	}
 
 	for _, paramPair := range paramPairs {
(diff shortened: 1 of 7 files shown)


internal/core/spec/params.go
```diff
@@ -61,10 +62,11 @@ func buildParams(ctx BuildContext, spec *definition, dag *core.DAG) error {
 		overrideParams(&paramPairs, overridePairs)
 	}
 
-	// Validate the parameters against the provided schema, if it exists
```
yohamta0, line 65: We might want to optimize this by checking the `BuildContext.BuildOpts.NoEval` flag (which is set to `false` only when the DAG executes) so that it returns without resolving or validating the JSON schema. This would prevent resolving it on every listing API call, which could impact performance when there are many DAGs with JSON schemas. What do you think?

ghansham:
I feel json schema should be evaluated only when actual execution starts, not even when it is enqueued.

internal/core/spec/params.go
thefishhat: Ah, is this code path exercised outside of DAG execution? I definitely think this should only run on execution exclusively, as @ghansham also pointed out.

Is checking `BuildContext.BuildOpts.NoEval` the best way of doing it? Or should this validation move altogether to a different part of the codebase?

yohamta0: Yes, this is the configuration to specify whether a field is treated as "metadata"`. If it is, the field will be parsed dynamically (yes, inefficient 😅 but this helps avoid the complexity of caching and reloading that is sometimes confusing in Airflow).
https://github.com/dagu-org/dagu/blob/230f7a0423fc77dfa3252e433e9d53bab7a0ad58/internal/core/spec/builder.go#L74-L101

I think checking `BuildOpts.NoEval` is currently most reliable way for now. We could rename the flag to make it more clear in future refactor.

>  Or should this validation move altogether to a different part of the codebase?

Good idea! I’m not entirely sure where the best place would be, but perhaps core/spec/loader.go or runtime/agent.go.
I’m fine with keeping it as-is for now and refactoring later if needed.

thefishhat: I think `runtime/agent.go` is probably the ideal place for it, but seeing as we'd need to thread down a bunch of metadata (like schemaref, etc.) through `core.DAG`, perhaps simply checking `BuildOpts.NoEval` is better, because it's simpler and we already have a paved path for it.

One small issue with that is things like `dagu validate` run `spec.WithoutEval()`. I would expect a missing/malformed JSON Schema to trigger an error during validate. Would `OnlyMetadata()` provide more accurate isolation of the run flows we want? 

If I'm reading this correctly:
* list APIs have both `OnlyMetadata` and `NoEval`
* validate has only `NoEval`
* execution has neither

WDYT?

yohamta0: Ah good point! I agree that validate should check for malformed JSON schemas. The `OnlyMetadata` flag could also be useful, but on the DAG details page in the Web UI, it loads the full data (i.e., `OnlyMetadata = false`) and I’m not sure if we should validate the JSON schema every time a user opens the DAG details page.
Maybe we could refactor these flags into bit flags to allow more granular control for different use cases?

yohamta0:
@thefishhat Sorry, I overlooked the question about the doc! I've moved out the documentation into a separate repository here: https://github.com/dagu-org/docs 

I welcome your PR to the doc, or I can take care of that, either way is totally fine.

thefishhat:
I refactored the flags as bitmasks, could you please have a look at whether I set the `SkipSchemaValidation` appropriately for the flows that are interested in it? @yottahmd

PR #1373, opened 2025-10-31
Fix: [Windows] Failure to Execute DAG from Queue

### LINKED ISSUE #1372: [Windows] Failure to Execute DAG from Queue
### Issue Description

When I start a DAG without using the enqueue feature, it runs fine. However, when I try to start the DAG in enqueue mode, it gets queued but never runs. This issue occurs only on Windows; on Linux, it runs fine.

### Error Log

I see this error in the log, which keeps repeating:

```log
level=ERROR msg="Failed to execute dag" err="command failed: exit status 1\noutput: Initialization error: could not determine home directory: %userprofile% is not defined"
```

### Proposed Cause
I think the issue is related to the environment variable filter "defaultWhitelist" in "internal/common/config/env.go", which doesn’t include Windows variables like USERPROFILE.

lvoeg:
## Description

Fixes a missing environment variables required to run an enqueued DAG.

## Additional notes

I cross-checked the environment variables between Windows and Linux. These three appear to be the only ones needed to successfully run an enqueued DAG.
I verified the fix by testing both the Cmd and PowerShell execution.

## Related Issue
Fix  #1372

### DIFF
--- a/internal/common/config/env.go
+++ b/internal/common/config/env.go
@@ -23,6 +23,9 @@ var defaultWhitelist = map[string]bool{
 	"TZ":              true,
 	"SHELL":           true,
 	"LD_LIBRARY_PATH": true,
+	"USERPROFILE":     true, // Windows Specifics equivalent of HOME
+	"PSModulePath":    true, // Windows Specifics, powershell modules path
+	"Path":            true, // Windows Specifics equivalent of PATH
 }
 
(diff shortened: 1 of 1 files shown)


yottahmd (approved):
@lvoeg Thank you very much for fixing this and testing on Windows and Linux!

PR #1381, opened 2025-11-01
feat(ui): allow rescheduling DAG-run

yottahmd:
- Add functionality for retrying a DAG with a new Run ID

Ref https://github.com/dagu-org/dagu/issues/1347
Co-author: @kriyanshii

image


ghansham:
It may be named reschedule dag.

Issue #1380, opened 2025-11-01
Mermaid component zoom out bug


ghansham:
When we zoom out the dag rendering, the vertical scroll bar adjusts according to zoom level but horizontal bar does not. This becomes a little awkward when we are working very wide dags

kriyanshii:
working on it.

kriyanshii:
this can be closed

yohamta0:
Yes, thanks a lot for fixing the issue!

PR #1383, opened 2025-11-02
fixed zoom out dag rendering #1380

kriyanshii:
this commit fixes dag ui rendering while zooming out. 
Fixes - #1380


yottahmd (approved):
LGTM, thank you very much!

Issue #1391, opened 2025-11-04
[Feat] Extend `jq` executor for raw-output option


yottahmd:
We can extend the jq executor for raw output option.
```yaml
- executor:
     type: jq
     config:
       raw: true
     command: .foo
     script: |
        {"foo": "bar"}
```
Expect: `bar`
Current: `"bar"`

file: https://github.com/dagu-org/dagu/blob/main/internal/runtime/builtin/jq/jq.go

yohamta0:
Fixed in #1392
Thanks a lot!

PR #1392, opened 2025-11-04
feat: Extended jq executor for raw-output option 

kriyanshii:
Fixes - #1391


yohamta0:
@kriyanshii Thank you very much! It looks great. Would it be possible to add some tests?

kriyanshii:
done!

Issue #1389, opened 2025-11-03
[Bug] Show DateTime of the run for each Execution (RepeatPolicy)


jeremydelattre59:
Hi,
Thanks for the job in [#1041](https://github.com/dagu-org/dagu/issues/1041#issuecomment-3482313923)
There's a bug. When I try to view the execution list, I don't see the date and time. When I switch to another window and then return to my browser, the date and time appear. That's when there's a call to https://dagu.homelinux.maison/api/v2/dag-runs/Cyclique_Pair/019a4b17-2781-7250-b247-ac6a2005451b/sub-dag-runs?remoteNode=local
I think it's necessary to call periodically this url.
 
https://github.com/user-attachments/assets/25f1c799-7394-468d-96f0-29ad97dfa472

At the same time, the list of executions is not updated. When a new execution is running (or done), i need to switch focus window to see it. Number of executions is 10 but I can see only 3 to 10 (1 and 2 disappear) and date time is not change for each line (date time are for 1 to 8 apply to 3 to 10).

https://github.com/user-attachments/assets/c6ad675e-7db9-4df6-9e74-ae4e380c57f3

When I refresh, I can see all executions and when I switch windows and return to my browser, I have good date and time.

<img width="945" height="491" alt="Image" src="https://github.com/user-attachments/assets/86e3f177-9951-4748-9ef3-8d64c6df2a8f" />

 Regards,
Jeremy

yohamta0:
Fixed in #1396. Will soon release a patch version. Thank you for reporting the issue!

PR #1396, opened 2025-11-08
ui: refresh sub dag runs list timestamps

yottahmd:
- poll `/sub-dag-runs` while the repeat list is expanded so timestamps and counts update without focusing the window
- force an immediate refresh whenever the underlying sub-run ids change so new executions show up right away
- reuse the shared duration formatter in `DAGHeader` and tighten typings in `SubDAGRunsList` to satisfy lint

Issue #1389


jeremydelattre59:
It's work fine ! Thanks

PR #1403, opened 2025-11-11
fix(docker/client): dont clear entrypoint for step-level docker executor

### LINKED ISSUE #1402: Entrypoint is overwritten in docker executor
Hello,

Normally, there are a lot of images that has an entrypoint like `binary` and then in the command we only need to supply `run` instead of the whole `binary run`. I would like to ask why do we need overwrite the entrypoint here ?
https://github.com/dagu-org/dagu/blob/c56cdf4fc0996de9423f0b76e8987228e711ba75/internal/runtime/builtin/docker/client.go#L485-L486

As always, thank you very much for creating this awesome Dagu :D

vnghia:
Fixes #1402.

Hi @yottahmd, this is my take for this issue. We only clear the container entrypoint for dag-level container and keep the step-level one

### DIFF
--- a/internal/runtime/builtin/docker/client.go
+++ b/internal/runtime/builtin/docker/client.go
@@ -203,7 +203,7 @@ func (c *Client) CreateContainerKeepAlive(ctx context.Context) error {
 	c.cancel = cancel
 	c.cancelMu.Unlock()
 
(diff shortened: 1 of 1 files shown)


yottahmd (approved):
LGTM, thank you very much!

PR #1405, opened 2025-11-11
status: rename canceled state to aborted across CLI, API, and UI

yottahmd:
- rename the user-facing “canceled” state to “aborted” in the public API specs, generated clients, and JSON schema so API consumers see the updated vocabulary
- align CLI/status output, the agent TUI, HTML reporter, and docs with the new label, including styling for the aborted badge
- update the React dashboard/views and associated tests/telemetry to surface “aborted” metrics consistently

feedback from @ghansham

### DIFF
--- a/internal/core/execution/dagrun_test.go
+++ b/internal/core/execution/dagrun_test.go
@@ -12,8 +12,9 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
-// Mock implementations for testing
+var _ execution.DAGRunStore = (*mockDAGRunStore)(nil)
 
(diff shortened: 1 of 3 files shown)


ghansham:
We may need to update migration tool for taking care of already existing dag definitions and dag runs.
Are we updating lifecycle hook (onCancelled to onAborted) also?

yohamta0:
@ghansham Thanks for the heads up! I think we don’t need a migration because it uses the same status enum (3), and it seems the status label wasn’t actually persisted in the DAG run data, so it shouldn’t be an issue.

PR #1420, opened 2025-11-23
feat(ui): Running/failed steps list on dag runs page 

### LINKED ISSUE #1401: Running/failed steps list on dag runs page
For the running and failed dags, we can have the running/failed step names viewed on the dag runs page so that user can have a quick status of the dag (at which step right now the execution is) without going to dag details page.

I request all to  give their feedback on this if any.

kriyanshii:
fixes #1401

api/v2: add running and failed step names to DAG runs page

Add display of running and failed step names on the DAG runs page
to provide quick status visibility without navigating to details.

Backend changes:
- Add runningStepNames and failedStepNames fields to DAGRunSummary
  schema in api/v2/api.yaml
- Update toDAGRunSummary transformer to extract step names from
  nodes with NodeRunning and NodeFailed status
- Handle all NodeStatus cases explicitly to satisfy exhaustive linter

Frontend changes:
- Create StepNamesDisplay component with responsive behavior:
  - Desktop: tooltip on hover showing all step names
  - Mobile: clickable "+n" indicator showing toast/snackbar
- Update DAGRunTable and DAGRunGroupedView to use new component
- Enhance SimpleToast to support multi-line messages for step lists
- Add configurable toast timeout (default 3000ms)

The implementation shows single step names directly, and for multiple steps displays "stepname +n" format with expandable details via tooltip (desktop) or toast (mobile), providing a clean UX that doesn't interfere with row navigation.

this is how it looks like - 
image

When more then one steps fail/running. 
image

refined ui:
image

### DIFF
--- a/ui/src/features/dag-runs/components/dag-run-list/StepDetailsTooltip.tsx
+++ b/ui/src/features/dag-runs/components/dag-run-list/StepDetailsTooltip.tsx
@@ -1,8 +1,5 @@
 import React from 'react';
-import {
-  components,
-  NodeStatus,
-} from '../../../../api/v2/schema';
+import { components, NodeStatus } from '../../../../api/v2/schema';
 import {
   Tooltip,
   TooltipContent,
@@ -23,11 +20,7 @@ function getStepName(node: Node, index: number) {
   return node.step?.name || `Step ${index + 1}`;
 }
 
-function renderStepList(
-  title: string,
-  nodes: Node[],
-  colorClass: string
-) {
+function renderStepList(title: string, nodes: Node[], colorClass: string) {
   if (nodes.length === 0) {
     return null;
   }
@@ -38,9 +31,7 @@ function renderStepList(
 
   return (
     <div className="space-y-1">
-      <div className={`text-[11px] font-semibold ${colorClass}`}>
-        {title}
-      </div>
+      <div className={`text-[11px] font-semibold ${colorClass}`}>{title}</div>
       <ul className="text-xs space-y-0.5">
         {visibleSteps.map((node, idx) => (
           <li key={`${node.step?.name || idx}-${idx}`}>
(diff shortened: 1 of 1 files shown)


ghansham:
Include a snapshot where multiple steps failed, when dag is running (multiple steps in parallel). This is list view of dag run. Also include for the grouped view. When multiple steps failed or running, you can include one stepname mentioned followed by (...) as a clickable hide/show component.

kriyanshii:
> Include a snapshot where multiple steps failed, when dag is running (multiple steps in parallel). This is list view of dag run. Also include for the grouped view. When multiple steps failed or running, you can include one stepname mentioned followed by (...) as a clickable hide/show component.

Adde relatable snapshots in the pull request. Thank you for the tip.

ghansham:
Add one more for group view

ghansham:
Can we include the skipped/failed steps for partial success dags as well.

ghansham:
I feel bulleted list may be better than numbered list in tooltip/toast.
@yottahmd , @kriyanshii  What do you think?

yohamta0:
I agree with the bullets since the numbers aren't really needed. I feel like showing the step name under the badge might make the row heights look a bit uneven. Maybe we could move it to a separate column and use '...' if the text is too long?

ghansham:
I am not sure about separate column as it is relevant only for failed, running and partially successful cases like. We can also think of putting (...) besides status (failed, running, partial success) to avoid uneven row height. This may help reducing impact on row height. Or we can also think of making these three status values as link on which if user hovers(desktop) or clicks (mobile), the tooltip/toast comes up. 
And I think we should make the popup components scrollable to take care of large number of steps if it happens to be so.

yohamta0:
Yes, we can show the details in a tooltip on hover instead of displaying them by default. This also saves us from adding fields to the API response, as we can fetch the step details via /api/v2/dag-runs/{name}/{dagRunId} only when the user hovers over the status badge.

ghansham:
Agree with @yottahmd . That will be better from performance point of view. And then this may be needed for three status values (partially successful, failed and running), only.

kriyanshii:
this architecture makes more sense as it is lightweight. Pushing the changes

ghansham:
Please include partial success case also.

ghansham:
Please include screenshots for list view  and group view after the latest updates.

yottahmd (approved):
Sorry for taking this long, it looks great! Thank you very much!

PR #1421, opened 2025-11-23
feat(cmd): add positional argument to `dequeue` command

yottahmd:
Updated the `dequeue` command to accept `<queueName>` as a positional argument.

**New Syntax:**
`dagu dequeue [--run-id=<dagName:runId>] <queueName>`

- `--run-id` is optional


ghansham:
That will be really cool for singleton dags. For non singleton, I think it will dequeue all the runids corresponding to the dagname. Right?

yohamta0:
> That will be really cool for singleton dags. For non singleton, I think it will dequeue all the runids corresponding to the dagname. Right?

Actually, the current implementation does not differentiate whether it's a singleton or not. It just dequeues the first item if `--run-id` is not provided.

ghansham:
Still it will be a big plus

Issue #976, opened 2025-06-01
Model Context Protocol (MCP) Support


yottahmd:
I'd like to explore integrating the [Model Context Protocol (MCP)](https://github.com/modelcontextprotocol/servers) into Dagu to enable AI agents, such as Claude Code, to create, execute, and update DAGs dynamically. MCP is an open standard that facilitates seamless integration between AI models and external tools or data sources.

**Benefits**:
- **Dynamic Workflow Management**: AI agents can programmatically manage DAGs, allowing for more adaptive and responsive workflows.
- **Standardized Integration**: MCP offers a consistent interface for AI agents to interact with Dagu, reducing the need for bespoke integrations.
- **Scalability**: Dagu MCP server enables AI agents execute complex workflow in parallel at scale in distributed environment.

**Future Enhancements**:
- **Human-in-the-Loop Interactions**: Introduce checkpoints within workflows that require human approval or input, ensuring critical decisions are overseen by humans.
- **Agent-to-Agent (A2A) Communication**: Enable AI agents to coordinate and delegate tasks among themselves, enhancing efficiency in complex workflows.

I'm open to collaborating on this initiative and would appreciate any feedback or suggestions.

SiwonP:
@yottahmd I'm quite interested in participating to this dev ! I guess the main resource is the [go-sdk](https://github.com/modelcontextprotocol/go-sdk). 
What do you have in mind ? Starting a MCP server as an independent process ?
What tools could be added ? 
- Creating a DAG
- Listing DAGs
- Executing DAGs
- Updating DAGs

yohamta0:
Hi @SiwonP,

Thanks for your interest. I’d be really happy to have your help with development.

Yes, we should use the official SDK. We could either modify the existing server process to expose a JSON RPC endpoint alongside the REST API, or run it on a separate port. I’m not sure yet which approach is better for us.

All of those tools look useful. We’ll also need something for listing DAG runs (execution history). That said, all of these features can be accessed through HTTP requests using curl from an AI agent like Claude Code, so I’m still not completely sure how much additional value MCP would give us. I feel like there might be other possibilities worth exploring. What do you think?

SiwonP:
@yottahmd I feel like a strict separation between the existing REST API and a future MCP server is right : that is, to use an entirely different process and port. 
Good point on the "Listing DAG runs", it seems valuable to expose this.
[...]

PR #1444, opened 2025-12-04
Model Context Protocol (MCP) server implementation

SiwonP:
Implementation of an MCP server for Dagu #976

## Summary by CodeRabbit

* **New Features**
  * Adds an "mcp" CLI command to run an MCP server.
  * Provides tools to list DAGs, list DAG runs, and fetch a DAG (via local API).
  * Runs over standard I/O for easy integration with pipelines.
  * Prevents concurrent runs via directory locking and respects runtime options from config/env.
  * Exposes a CLI entrypoint to start and manage the MCP server.


jyroscoped:
* I would recommend removing the hardcoded values in `listDags()`. The function currently has hardcoded `host`, `port`, and `api_base_url` with commented-out environment variable lookups. These should either use environment variables or be configurable through the MCPServer struct to avoid brittleness.
* It looks like the `mcpFlags` slice is empty in `mcp.go`. Consider adding relevant flags (e.g., for configuring the host, port, or other MCP server settings) if they're needed, or document why no flags are necessary.
* I would recommend avoiding the use of `context.Background()` in the `Start()` method. Since the method receives a `ctx` parameter, you should pass that context to `s.server.Run()` instead of creating a fresh background context, which will ignore cancellation signals from the caller.
* It looks like the commented-out `executeDag()` and `createDag()` functions should either be completed and uncommented or removed entirely to reduce code clutter and maintenance burden.
* I would recommend storing the `config` field in the MCPServer struct (it's passed to `New()` but never stored), as you may need it later for accessing configuration values or passing to other methods.
* It looks like the `running` atomic field is declared but never used. Consider removing it if it's not needed, or implement start/stop logic if it is meant to track the server's state.

yohamta0:
Please let me close this PR for now. I need more thinking whether we need MCP or not.

PR #1447, opened 2025-12-05
[ui] stop step retry while dag is running.

kriyanshii:
1. ui fix to stop step retry while dag is running.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry controls are now disabled while a DAG run is actively running, preventing accidental retry attempts.
* **Style**
  * Disabled retry buttons show reduced opacity and a not-allowed cursor for clearer visual feedback.

### DIFF
--- a/internal/persistence/filequeue/queuefile.go
+++ b/internal/persistence/filequeue/queuefile.go
@@ -354,22 +354,11 @@ func parseQueueFileName(path, fileName string) (ItemData, error) {
 	}
 	timestamp = timestamp.Add(time.Duration(nanos) * time.Nanosecond)
 
-	// Read the JSON file to get the actual DAG name
-	var itemData ItemData
-	fileData, err := os.ReadFile(path) // nolint: gosec
-	if err != nil {
-		return ItemData{}, fmt.Errorf("failed to read queue file %s: %w", path, err)
-	}
-
-	if err := json.Unmarshal(fileData, &itemData); err != nil {
-		return ItemData{}, fmt.Errorf("failed to unmarshal queue file %s: %w", path, err)
-	}
-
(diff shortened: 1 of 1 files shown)


ghansham:
After the completion of dag, the button should remain disabled for the succesful steps, it should be enabled only for failed steps or those steps which user marks as failed.

internal/persistence/filequeue/queuefile.go
```diff
@@ -354,11 +354,22 @@ func parseQueueFileName(path, fileName string) (ItemData, error) {
...
+	if err := json.Unmarshal(fileData, &itemData); err != nil {
+		return ItemData{}, fmt.Errorf("failed to unmarshal queue file %s: %w", path, err)
+	}
+
+	// Create the ItemData struct with the correct DAG name from the file
 	item := ItemData{
 		FileName: fileName,
 		DAGRun: execution.DAGRunRef{
-			Name: filepath.Base(filepath.Dir(path)),
+			Name: itemData.DAGRun.Name,
```
yohamta0: @kriyanshii I'm afraid this might not be the right fix. Could you please revert the changes for `queuefile.go`? 
-- EDIT --
Actually, let me check what's the right fix first.

yohamta0: ~@kriyanshii I have not been able to reproduce the queueing issue right now. Could you please update the PR to focus on the fix for the step button control?~

I'm able to repro the issue now.

kriyanshii:
@yottahmd done

ghansham:
Even after the dag is successful, we should keep them disabled if step is successful. It should be enabled only if step has failed or marked as failed.

PR #1451, opened 2025-12-06
feat(spec): add support for negate condition

yottahmd:
## Summary by CodeRabbit

* **New Features**
  * Added an optional "negate" flag for preconditions at both DAG and step levels to invert condition evaluation.
  * API and schema updated to include the new negate field.

* **Tests**
  * Expanded test coverage for negated preconditions across DAGs, steps, commands, environment variables, and error scenarios.

### DIFF
--- a/internal/runtime/condition_test.go
+++ b/internal/runtime/condition_test.go
@@ -11,161 +11,142 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestCondition_Eval(t *testing.T) {
+func newTestContext() context.Context {
+	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


ghansham:
Rather than putting a negate field, I suggest should have introduced an operator field which could have been more versatile (values could be `eq, ne, gt, ge, lt, le` for `==, !=,  >, >=, <, <=`, respectively). That would have made it explicit and clear. What do you think?

yohamta0:
@ghansham Great suggestion, but I’m concerned that handling complex logical conditions might overcomplicate the YAML format. If complex logic is needed, users can handle it via Python or Bash scripts. I’d prefer to keep the YAML syntax minimal at the moment.

PR #1455, opened 2025-12-06
feat(spec): add init handler field

yottahmd:
## Summary by CodeRabbit

* **New Features**
  * Added an Init handler that runs (after preconditions) before workflow steps; its status is surfaced and failure prevents steps from running.
  * Added a canonical Abort handler and retained Cancel as deprecated.
  * Runtime and status now include OnInit nodes; option to skip inheriting base handlers for sub-DAG runs.

* **Documentation**
  * Schema updated to document init and abort hooks and lifecycle ordering.

* **Tests**
  * New integration tests for Init, abort/cancel behavior, skipping base handlers, and lifecycle paths.

### DIFF
--- a/internal/integration/handler_on_test.go
+++ b/internal/integration/handler_on_test.go
@@ -35,7 +35,7 @@ steps:
 				require.NotNil(t, dag.HandlerOn.Init)
 				require.Equal(t, "onInit", dag.HandlerOn.Init.Name)
 			},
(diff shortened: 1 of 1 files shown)


ghansham:
Can we create variables derived from parameters in the initHandler? Some example will be appreciable.

yohamta0:
@ghansham Yes, you can use `output` field to define variable same as regular steps.

PR #1481, opened 2025-12-14
fix(runtime): include queue name in dequeue command

kriyanshii:
The SubCmdBuilder.Dequeue method was missing the queue name parameter when building the dequeue command. The CLI command expects the queue name as the first positional argument after "dequeue", but it was not being included, causing dequeue operations via the API to fail.

This change:
- Updates Dequeue to use dag.ProcGroup() to get the queue name
- Includes the queue name as the first argument after "dequeue"
- Updates tests to verify the queue name is included
- Adds test case for DAGs with custom queue names

Fixes the issue where dequeuing from the API failed because the queue name was not being passed to the CLI command.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dequeue command now correctly incorporates queue name from DAG configuration, allowing proper queue identification during execution.

### DIFF
--- a/internal/runtime/subcmd.go
+++ b/internal/runtime/subcmd.go
@@ -104,8 +104,9 @@ func (b *SubCmdBuilder) Enqueue(dag *core.DAG, opts EnqueueOptions) CmdSpec {
 }
 
 // Dequeue creates a dequeue command spec.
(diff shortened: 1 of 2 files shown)


ghansham:
Also check if the dag run is assigned to default queue. Do regression testing for default case via api route

yottahmd (approved):
Thank you very much, LGTM! 🚀🚀🚀

PR #1485, opened 2025-12-14
fix (scheduler) fix queue maxConcurrency initialization from config

kriyanshii:
When queues are discovered dynamically (when items are enqueued), they were being auto-created with maxConcurrency=1 by default, even if the queue was configured in the config file with a higher value. This caused only 1 DAG to run at a time instead of the configured maxConcurrency value.

The fix ensures that:
1. When a new queue is discovered, it checks the config first and uses the configured maxConcurrency value if found
2. Existing queues that were created before config was loaded are updated to use config values on subsequent processing cycles
3. Added debug logging to help diagnose queue configuration issues

This ensures queue-level concurrency (maxConcurrency=n) works correctly, allowing n DAGs to run concurrently from the same queue as configured.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced queue initialization and processing with improved concurrency management based on configuration and thread-safety improvements for better system reliability and performance.


yohamta0:
@kriyanshii Thanks a lot for opening the PR! I might be wrong, but I believe the issue below has already been fixed in the latest version.

> When queues are discovered dynamically (when items are enqueued), they were being auto-created with maxConcurrency=1 by default, even if the queue was configured in the config file with a higher value. This caused only 1 DAG to run at a time instead of the configured maxConcurrency value.

ghansham:
@yottahmd The observations look to be on the latest main branch based on timeline. Still if you think we can confirm by running a relevant test case on the latest release (1.26.4) to be double sure.

internal/service/scheduler/queue_processor.go
```diff
@@ -114,7 +117,7 @@ func NewQueueProcessor(
 		conc := max(queueConfig.MaxActiveRuns, 1)
 		p.queues.Store(queueConfig.Name, &queue{
```
yohamta0, line 118: It seems that the global queue config is already stored in the queue list here. If I understand correctly, this PR might not result in any actual changes.

kriyanshii:
It looks like i might have worked on a stale commit. I didn't pull after my pr. I should have understood when I got merge conflict 🥲

ghansham:
Thanks @kriyanshii 
Sorry for the confusion @yottahmd

kriyanshii:
Gimme some time to test the latest commit tho

kriyanshii:
closing due to based on stale commit. sorry

yohamta0:
No worries! Thank you very much.

PR #1493, opened 2025-12-20
feat(all): Add support for multiple-level nested local DAGs

yottahmd:
## Summary by CodeRabbit

* **New Features**
  * Optional parentSubDAGRunId query to fetch nested sub-DAG runs; sub-run lists and modal show root/parent DAG context, status filters with counts, and improved keyboard navigation.

* **Bug Fixes / UX**
  * Consolidated local DAG handling for multi-level execution, improved error responses for missing runs, and small UI spacing tidy-ups.

* **Tests**
  * Reorganized and expanded integration/unit tests for multi-level sub-DAG scenarios and temp-DAG file handling.

### DIFF
--- a/internal/service/frontend/api/v2/dags.go
+++ b/internal/service/frontend/api/v2/dags.go
@@ -531,6 +531,13 @@ func (a *API) GetDAGDAGRunDetails(ctx context.Context, request api.GetDAGDAGRunD
 
 	dagStatus, err := a.dagRunMgr.GetCurrentStatus(ctx, dag, dagRunId)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ghansham:
We need to take care of those cases where the multiple dags defined within a single dag does not contain a dag whose name matches with another dag yaml or a dag defined in another file containing multiple dags. Orvis it that these multiple subdags defined within a single subdag are qualified by the dagyaml name to avoid name conflict (like namespaces in c++ and packages in java). If it is not so, either we should do that or we should a avoid multiple dag definitions within a single dag . I hope there is no confusion.

PR #1506, opened 2025-12-25
feat(all): multiple commands syntax

yottahmd:
## Summary by CodeRabbit

* **New Features**
  * Steps can contain multiple commands executed sequentially; UIs and runners display and run command lists.

* **Validation**
  * Executors advertise capabilities and produce clearer errors when features conflict or are unsupported (multi-command vs sub-workflow/script/container).

* **API**
  * Step payloads now use a commands list for command entries; legacy single-command fields remain supported for compatibility.

* **Integration & UX**
  * Schemas, APIs, reporting, and UI components updated to present and handle the commands array.

### DIFF
--- a/internal/runtime/builtin/ssh/ssh.go
+++ b/internal/runtime/builtin/ssh/ssh.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


ghansham:
All looks good. One point, the granularity of stepwise error tracking may not be available. And the short hand syntax for steps actually had similar effect.

PR #1592, opened 2026-01-18
feat(#1589): add shell config option to container

### LINKED ISSUE #1589: Generic/shell entrypoint overwrite for container executor
New `startup:` option to run commands with arbitrary entrypoint (overwrite current). This is what docker compose [entrypoint](https://docs.docker.com/reference/compose-file/services/#entrypoint) does. This addresses this limitation from the [docs](https://github.com/dagu-org/docs/blob/main/writing-workflows/container.md#execution-model-and-entrypoint-behavior):

> **Entrypoint/CMD not used for steps:** Because steps are executed with `docker exec`, your image’s `ENTRYPOINT` or `CMD` are not invoked for step commands. Steps run directly in the running container process context.

as if the containers is started in image mode (not exec) with the entrypoint overwrite commands are executed by the entrypoint.

Example use-case: Run commands in a shell with custom options (for debug or etc.)

How you can do it know:

```yaml
container:
  image: debian
  pullPolicy: true

steps:
  - command: /bin/bash -o errexit -o xtrace -o pipefail -c 'echo "container running with custom entrypoint"'
  - command: /bin/bash -o errexit -o xtrace -o pipefail -c 'echo "this one too"'
```

Proposal:

```yaml
# Startup: command - run custom startup command
container:
  image: debian
  startup: custom
  command: ["/bin/bash", "-o", "errexit", "-o", "xtrace", "-o", "pipefail", "-c"]

steps:
  - command: echo "container running with custom entrypoint"
  - command: echo "this one too"
```

If you think about it this is equivalent to:

```yaml
- name: shell debug mode
  command:
    - echo "container running with custom entrypoint"
    - echo "this one too"
  shell: /bin/bash -o errexit -o xtrace -o pipefail -c
```

with host executor.

Wouldn't this also allow for `script:` field support when using `startup: entrypoint / custom (or whaterver name will be)`? Contents of `script:` field can again as with host executor be bundled into single file and executed by the entrypoint.

yottahmd:
Resolves #1589 

## Summary by CodeRabbit

* **New Features**
  * Configurable shell wrapper for container step execution (supports advanced shell flags, multi-command patterns) in both exec and image modes.
  * Runtime will honor configured shell wrappers when running step commands.

* **Configuration / Schema**
  * Added a documented "shell" array to container config/schema to declare shell wrappers.

* **Tests**
  * Extensive tests added covering shell wrapping, flag injection, chaining, pipes, platform shells, exec-mode, and backward compatibility.

### DIFF
--- a/internal/runtime/builtin/docker/client.go
+++ b/internal/runtime/builtin/docker/client.go
@@ -253,19 +253,21 @@ func getContainerStatus(info types.ContainerJSON, err error) string {
 		}
 		return fmt.Sprintf("error: %v", err)
 	}
-	if info.State == nil {
+
+	state := info.State
+	if state == nil {
 		return "unknown"
 	}
-	if info.State.Running {
+	if state.Running {
 		return "running"
 	}
-	if info.State.Status != "" {
-		if info.State.ExitCode != 0 {
-			return fmt.Sprintf("%s (exit code: %d)", info.State.Status, info.State.ExitCode)
-		}
-		return info.State.Status
+	if state.Status == "" {
+		return "not running"
 	}
-	return "not running"
+	if state.ExitCode != 0 {
+		return fmt.Sprintf("%s (exit code: %d)", state.Status, state.ExitCode)
+	}
+	return state.Status
 }
 
 // Close closes the container client and cleans up resources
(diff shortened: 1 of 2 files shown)


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

PR #1594, opened 2026-01-21
feat: support model fallback of chat step

yottahmd:
## New Features
* Added multi-model fallback support for LLM configurations—users can now specify multiple models in DAG definitions with automatic fallback to the next model if one fails.
* Per-model configuration overrides now available, including temperature, max tokens, top-p, base URL, and API key settings.


yohamta0:
@coderabbitai review

Issue #1609, opened 2026-01-25
Add a url (file path) field to the /dag-runs/enqueue API.
This will allow DAGs to be enqueued and executed using only the DAG file path.


sahalbelam:

ghansham:
This will be an alternative to inline spec ingest where dags are very large leading to long inline enqueue urls. Either of the spec or url will be required at any moment. It will raise an error if both are specified or none of them are specified. Right now it will support only files locatable on the system on which dag is expected to run (local disk or nfs paths). But we keep it open for extension to support other protocols (like s3, ftp/sftp, http, https, etc) based on user needs. This will also help to ingest dag definitions not stored in `dags_dir`.

ghansham:
Closeable

yohamta0:
Yes thanks a lot!

PR #1614, opened 2026-01-28
added url (filepath) for /dag-runs/enqueue api

sahalbelam:
Adding a **url** field allows DAGs to be executed directly from different file locations. If a file path is provided using url, the **spec** field becomes **optional** and is not required. #1609 

## Release Notes

* **New Features**
  * Enhanced DAG run enqueuing with support for external file paths. Users can now specify DAG specifications via external file references or provide them inline when enqueueing DAG runs. Note: inline specifications and file paths are mutually exclusive in a single request.

### DIFF
--- a/internal/service/frontend/api/v2/dagruns.go
+++ b/internal/service/frontend/api/v2/dagruns.go
@@ -183,21 +183,8 @@ func (a *API) EnqueueDAGRunFromSpec(ctx context.Context, request api.EnqueueDAGR
 	}
 
 	var finalSpec string
-	baseDir, err := os.UserHomeDir()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


ghansham:
@yottahmd, we can add similar functionality to start dag api similar to enqueue api.

sahalbelam:
we can define new dags directory in config.yaml like
~/.config/dagu/config.yaml
paths:
  dagsDir: /path/to/dags/dir

for the request we can just give the dag name instead of full path

ghansham:
We plan to introduce alternative dags directory which can be used by filepath for picking dags. This is useful where user does not want to expose dag definitions to user, specifically useful for batch processing.

env: 
       custom_dagsDir: /path/to/dags/dir

PR #1623, opened 2026-02-02
feat: self-upgrade command

yottahmd:
## Release Notes

* **New Features**
  - Added new upgrade command to check for and install new versions with support for target version selection, dry-run mode, and backup creation
  - Added automatic update checking with caching to reduce repeated checks
  - Added UI notification banner when a new version is available

### DIFF
--- a/internal/upgrade/upgrade_test.go
+++ b/internal/upgrade/upgrade_test.go
@@ -1,9 +1,22 @@
 package upgrade
 
 import (
+	"archive/tar"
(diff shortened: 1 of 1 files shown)


ghansham:
In airgapped systems, we should have a configuration to switch this off

yohamta0:
In airgapped systems, it just does not work silently.

ghansham:
Will it attempt and then timeout?

PR #1630, opened 2026-02-04
ingest yaml with filename

sahalbelam:
now dags can only be ingest with the name.
user has to define ALT_DAGS_DIR variable in base.yaml 
Spec is still working. #1609 

## Summary by CodeRabbit

* **New Features**
  * DAG run enqueue now supports filename-based references: Users can trigger DAG execution by specifying an existing YAML DAG filename, in addition to providing inline specifications. The enqueue endpoint accepts either a spec or filename, but not both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ Full documentation is available at [docs.daguit.dev](https://docs.daguit.dev/).
 |---------------------|---------|-------------|
 | `DAGU_HOME` | - | Base directory that overrides all path configurations |
 | `DAGU_DAGS_DIR` | `~/.config/dagu/dags` | Directory for DAG definitions |
+| `DAGU_ALT_DAGS_DIR` | - | Additional directory to search for DAG definitions |
 | `DAGU_LOG_DIR` | `~/.local/share/dagu/logs` | Directory for log files |
 | `DAGU_DATA_DIR` | `~/.local/share/dagu/data` | Directory for application data |
 | `DAGU_SUSPEND_FLAGS_DIR` | `~/.local/share/dagu/suspend` | Directory for suspend flags |
--- a/internal/cmd/context.go
+++ b/internal/cmd/context.go
@@ -377,10 +377,16 @@ type dagStoreConfig struct {
 
 // dagStore returns a new DAGRepository instance.
 func (c *Context) dagStore(cfg dagStoreConfig) (exec.DAGStore, error) {
+	// Merge configured alternate DAGs directory into search paths if provided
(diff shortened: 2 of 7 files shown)


internal/service/frontend/api/v2/dagruns.go
```diff
@@ -171,8 +179,64 @@ func (a *API) EnqueueDAGRunFromSpec(ctx context.Context, request api.EnqueueDAGR
...
+		obj := make(map[string]any)
+		err = yaml.Unmarshal(baseData,&obj) 
+		if err != nil {
+			return nil, fmt.Errorf("error unmarshalling base config: %w", err)
+		}
+
+		altDagDir := obj["env"].([]any)[0].(string)
+		fmt.Println("ALDAGDIR",altDagDir)
+
+		if strings.HasPrefix(altDagDir, "ALT_DAGS_DIR=") {
```
yohamta0: I'm thinking we should use ALT_DAGS_DIR to be set in `SearchPaths` field of DAG store.
https://github.com/dagu-org/dagu/blob/e05708d1e27f451b109bad70fdcaa58196c2bf8e/internal/persis/filedag/store.go#L34
The fix would be adding `ALT_DAGS_DIR` to config package, update docs for the new environment variable, and adding ALT_DAGS_DIR value to the `SearchPaths`.  In this case there's no need to fix this API. What do you think?

ghansham: Yes that will be a much cleaner and systematic approach. Will work on that

sahalisro-blip:
for this, config.yaml will have one more variable altDagDir under paths: which will be used for locating the alternate dags 
right @yottahmd ?

yohamta0:
@sahalisro-blip Yes, I think that's correct.

README.md
```diff
@@ -290,6 +290,7 @@ Full documentation is available at [docs.dagu.cloud](https://docs.dagu.cloud/).
 |---------------------|---------|-------------|
 | `DAGU_HOME` | - | Base directory that overrides all path configurations |
 | `DAGU_DAGS_DIR` | `~/.config/dagu/dags` | Directory for DAG definitions |
+| `DAGU_ALT_DAGS_DIR` | - | Additional directory to search for DAG definitions |
```
ghansham, line 292: Please add default path

`~/.config/dagu/alt_dags`

yohamta0: Let's keep default value empty to avoid surprises for those users who are not familiar with this config.

PR #1634, opened 2026-02-05
feat: dollar sign escape syntax

### LINKED ISSUE #1628: No way to escape Dollar sign ($) in many places, it is always being interpreted as a variable
There are many places that are automatically trying to replace everything that begins with dollar sign to variable, if the variable is not found it is by default replaced with empty string. 

## Some examples:

- Putting params

Image

Image

- HTTP body, query, headers etc.
```
- name: fetch
    type: http
    command: POST https://example.com/
    config:
      debug: true
      body: '{"test": "$test"}'
      headers:
        Authorization: $test
        Content-Type: application/json
```
Request sent:
```
~~~ REQUEST ~~~
POST  /  HTTP/1.1
HOST   : example.com
HEADERS:
	Accept: application/json
	Authorization:
	Content-Type: application/json
	User-Agent: go-resty/2.16.5 (https://github.com/go-resty/resty)
BODY   :
"eyJ0ZXN0IjogIiJ9" # -> {"test": ""}
```

And many more.
The problem is, I found no easy way to prevent that.
[Docs are mentioning single quotes](https://docs.daguit.dev/writing-workflows/environment-variables#basic-syntax), but that does not work (as seen in the http example where `body` is put in the single quotes.
I tried using:
- single quotes:`'$test'` -> `` 
- escaping by backslash: `\$test` -> `\$test`
- using double dollar: `$$test` -> `test` # this one removes both dollar signs, probably because first dollar sign is interpreted as a variable of name `$`
- escaping by slash: `/$test` -> `/`
- using brackets: `{$}test` -> `{$}test` # this one preserves the sign but also preserves brackets.
- using hex value: `\x24test` -> `\x24test`
- using hex value in quotes: `"\x24test"` -> `` # interpreted as dollar sign and then replaced by eval
- using pipe `| \n $test` -> ``
- using `> \n $test` -> ``

There was probably more, but nothing resulted in a result: `$test`.

## Expected result
There is a way to prefix a word with dollar sign. There are many real life scenarios that require that:
- GraphQL queries
- Prices $10.99
- Passwords

yottahmd:
Resolves #1628 

## Summary by CodeRabbit

* **New Features**
  * Dollar-sign escaping: a backslash before $ (e.g. \ $) yields a literal $ in non-shell evaluation contexts; shell-executed commands keep previous behavior.
  * Escaping enabled by default with an option to disable it.

* **Tests**
  * Added comprehensive unit and integration tests covering escaping, quoting, and shell vs non-shell scenarios.

* **Documentation**
  * RFC describing dollar-sign escaping behavior and examples.

### DIFF
--- a/internal/cmn/cmdutil/shellutil.go
+++ b/internal/cmn/cmdutil/shellutil.go
@@ -23,7 +23,8 @@ func HasShellArgs(shell []string) bool {
 func IsShellValueSet(shellValue any) bool {
 	switch v := shellValue.(type) {
 	case string:
-		return !isDirectShell(strings.TrimSpace(v)) && strings.TrimSpace(v) != ""
+		trimmed := strings.TrimSpace(v)
+		return trimmed != "" && !isDirectShell(trimmed)
 	case []string:
 		return HasShellArgs(v)
 	case []any:
--- a/internal/cmn/cmdutil/shellutil_test.go
+++ b/internal/cmn/cmdutil/shellutil_test.go
@@ -22,7 +22,6 @@ func TestHasShellArgs(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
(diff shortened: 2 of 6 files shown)


yohamta0:
@coderabbitai full review again

PR #1635, opened 2026-02-06
Fix/windows process tree and timeout fix

prods:
# Fix: Enforce timeoutSec on Windows and prevent process hangs

## Problem Description

When `timeoutSec` is set on a DAG step, processes that exceed the timeout should be forcibly terminated. However, on Windows, some processes would continue running for hours even when `timeoutSec` was set (e.g., to 900 seconds).

### Root Cause

The issue had two components:

1. **Blocking `cmd.Wait()`**: The `commandExecutor.Run()` method blocked indefinitely on `cmd.Wait()` without listening to context cancellation. When the timeout expired, the process would not be killed because the code was stuck waiting for `Wait()` to return.
  
2. **Incomplete process termination on Windows**: The Windows `KillProcessGroup()` function only killed the immediate process, leaving child processes (subprocess tree) orphaned and running.
  

### 1. `internal/runtime/builtin/command/command.go`

Modified the `Run()` method to:

- Run `cmd.Wait()` in a goroutine with a result channel
- Use `select` to watch for both context cancellation and command completion
- Forcibly kill the process when context is cancelled (timeout)
- Return exit code 124 (standard timeout exit code) on timeout

### 2. `internal/cmn/cmdutil/cmd_windows.go`

Updated `KillProcessGroup()` to:

- Use `killProcessTree()` instead of just `cmd.Process.Kill()`
- Ensure child processes are also terminated on Windows
- Match the behavior used in `KillMultipleProcessGroups()` and Unix implementation

## Impact

- **timeoutSec** now works reliably on all platforms, especially Windows
- Processes that hang due to I/O blocking or zombie states are properly terminated
- Child processes on Windows are no longer orphaned when the parent is killed
- Exit code 124 is returned for timeout scenarios (consistent with Unix timeout behavior)

## Files Changed

1. `internal/runtime/builtin/command/command.go` - Core timeout enforcement fix
2. `internal/cmn/cmdutil/cmd_windows.go` - Windows process tree termination fix

## Bug Fixes
* Improved Windows subprocess cleanup to properly terminate the complete process tree instead of only the immediate target process, ensuring full resource cleanup.
* Enhanced command execution with context-aware timeout handling to prevent hung processes, explicit termination when context is cancelled, and improved error reporting that includes recent stderr information for better diagnostics.


yottahmd (approved):
LGTM, thank you so much for the improvement!

PR #1642, opened 2026-02-07
feat: default execution mode

yottahmd:
## Summary by CodeRabbit

* **New Features**
  * Added configurable server-wide default execution mode (local or distributed) to control how DAGs are executed by default
  * Added ability to force local execution of individual DAGs using `workerSelector: local` in DAG specifications
  * Improved consistency of execution dispatch decisions across all execution paths

### DIFF
--- a/charts/dagu/README.md
+++ b/charts/dagu/README.md
@@ -34,7 +34,7 @@ The chart deploys four components:
 
 - **Coordinator**: gRPC server for distributed task execution (port 50055)
 - **Scheduler**: Manages DAG execution schedules (port 8090 for health)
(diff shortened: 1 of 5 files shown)


ghansham:
Should we hide the distributed execution related UI elements based on this?

yohamta0:
@ghansham No, I don't think so. The setting is just default, so distributed execution can be used even `defaultExecutionMode = local`.

ghansham:
> @ghansham No, I don't think so. The setting is just default, so distributed execution can be used even `defaultExecutionMode = local`.

Yeah Later I realized. It is default mode

PR #1645, opened 2026-02-07
fix: Replace TerserPlugin with esbuild.

yonas:
TerserPlugin can cause the build process to segfault due to insufficient memory.

Fixes https://github.com/dagu-org/dagu/issues/901 again.

## Summary by CodeRabbit

* **Chores**
  * Optimized build tooling for faster compilation and improved minification performance.


yottahmd (approved):
LGTM! Thank you so much for the fix!

PR #1665, opened 2026-02-13
fix(#1658): parallel + call doesn't split expanded positional params

### LINKED ISSUE #1658: `parallel:` + `call:` Doesn't Split Expanded Params
# Dagu `parallel:` + `call:` Doesn't Split Expanded Params

## Issue Summary

**Affected Versions**: 1.30.3 (likely all versions)
**Severity**: Critical - params silently mangled or dropped

## Problem

When a `call:` step uses `parallel:` with `items:`, params passed via YAML variable expansion are not correctly split into individual `KEY=VALUE` pairs. If an expanded variable contains spaces (i.e., multiple params), dagu treats the expanded result as a single value rather than splitting it.

```yaml
steps:
  - name: run-bundles
    call: exec-bundle.yaml
    parallel:
      items: ${BUNDLE_ITEMS}
    params: "BUNDLE_NAME=${ITEM.name} BUNDLE_PARAMS=${ITEM.params_str}"
```

If `${ITEM.params_str}` expands to `DAG_FOO=bar DAG_BAZ=qux`, the full `params:` string becomes:

```
BUNDLE_NAME=install-node BUNDLE_PARAMS=DAG_FOO=bar DAG_BAZ=qux
```

**Expected**: Dagu splits this into three params:
- `BUNDLE_NAME` = `install-node`
- `BUNDLE_PARAMS` = `DAG_FOO=bar`
- `DAG_BAZ` = `qux`

**Actual**: The splitting is unreliable. `BUNDLE_PARAMS` may receive the entire remainder of the string (`DAG_FOO=bar DAG_BAZ=qux`), or `DAG_BAZ` may be silently dropped. The behavior is inconsistent.

This means you cannot pass a variable number of params through `parallel:` + `call:` using a single expanded string.

## Why This Matters

The natural dagu pattern for running sub-workflows with dynamic params is `call:` + `params:`. This works fine when all values are known at YAML-authoring time or come from simple single-token expansions. It breaks when a param value itself contains spaces, or when the number of params varies per iteration item.

In a specific case, each bundle has a different set of params (`DAG_K3S_VERSION=1.31`, `DAG_NODE_VERSION=22.13.1 DAG_GCS_INSTALLER_PATH=installers/node`, etc.). These need to be forwarded from the parent workflow to the child bundle workflow.

## Symptoms

- Child workflow receives wrong param values
- Params after a space-containing expansion are silently dropped
- No error or warning from dagu
- Difficult to diagnose because the child workflow starts and runs -- just with wrong/missing values

## Reproduction

```yaml
# parent.yaml
params:
  - ITEMS  # JSON array
steps:
  - name: run-children
    call: child.yaml
    parallel:
      items: ${ITEMS}
    params: "NAME=${ITEM.name} EXTRA=${ITEM.extra_params}"
```

```yaml
# child.yaml
params:
  - NAME
  - EXTRA
steps:
  - name: check
    command: bash -c "echo NAME=${NAME} EXTRA=${EXTRA}"
```

```bash
# Start with an item where extra_params contains a space
dagu start parent.yaml -- 'ITEMS=[{"name":"test","extra_params":"A=1 B=2"}]'
```

`child.yaml` will not receive `NAME=test`, `EXTRA=A=1`, and `B=2` as three separate params. The splitting is unreliable.

## Specific Workaround: Single Nushell Script

Abandoned `parallel:` + `call:` entirely. Instead, `exec-bundles.yaml` calls a single nushell script (`exec-bundles.nu`) that handles everything: decoding the bundles JSON, fetching each bundle, running each bundle's workflow via `dagu start --run-id`, and querying the dagu REST API to retrieve each run's output variables.

### Flow

```
exec-bundles.yaml
  └─ exec-bundles.nu
       For each bundle:
         1. Fetch bundle via get_bundles.nu
         2. Write args.env (params with DAG_ prefix)
         3. dagu start --run-id=<id> workflow.yaml -- DAG_KEY=val DAG_KEY2=val2
         4. http get localhost:8080/api/v2/dag-runs/<name>/<id>/outputs
       Output: JSON array of [{name, status, outputs}, ...]
```

Each bundle runs as a separate top-level DAG run. The nushell script queries the dagu API after each run to retrieve output variables (e.g., `RESULT` from `verify-installation.nu`), which `dagu start` CLI does not expose via stdout.

### Why not `call:`?

Dagu's `call:` step would be the natural way to invoke sub-workflows, but:

- `call:` + `parallel:` doesn't split space-separated params from variable expansion (this issue)
- `dagu start` CLI doesn't return workflow output variables to stdout (see #1660)
- Using a single nushell script gives full control over param passing and output collection

### Trade-offs

- Bundle runs don't appear as child nodes in the parent workflow's dagu UI
- Each bundle shows up as an independent run in the DAG history
- The nushell script queries the dagu API (localhost) to retrieve outputs

## References

- Parent workflow: `core/workflows/exec-bundles.yaml`
- Related issue: #1660 (dagu 1.30.3 `--` separator requirement)

yottahmd:
Resolves #1658 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parameter handling so expanded variable references are preserved during splitting, fixing incorrect parameter values in parallel steps.
  * Prevented progress updates from overwriting final run status and ensured stale "Running" states are detected and marked appropriately.

* **Tests**
  * Added end-to-end and unit tests covering conditional parameter quoting, parallel parameter expansion, and stale-running status detection.

### DIFF
--- a/internal/runtime/manager.go
+++ b/internal/runtime/manager.go
@@ -324,6 +324,11 @@ func (m *Manager) GetLatestStatus(ctx context.Context, dag *core.DAG) (exec.DAGR
 				st = currentStatus
 			} else {
 				logger.Debug(ctx, "Failed to get current status from socket", tag.Error(err))
+				// Socket is unavailable while persisted status is still running.
+				// Verify process liveness and downgrade stale running status.
+				if err := m.checkAndUpdateStaleRunningStatus(ctx, attempt, st); err != nil {
+					logger.Error(ctx, "Failed to check and update stale running status", tag.Error(err))
+				}
 			}
 		}
 	}
--- a/internal/runtime/manager_test.go
+++ b/internal/runtime/manager_test.go
@@ -155,6 +155,28 @@ steps:
 		err := cli.UpdateStatus(ctx, root, status)
 		require.Error(t, err)
 	})
+	t.Run("GetLatestStatusMarksStaleRunningAsFailed", func(t *testing.T) {
+		dag := th.DAG(t, `steps:
+  - name: "1"
+    command: sleep 1
+`)
+
+		dagRunID := uuid.Must(uuid.NewV7()).String()
+		now := time.Now()
+		ctx := th.Context
+
+		att, err := th.DAGRunStore.CreateAttempt(ctx, dag.DAG, now, dagRunID, exec.NewDAGRunAttemptOptions{})
+		require.NoError(t, err)
+		require.NoError(t, att.Open(ctx))
+
+		runningStatus := testNewStatus(dag.DAG, dagRunID, core.Running, core.NodeRunning)
+		require.NoError(t, att.Write(ctx, runningStatus))
+		require.NoError(t, att.Close(ctx))
+
+		latest, err := th.DAGRunMgr.GetLatestStatus(ctx, dag.DAG)
+		require.NoError(t, err)
+		require.Equal(t, core.Failed, latest.Status)
+	})
 }
 
 func testNewStatus(dag *core.DAG, dagRunID string, dagStatus core.Status, nodeStatus core.NodeStatus) exec.DAGRunStatus {


yohamta0:
@coderabbitai full review

PR #1666, opened 2026-02-13
fix(#1661): expand env vars for non-Unix shells (PowerShell/cmd)

### LINKED ISSUE #1661: `${VAR}` Command Interpolation Broken on Windows
# Dagu `${VAR}` Command Interpolation Broken on Windows

## Issue Summary

**Affected Versions**: 1.30.3 (likely all versions with shell detection)
**Severity**: Critical on Windows — dotenv values silently expand to empty strings in command strings
**Platform**: Windows only (Linux/macOS unaffected)

## Problem

When a dagu workflow loads variables via `dotenv:`, those variables cannot be referenced with `${VAR}` in `command:` strings on Windows. They silently expand to empty strings.

```yaml
dotenv:
  - envs/default.env   # BUCKET_PREFIX=gs://
  - envs/args.env       # PROJECT_BUCKET=my-bucket

steps:
  - name: get-bundles
    command: nu scripts/get_bundles.nu --bucket ${BUCKET_PREFIX}${PROJECT_BUCKET}
    # Expected: --bucket gs://my-bucket
    # Actual:   --bucket (empty — nushell error: missing flag argument)
```

Params defined in the `params:` block work fine — only `dotenv:` values are affected.

## Root Cause

Dagu's command executor disables its own `${VAR}` expansion whenever a shell is detected, assuming the shell will handle variable references natively.

In `internal/runtime/builtin/command/command.go`:

```go
GetEvalOptions: func(ctx context.Context, step core.Step) []cmdutil.EvalOption {
    env := runtime.GetEnv(ctx)
    if len(env.Shell(ctx)) > 0 {
        return []cmdutil.EvalOption{cmdutil.WithoutExpandEnv()}
    }
    return nil
},
```

When `WithoutExpandEnv()` is set, `EvalString` in `internal/runtime/env.go` excludes `rCtx.Envs` (which contains dotenv values) from the variable substitution scope:

```go
if option.ExpandEnv {
    opts = append(opts, cmdutil.WithVariables(e.Envs))
    opts = append(opts, cmdutil.WithVariables(e.Variables.Variables()))
    opts = append(opts, cmdutil.WithVariables(rCtx.SecretEnvs))
    opts = append(opts, cmdutil.WithVariables(rCtx.Envs))       // dotenv values HERE
} else {
    opts = append(opts, cmdutil.WithVariables(e.Envs))           // step-level only
    opts = append(opts, cmdutil.WithVariables(e.Variables.Variables()))  // params only
    // rCtx.Envs EXCLUDED — dotenv values not available
}
```

**On Linux** this is fine: dagu detects `sh` as the default shell and passes dotenv values to the subprocess via `cmd.Env`. When `sh -c "command ${VAR}"` runs, `sh` natively expands `${VAR}` from its process environment.

**On Windows** dagu detects PowerShell as the default shell (via `getWindowsDefaultShell()` in `internal/common/cmdutil/cmdutil.go`). PowerShell uses `$env:VAR` syntax — not `${VAR}` — to reference environment variables. So `${VAR}` is interpreted as a PowerShell variable (which doesn't exist), and expands to nothing.

The `env:` forwarding workaround also fails:

```yaml
env:
  - FWD_BUCKET: ${PROJECT_BUCKET}   # Also expands to empty
```

Because the `env:` block values go through the same `EvalString` code path with `WithoutExpandEnv()`.

## What Works vs. What Doesn't

Tested on Windows Server 2022 with dagu 1.30.3, nushell 0.109.1:

| Mechanism | `${VAR}` in command string | Available in subprocess `$env` |
|---|---|---|
| `dotenv:` values | **BROKEN** (empty) | **Works** |
| `params:` values | Works | Works |
| `env:` with literal | N/A | Works |
| `env:` forwarding `${DOTENV_VAR}` | **BROKEN** (empty) | Empty string |
| Step output `${step.stdout}` | Works | Works |
| `workingDir: ${VAR}` | Works (build-time) | N/A |

### Key insight: dotenv values ARE in the subprocess environment

Despite failing in command string interpolation, dotenv values are correctly passed to the subprocess via `cmd.Env` (through `AllEnvs()` which includes `DAG.Env`). Nushell scripts can read them with `$env.VAR_NAME`. This is the basis for the workaround.

## Reproduction

```yaml
# test-interpolation.yaml
name: test-interpolation
dotenv:
  - test.env    # Contains: MY_VALUE=hello
params:
  - MY_PARAM: from-param
steps:
  - name: test-dotenv
    command: nu -c "print $'got: ${MY_VALUE}'"
    # Windows: prints "got: " (empty)
    # Linux:   prints "got: hello"
  - name: test-param
    command: nu -c "print $'got: ${MY_PARAM}'"
    # Both: prints "got: from-param"
  - name: test-env-read
    command: nu -c "print $'got: ($env.MY_VALUE)'"
    # Both: prints "got: hello" — subprocess env works
```

## Suggested Fix

The `GetEvalOptions` callback should be platform/shell-aware. When the detected shell is PowerShell (or any shell that doesn't support `${VAR}` syntax), dagu should keep its own expansion enabled.

### Option A: Shell-aware expansion (minimal change)

```go
GetEvalOptions: func(ctx context.Context, step core.Step) []cmdutil.EvalOption {
    env := runtime.GetEnv(ctx)
    shell := env.Shell(ctx)
    if len(shell) > 0 {
        shellName := strings.ToLower(filepath.Base(shell[0]))
        // Only disable expansion for shells that natively handle ${VAR}
        switch {
        case shellName == "sh", shellName == "bash", shellName == "zsh",
             strings.HasSuffix(shellName, "/sh"),
             strings.HasSuffix(shellName, "/bash"),
             strings.HasSuffix(shellName, "/zsh"):
            return []cmdutil.EvalOption{cmdutil.WithoutExpandEnv()}
        default:
            // PowerShell, cmd.exe, nu, etc. don't expand ${VAR}
            // Keep dagu's own expansion enabled
            return nil
        }
    }
    return nil
},
```

### Option B: Always expand, let shells re-expand (safest)

Remove `WithoutExpandEnv()` entirely. If a shell also expands `${VAR}`, the value is already substituted so re-expansion is a no-op (the literal value contains no `${...}` patterns). This is safe as long as values don't contain literal `${...}` text, which is rare.

```go
GetEvalOptions: func(ctx context.Context, step core.Step) []cmdutil.EvalOption {
    // Always let dagu expand ${VAR} — shells will see literal values
    return nil
},
```

### Option C: Use WithoutExpandEnv only for `e.Variables` overlap

The original intent was likely to avoid double-expansion of `params:` values (which go through `e.Variables` and are always in scope). A more surgical fix would be to always include `rCtx.Envs` (dotenv values) in the expansion scope, regardless of shell:

```go
// In EvalString, always include rCtx.Envs:
if option.ExpandEnv {
    // ... existing full expansion
} else {
    opts = append(opts, cmdutil.WithVariables(e.Envs))
    opts = append(opts, cmdutil.WithVariables(e.Variables.Variables()))
    opts = append(opts, cmdutil.WithVariables(rCtx.Envs))  // ADD THIS
}
```

## References

- Source: `internal/runtime/builtin/command/command.go` — `GetEvalOptions` callback
- Source: `internal/runtime/env.go` — `EvalString()`, `Shell()` methods
- Source: `internal/common/cmdutil/cmdutil.go` — `GetShellCommand()`, `getWindowsDefaultShell()`

yottahmd:
Fixes #1661 

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Improved variable expansion handling for edge cases involving single quotes and adjacent characters
  * Enhanced consistency in variable substitution across different shell contexts

* **Tests**
  * Added comprehensive test coverage for variable expansion edge cases, including variables followed by single quotes and missing variable scenarios

### DIFF
--- a/internal/cmn/eval/envscope.go
+++ b/internal/cmn/eval/envscope.go
@@ -172,8 +172,12 @@ func expandWithLookup(s string, lookup func(key string) (string, bool)) string {
 		var key string
 		if loc[2] >= 0 { // Group 1: ${...}
 			key = s[loc[2]:loc[3]]
-		} else { // Group 2: $VAR
+		} else if loc[4] >= 0 { // Group 2: $VAR
 			key = s[loc[4]:loc[5]]
+		} else {
+			// Neither group captured — preserve original text.
+			b.WriteString(match)
+			continue
 		}
 
 		if val, found := lookup(key); found {
--- a/internal/cmn/eval/resolve.go
+++ b/internal/cmn/eval/resolve.go
@@ -111,20 +111,12 @@ func (r *resolver) resolveJSONSource(name string) (string, bool) {
 	return r.lookupScopeNonOS(name)
 }
 
-// extractVarKey extracts the variable key from a regex match.
-// Returns the key and false if the match is single-quoted.
-func extractVarKey(match string) (string, bool) {
-	if match[0] == '\'' && match[len(match)-1] == '\'' {
-		return "", false
-	}
-	if strings.HasPrefix(match, "${") {
-		return match[2 : len(match)-1], true
-	}
-	return match[1:], true
-}
-
 // isSingleQuotedVar reports whether the matched variable token is enclosed
 // in single quotes in the original input (e.g., '${VAR}' or '$VAR').
+// Note: this is a simple adjacent-character heuristic. It may misidentify
+// nested quote contexts such as 'foo'${BAR}'baz' where the quote before
+// ${BAR} actually closes a prior segment. This is acceptable for the
+// targeted use cases (nu shell $'...' syntax, simple quoting).
 func isSingleQuotedVar(input string, start, end int) bool {
 	return start > 0 && end < len(input) && input[start-1] == '\'' && input[end] == '\''
 }
@@ -152,8 +144,12 @@ func (r *resolver) replaceVars(template string) string {
 		var key string
 		if loc[2] >= 0 { // Group 1: ${...}
 			key = template[loc[2]:loc[3]]
-		} else { // Group 2: $VAR
+		} else if loc[4] >= 0 { // Group 2: $VAR
 			key = template[loc[4]:loc[5]]
+		} else {
+			// Neither group captured — preserve original text.
+			b.WriteString(match)
+			continue
 		}
 
 		if strings.Contains(key, ".") {
--- a/internal/cmn/eval/resolve_test.go
+++ b/internal/cmn/eval/resolve_test.go
@@ -406,62 +406,6 @@ func TestReplaceVarsWithScope(t *testing.T) {
 	}
 }
 
-// --- extractVarKey ---
-
-func TestExtractVarKey(t *testing.T) {
-	tests := []struct {
-		name   string
-		match  string
-		wantK  string
-		wantOK bool
-	}{
-		{
-			name:   "BracedVar",
-			match:  "${FOO}",
-			wantK:  "FOO",
-			wantOK: true,
-		},
-		{
-			name:   "SimpleDollarVar",
-			match:  "$FOO",
-			wantK:  "FOO",
-			wantOK: true,
-		},
-		{
-			name:   "SingleQuotedBraced",
-			match:  "'${FOO}'",
-			wantK:  "",
-			wantOK: false,
-		},
-		{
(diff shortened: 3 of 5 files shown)


yohamta0:
@coderabbitai full review

PR #1669, opened 2026-02-14
feat: singleton option for enqueue API

sahalbelam:
Whats in this PR?

1️⃣ Add Singleton Support to /dag-runs/enqueue
Implemented a singleton constraint for the /dag-runs/enqueue endpoint.
Ensures that only one DAG with the same name can either:
Be running, or
Exist in the queue at any given time.
Prevents duplicate DAG executions and avoids unintended concurrent runs.
This improves execution safety and ensures predictable DAG behavior.

2️⃣ Require filename for /dag-runs/enqueue (with altDagsDir support)
Added support for running DAGs explicitly using a filename.
DAG execution via filename requires altDagsDir to be defined in config.yaml.
This ensures DAGs are resolved from the correct alternate directory when specified.
Note: The altDagsDir configuration was previously merged in an earlier PR, but the filename enforcement logic was unintentionally omitted.
This PR completes that missing portion.

3️⃣ Fix Makefile Color Output
Updated all echo commands in the Makefile to use the -e flag.
Ensures ANSI escape sequences (e.g., \033[0;32m) render correctly in terminal output.
Restores proper colored logging for better CLI readability.

## Summary by CodeRabbit

* **New Features**
  * DAG run enqueueing now supports specifying a DAG via filename reference in addition to inline YAML specifications
  * Added singleton mode to prevent concurrent execution of the same DAG run


sahalbelam:
@yottahmd  already fixed the makefile 😅

yohamta0:
Thank you very much, but since we have added the `ALT_DAGS_DIR` config, I don't think the adding the parameter is necessary. Also the Makefile is already fixed in the main branch. Sorry about racing for the same fix!

ghansham:
@sahalbelam request you to push pr only for singleton. And  request you to create multiple PRs for individual issues.

sahalbelam:
what about the singleton in /dag-runs/enqueue? 
@yottahmd

yohamta0:
@sahalbelam Yes, could you please update the PR to include only necessary changes for singleton?

yohamta0:
So let me close this in favor of #1672

Issue #1673, opened 2026-02-14
Retrying dags with global queue configured dont respect global queue capacity


ghansham:
When a dag with global queue configured fails and after sometime the dag is retried by the operator, if the global queue is full, still dag starts running breaching the queue capacity and UI becomes inconsistent showing 4(running dags)/3(capacity). The retry job should respect queue capacity.

PR #1676, opened 2026-02-14
scheduler: respect global queue capacity on retry

kriyanshii:
When a DAG with a global queue was retried (API or CLI), the retry ran immediately and bypassed queue capacity checks. This allowed more DAGs to run than configured (e.g. 4 running / 3 capacity) and led to inconsistent UI display.

Retries for DAGs using global queues are now enqueued instead of executed immediately. The queue processor picks them up when capacity is available, so retries honor queue limits.

- API RetryDAGRun: enqueue retry when DAG uses global queue
- CLI dagu retry: same enqueue behavior for global-queue DAGs
- Add config.FindQueueConfig for shared queue lookup
- Step retry still runs immediately (queue processor has no step info)

Fixes: #1673

## Release Notes

* **New Features**
  * DAG retries now support global queue integration: when a global queue is configured, retries are enqueued as low-priority operations and processed based on system capacity rather than executing immediately. Queue status, timestamps, and trigger information are now tracked for enhanced visibility and management of retry operations.

### DIFF
--- a/internal/core/exec/enqueue_retry_test.go
+++ b/internal/core/exec/enqueue_retry_test.go
@@ -30,7 +30,7 @@ func TestEnqueueRetry(t *testing.T) {
 			dag:      &core.DAG{Name: "test-dag"},
 			status:   &exec.DAGRunStatus{Status: core.Queued},
 			dagRunID: "run-1",
-			setupMocks: func(att *exec.MockDAGRunAttempt, qs *exec.MockQueueStore) {
+			setupMocks: func(_ *exec.MockDAGRunAttempt, _ *exec.MockQueueStore) {
 				// No mock setup needed — function should return early
 			},
 			assertMocks: func(t *testing.T, att *exec.MockDAGRunAttempt, qs *exec.MockQueueStore) {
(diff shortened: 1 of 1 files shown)


yohamta0:
Thank you so much for the PR! Looking very good. I'll add some tests to this and merge later.

ghansham:
Thanks a ton @kriyanshii for addressing this issue. I have a suggestion, can we make it a bit flexible like there is between start and enqueue. I mean when we reschedule, api should get a flag whether to enqueue or not. From developer/operator perspective, if waiting queue is populated with a lot jobs and user wants to debug a particular dagrun, he should be able to run it immediately. Default behavior should be enqueue no doubt. What do you think?

yohamta0:
I am a bit worried that it is starting to be complex if we add arguments like `--no-queue`. I think that adding an option to override `queue:` setting for a specific retry would be simpler (e.g., `--queue emergency`). However let's make it a separate, future enhancement rather than adding it to this PR.

ghansham:
> I am a bit worried that it is starting to be complex if we add arguments like `--no-queue`. I think that adding an option to override `queue:` setting for a specific retry would be simpler (e.g., `--queue emergency`). However let's make it a separate, future enhancement rather than adding it to this PR.

Agree with the solution proposed. We need to add an optiona in UI to set the queue while retrying or rescheduling a dagrun in that case.

PR #1695, opened 2026-02-22
feat: access log config

### LINKED ISSUE #1694: Option to skip access logging for /api/v2/health (e.g. for Kubernetes probes)
### Problem:
When Kubernetes liveness/readiness probes hit /api/v2/health every few seconds, the HTTP access logger writes a log line for every request. In clusters with multiple pods this produces a lot of noise and makes it harder to see real traffic and errors.
Example log lines:
level=INFO msg="Response: 200 OK" service: "http" httpRequest: {url: "http://...:8080/api/v2/health" method: "GET" path: "/api/v2/health" ... header: {user-agent: "kube-probe/1.30" ...}} httpResponse: {status: 200 ...}

### Desired behavior:
Either:
Do not log requests to /api/v2/health, or
Make this configurable (e.g. path skip list or “quiet” routes).
Where it comes from
The logging is set up in internal/service/frontend/server.go when creating the request logger and mounting the middleware:
httplog.NewLogger("http", httplog.Options{...}) (around lines 643–652)
QuietDownRoutes is already used for /api/v1/events only.
The middleware is applied via sanitizedRequestLogger(requestLogger) (around line 657).
So the behavior is hardcoded and there is no config or env to turn off health-check logging.

### Suggested fix:
Minimal change: Add /api/v2/health to QuietDownRoutes in the same httplog.Options block so health checks are not logged (or are quieted the same way as /api/v1/events).
More flexible: Use Skip in httplog.Options to skip logging when:
req.URL.Path == "/api/v2/health" (or the configured API base path + /health), and/or
User-Agent indicates a probe (e.g. strings.Contains(req.UserAgent(), "kube-probe")).
Either approach would stop the health-check log spam while keeping access logs for real API traffic.

### Environment
Dagu run in Kubernetes with HTTP GET liveness/readiness probes to /api/v2/health on the server port.
Probes use the default kube-probe/1.x User-Agent.
Thanks for considering this.

yottahmd:
Resolves #1694

## Summary by CodeRabbit

* **New Features**
  * Added configurable HTTP access logging modes: log all endpoints, exclude public endpoints from logging, or disable logging entirely.
  * Access log behavior can be set via configuration file, environment variable, or uses the default (all endpoints logged).

### DIFF
--- a/internal/cmn/config/definition.go
+++ b/internal/cmn/config/definition.go
@@ -13,11 +13,11 @@ type Definition struct {
 	TLS         *TLSDef `mapstructure:"tls"`
 
 	// Core settings
-	Debug        bool   `mapstructure:"debug"`
-	DefaultShell string `mapstructure:"default_shell"`
-	LogFormat    string `mapstructure:"log_format"`  // "json" or "text"
-	AccessLog    *string `mapstructure:"access_log"` // "all" (default), "non-public", or "none"
-	TZ           string `mapstructure:"tz"`
+	Debug        bool    `mapstructure:"debug"`
+	DefaultShell string  `mapstructure:"default_shell"`
+	LogFormat    string  `mapstructure:"log_format"`      // "json" or "text"
+	AccessLog    *string `mapstructure:"access_log_mode"` // "all" (default), "non-public", or "none"
+	TZ           string  `mapstructure:"tz"`
 
 	// Authentication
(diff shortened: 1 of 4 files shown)


ghansham:
If possible set log levels also, e.g. info, debug, error,  warning, etc if set to error or warning, the logging will reduce drastically

yohamta0:
@ghansham Yes, I agree. We can add that later.

PR #1699, opened 2026-02-23
fix: resolve ambiguous HOME folder substitution in Docker

### LINKED ISSUE #1698: Ambigious home folder substitution of default config, etc
The default config path is `~/.config/dagu/config.yaml` but when dagu is executed in docker container the `~` is expanded to `/root`. This is because a `-E` flag is set when using `sudo` to switch users (https://github.com/dagu-org/dagu/blob/main/entrypoint.sh#L53 ). This also leads to all the path values with `~` in the example config file resolved to `/root/...`.

As a result, user must either mount their config to `/root/.config/dagu/config.yaml` instead or explicitly write `/home/dagu` instead of `~` in the config file and specify `--config /home/dagu/.config/dagu/config.yaml`.

This is unclear in the document and doesn't look like a good practice. 

I would suggest two solutions:

* Modify `Dockerfile` and `entrypoint.sh` so that the container is started as user `dagu`. This includes moving the UID / GID manipulations to `Dockerfile`.
* Add `-H` flag to the `sudo` command at the end of `entrypoint.sh`, so that `$HOME` is set to `/home/dagu` and the document is more accurate. 

Personally I would prefer the first way as it follows the principle of least privilege and modern containerization standards like in other Docker applications. However, this might break some users' setup. I am happy to submit a Pull Request to address this if the maintainers agree with the direction.

yottahmd:
## Summary

Fixes #1698.

When running Dagu in Docker, `sudo -E` in `entrypoint.sh` preserves root's `$HOME=/root` when switching to the `dagu` user. This causes `~` to expand to `/root` instead of `/home/dagu`, forcing users to use unintuitive mount paths.

- Add `-H` flag to `sudo` in `entrypoint.sh` so `$HOME` is set to the target user's home directory from `/etc/passwd`, while `-E` continues to preserve all other environment variables
- Fix both Alpine Dockerfiles to create user with home `/home/dagu` (was `/dagu`), aligning with `WORKDIR /home/dagu` and the Ubuntu Dockerfiles

## Files changed

- `entrypoint.sh` — add `-H` flag to sudo invocation
- `Dockerfile.alpine` — change user home from `/dagu` to `/home/dagu`
- `deploy/docker/Dockerfile.alpine` — same home directory fix

## Test plan

Verified on both Alpine and Ubuntu images:

- [x] `docker run --rm <image> sh -c 'echo $HOME'` → `HOME=/home/dagu`
- [x] `docker run --rm <image> sh -c 'echo $DAGU_HOME'` → `DAGU_HOME=/var/lib/dagu` (confirms `-E` still works)
- [x] `docker run --rm -e MY_VAR=test <image> sh -c 'echo $MY_VAR'` → `test` (env passthrough intact)
- [x] `docker run --rm -e PUID=1001 <image> sh -c 'id -u && echo $HOME'` → uid 1001, `HOME=/home/dagu`
- [x] `docker run --rm <image> dagu version` → runs without errors about `/root/.config/dagu/`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Docker container user home directory paths to align with standard Linux conventions
  * Enhanced environment variable handling during container startup to properly configure the HOME directory setting

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,12 +60,12 @@ RUN set -eux; \
     # Try to create group with specified GID, fallback if GID in use
     (groupadd -g "${USER_GID}" "${USER}" || groupadd "${USER}") && \
     # Try to create user with specified UID, fallback if UID in use
-    (useradd -m -d /home/dagu \
+    (useradd -m -d /home/${USER} \
               -u "${USER_UID}" \
               -g "$(getent group "${USER}" | cut -d: -f3)" \
               -s /bin/bash \
               "${USER}" \
-    || useradd -m -d /home/dagu \
+    || useradd -m -d /home/${USER} \
                -g "$(getent group "${USER}" | cut -d: -f3)" \
                -s /bin/bash \
                "${USER}") && \
@@ -88,7 +88,7 @@ RUN mkdir -p "${DAGU_HOME}" && \
     chown -R "${USER_UID}:${USER_GID}" "${DAGU_HOME}" && \
     chmod 755 "${DAGU_HOME}"
 
-WORKDIR /home/dagu
+WORKDIR /home/${USER}
 ENV DAGU_HOST=0.0.0.0
 ENV DAGU_PORT=8080
 ENV DAGU_HOME=${DAGU_HOME}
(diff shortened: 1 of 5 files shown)


yohamta0:
@coderabbitai review

PR #1725, opened 2026-03-05
fix: make update banner dynamic instead of static at startup

yottahmd:
The update check was only evaluated once at server startup and injected as static values into the HTML template. Long-running servers would never show the update banner even if new versions were released.

Now the template functions read from the cache store on every page load, and a periodic background goroutine refreshes the cache every 24 hours.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced update checking to run periodically throughout the application lifecycle, ensuring users have access to the latest version information with improved freshness and accuracy.

### DIFF
--- a/internal/service/frontend/server.go
+++ b/internal/service/frontend/server.go
@@ -93,6 +93,7 @@ type Server struct {
 	dagStore           exec.DAGStore
 	licenseManager     *license.Manager
 	remoteNodeResolver *remotenode.Resolver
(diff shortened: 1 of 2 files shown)


ghansham:
For airgapped systems, should it be made configurable (default  value: true) to switch it off permanently.

yohamta0:
Yes, let's add it in backlog.

PR #1730, opened 2026-03-07
feat: add tags parameter to start and enqueue APIs

yottahmd:
Replace fragile frontend YAML regex manipulation with a first-class tags API parameter. All 5 start/enqueue endpoints now accept an optional tags array, threaded through the CLI subprocess via --tags flag and merged into the DAG struct before execution. The cockpit switches from inline-spec enqueue with YAML manipulation to file-based enqueue with structured tags.

## Release Notes

* **New Features**
  * Added support for tagging DAG runs with custom key-value pairs across all DAG execution endpoints (start, start-sync, enqueue).
  * Tags accept comma-separated key=value or key-only format and merge with DAG spec tags.
  * Extended CLI with --tags parameter for start and enqueue commands.
  * Updated UI to support tags input during DAG execution and enqueueing.

### DIFF
--- a/internal/cmd/start.go
+++ b/internal/cmd/start.go
@@ -470,7 +470,9 @@ func dispatchToCoordinatorAndWait(ctx *Context, d *core.DAG, dagRunID string, co
 	)
 
 	var taskOpts []executor.TaskOption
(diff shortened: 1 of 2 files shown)


ghansham:
This is going to be s big help

PR #1753, opened 2026-03-10
Relax step name limit to 255 characters

### LINKED ISSUE #1189: 40 character limit for step name
Hi! In the post 1.17.0 version the step names have a 40 character limit, meaning that when we migrated we had to change the length of the names. Potentially we should be using the description if its long, but any reason for the change?

SergioChan:
## Summary
- raise `maxStepNameLen` from 40 to 255 so longer descriptive step names are accepted
- update the step-name validation error text to match the new limit
- adjust validator tests to cover the new boundary (`255` passes, `256` fails)

## Testing
- Attempted: `go test ./internal/core -run 'TestValidateSteps|TestErrors'`
- Not runnable in this environment because the repository currently requires Go 1.26 toolchain download, and that toolchain is unavailable here (`go: toolchain not available`).

## Related
Fixes #1189

## Release Notes

* **Enhancements**
  * Step names can now support up to 255 characters, increased from the previous 40-character limit, enabling more descriptive naming conventions.

* **Validation Updates**
  * Updated validation constraints to enforce the new step name length limits.

### DIFF
--- a/internal/core/validator.go
+++ b/internal/core/validator.go
@@ -12,6 +12,7 @@ import (
 // Constants for validation limits.
 const (
 	DAGNameMaxLen  = 40
+	maxStepIDLen   = 40
 	maxStepNameLen = 255
 )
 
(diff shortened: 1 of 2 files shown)


SergioChan:
Thanks for the review — I pushed a follow-up commit addressing the two requested fixes:

- Restored the dedicated 40-character limit for `step.ID` by introducing `maxStepIDLen` and using it in ID validation.
- Updated direct `validateStep` boundary tests to the new step name limit (`255` pass / `256` fail), including the promoted-ID-long case.

Validation note:
- I attempted to run the targeted validator tests, but this environment currently cannot download the required Go 1.26 toolchain, so I couldn’t complete a local test run in this pass.

zph:
As a data point, the change to 40 characters broke some integrations I have with Dagu (eg break in compatibility from older version).

I came here seeing what the original reason was for this change and see from 5hr ago there's a PR to bump it.

Thanks for taking a look and considering the change!

PR #1755, opened 2026-03-11
fix: prevent duplicate catchup DAG runs on DAG file change

yottahmd:
When a catchup item was popped from the buffer in Plan(), the per-DAG watermark was only updated asynchronously in DispatchRun (a separate goroutine). If a DAG file change event arrived between Plan() releasing entryMu and the goroutine updating the watermark, recomputeBuffer() would read stale watermark values and re-create the already-dispatched item — causing a duplicate run.

Fix: call advanceDAGWatermark() inside Plan() immediately after popping a catchup item, while entryMu is held. Since recomputeBuffer() also requires entryMu (via handleEvent/drainEvents), the watermark is always current when it runs. This follows the existing pattern used for skip/latest overlap policies.

Also remove the now-redundant catchup watermark update from DispatchRun() and the catchup skip in Advance(), so all start runs are treated uniformly.

Add a regression test that reproduces the exact interleave: Plan() pops item → handleEvent(Updated) before Advance() → verify recomputed buffer excludes the dispatched item.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate catch-up scheduling and ensured catch-up items are retained for retry if persistence fails.
  * Scheduler now checks the queue for pending runs when no active workers exist to avoid missed overlap detection.

* **Chores**
  * Catch-up runs are persisted before advancing per-DAG watermarks, improving watermark correctness and reliability.

* **Tests**
  * Expanded coverage for enqueue failures, concurrent DAG updates, and retry/recovery paths.

### DIFF
--- a/internal/cmd/enqueue.go
+++ b/internal/cmd/enqueue.go
@@ -77,9 +77,9 @@ func runEnqueue(ctx *Context, args []string) error {
 		return err
 	}
 
-	scheduledTime, err := ctx.StringParam("scheduled-time")
+	scheduledTime, err := parseScheduledTimeParam(ctx, "scheduled-time")
 	if err != nil {
-		return fmt.Errorf("failed to get scheduled-time: %w", err)
+		return err
 	}
 
 	return enqueueDAGRun(ctx, dag, runID, triggerType, scheduledTime)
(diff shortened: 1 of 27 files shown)


yohamta0:
@coderabbitai full review

Issue #546, opened 2024-03-26
high memory usage


helmut72:
I run dagu with a single instance for about 9 days. Still testing dagu with 3 cronjobs that runs every few minutes. RAM usage starts at around 15MB. After 9 days it use 92MB RAM. In the past it even use over 300MB RAM with a longer uptime.

yohamta:
Thanks for reporting the issue. Let's try to fix this.

helmut72:
Thank you. I have started a second instance with example cronjobs. If RAM usage goes up too after some days, I'll share this example.

helmut72:
My test instance starts at about 20 MB RAM and now need 60 MB RAM. In 3 days.

Command:
```
docker run -d --name dagu-test --rm -p 8080:8080 -v ./dags:/home/dagu/.dagu/dags -v ./dagu/data:/home/dagu/.dagu/data -v ./logs:/home/dagu/.dagu/logs yohamta/dagu:latest dagu start-all
```

Using 4 dags from your example:

`./dags/hello1.yaml`:
```
name: hello1 

schedule: "*/9 * * * *"

steps:
  - name: get fake json data
    executor: http
    command: GET https://jsonplaceholder.typicode.com/comments
    script: |
      {
        "timeout": 10,
        "headers": {},
        "query": {
          "postId": "1"
        },
        "body": ""
      }
```

`./dags/hello2.yaml`:
```
name: hello2 

schedule: "*/3 * * * *"

steps:
  - name: get fake json data
    executor: http
    command: GET https://jsonplaceholder.typicode.com/comments
    script: |
      {
        "timeout": 10,
        "headers": {},
        "query": {
          "postId": "1"
        },
        "body": ""
      }
```

`./dags/hello3.yaml`:
```
name: hello3 

schedule: "*/5 * * * *"

steps:
  - name: get fake json data
    executor: http
    command: GET https://jsonplaceholder.typicode.com/comments
    script: |
      {
        "timeout": 10,
        "headers": {},
        "query": {
          "postId": "1"
        },
        "body": ""
      }
```

`./dags/hello4.yaml`:
```
name: hello4 

schedule: "*/7 * * * *"

steps:
  - name: get fake json data
    executor: http
    command: GET https://jsonplaceholder.typicode.com/comments
    script: |
      {
        "timeout": 10,
        "headers": {},
        "query": {
          "postId": "1"
        },
        "body": ""
      }
```

My other instance now runs for 12 days and need 122 MB RAM.

[...]

PR #1769, opened 2026-03-13
scheduler: reduce queue startup lookup churn

yottahmd:
## Summary
- replace root dag-run lookup in `FindByDAGRunID` with a reverse-chronological exact-match scan instead of `filepath.Glob + sort`
- reduce queue startup polling churn by skipping status-store fallback during a short startup grace period and only aborting early on true pre-start execution failures
- add focused regression coverage for duplicate run IDs, exact matching, startup grace behavior, and pre-start failure classification

## Motivation
This change targets the scheduler memory-growth investigation around repeated queue startup polling and repeated filesystem scans of the growing `dag-runs` tree.

Related to #546.

## Risk Areas
- queue startup acknowledgment semantics in the scheduler retry path
- dag-run lookup correctness when duplicate run IDs exist across days or within the same day
- classification of execution failures that should stop startup polling immediately versus failures that should still rely on heartbeat/status observation

## Validation
- `make fmt`
- `go test ./internal/persis/filedagrun`
- `go test ./internal/service/scheduler -run '^TestQueueProcessor_'`
- `go test ./internal/persis/filedagrun ./internal/service/scheduler -run '^(TestDataRootRuns|TestQueueProcessor_CheckStartupStatus_|TestIsPreStartExecutionFailure)'`

## Additional Notes
- I did not rely on the full `go test ./internal/service/scheduler` result because this environment still has existing scheduler test dependencies on a local `dagu` binary and broader port-sharing behavior unrelated to this patch.

## Summary by CodeRabbit

* **Performance**
  * DAG run lookup now uses hierarchical, reverse-chronological directory traversal instead of full-scan search.

* **Improvements**
  * Scheduler startup logic now includes grace period configuration and enhanced error detection.
  * Improved startup failure handling with pre-execution error monitoring.

* **Tests**
  * Added test coverage for startup grace period behavior and DAG run lookup accuracy with duplicate IDs.

### DIFF
--- a/internal/persis/filedagrun/dataroot.go
+++ b/internal/persis/filedagrun/dataroot.go
@@ -93,6 +93,8 @@ func NewDataRootWithPrefix(baseDir, prefix string) DataRoot {
 	}
 }
 
+const dagRunTimestampLen = len("20060102_150405Z")
+
 // FindByDAGRunID locates a dag-run by its ID.
 // It scans the year/month/day hierarchy in reverse chronological order and
 // returns the first exact match, which preserves the "newest run wins" behavior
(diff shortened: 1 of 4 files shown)


yohamta0:
@coderabbitai full review

PR #1782, opened 2026-03-16
fix: graceful shutdown for HTTP server and scheduler

yottahmd:
## Summary
- **HTTP server**: Set `BaseContext` on `http.Server` so all request contexts (`r.Context()`) inherit from the service context. Without this, SSE handlers blocked `httpServer.Shutdown()` for 5s because their contexts were never cancelled on shutdown.
- **Scheduler**: Call `scheduler.Stop()` during shutdown in both `startall` and standalone `scheduler` commands. Previously only context cancellation was used, skipping critical cleanup: dir lock release (caused 30s delay on restart), planner watermark flush, queue processor shutdown, and entry reader file watcher close.

## Test plan
- [x] `TestScheduler_StopReleasesLock` — verifies lock is released after Stop(), second scheduler starts immediately
- [x] `TestScheduler_StopAfterContextCancellation` — simulates startall pattern (ctx cancel then Stop())
- [x] Command tests assert "Scheduler stopped" appears in shutdown output
- [x] All existing scheduler and frontend tests pass
- [ ] Manual: `make run` → Ctrl+C → exits promptly, no "Failed to shutdown server gracefully" error, immediate restart works (no 30s lock wait)

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified scheduler cleanup to execute during initialization and shutdown sequence.
  * Added context propagation to HTTP server initialization.

* **Tests**
  * Introduced tests validating scheduler lock release and cleanup behavior.
  * Updated test expectations to reflect additional shutdown output.

### DIFF
--- a/internal/cmd/scheduler_test.go
+++ b/internal/cmd/scheduler_test.go
@@ -21,7 +21,7 @@ func TestSchedulerCommand(t *testing.T) {
 
 		th.RunCommand(t, cmd.Scheduler(), test.CmdTest{
 			Args:        []string{"scheduler"},
-			ExpectedOut: []string{"Scheduler started"},
+			ExpectedOut: []string{"Scheduler started", "Scheduler stopped"},
 		})
 	})
 	t.Run("StartSchedulerWithConfig", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ghansham:
Request you to see if coordinator service is configured, does that also cleanly gets shutdown. We have observed that when we restart by saying dagu start-all, the coordinator service port binding error is seen.

PR #1786, opened 2026-03-17
fix(api): ensure retry endpoint returns immediately without blocking

### LINKED ISSUE #608: API dag "retry" POST request hangs until job completion
good day to you!

today I was playing with API and discovered that "retry" "action" post request hangs until the dag is not finished. The POST request hangs and if the job runs longer than the default request timeout (e.g. 30 sec) - the request fails.
i'd expect it to restart the dag and return 200 code back and not wait until finished ;)

Is this a bug?

P.S. v1.13.0

regards

mvanhorn:
## Summary
- Fix the retry API endpoint (`POST /dag-runs/{name}/{dagRunId}/retry`) blocking until the DAG run completes instead of returning 200 immediately
- Add `waitForRetryStarted` to the local retry path that briefly polls (up to 3s) for the subprocess to reach Running status before returning, matching the pattern used by the start endpoint
- The wait is best-effort: if the timeout elapses the handler still returns 200 since `runtime.Start` already spawned the subprocess asynchronously

## Test plan
- [ ] Verify `go build ./...` passes
- [ ] Verify `go vet ./internal/service/frontend/api/v1/...` passes
- [ ] Manual test: call the retry endpoint on a long-running DAG and confirm the API returns within seconds

Fixes #608

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Bug Fixes**
  * Retry operations now complete more quickly with asynchronous processing instead of blocking.
  * Improved retry startup confirmation to ensure the operation has begun before responding to requests.

### DIFF
--- a/internal/service/frontend/api/v1/dagruns.go
+++ b/internal/service/frontend/api/v1/dagruns.go
@@ -1715,7 +1715,7 @@ func (a *API) waitForRetryStarted(ctx context.Context, dag *core.DAG, dagRunID s
 			return
 		default:
 			status, _ := a.dagRunMgr.GetCurrentStatus(ctx, dag, dagRunID)
-			if status != nil && status.Status == core.Running {
+			if status != nil && (status.Status == core.Running || status.Status == core.Queued) {
 				return
 			}
 			time.Sleep(pollInterval)


ghansham:
Will retry start the dag or enqueue if queue is full? Will the retry wait for enqueing in that case and then return?

yohamta0:
@ghansham Good catch, retry run will be queued if the queue is configured for the DAG. So I believe this should include `Queued` status as well.

```diff
- if status != nil && status.Status == core.Running {
+ if status != nil && (status.Status == core.Running || status.Status == core.Queued) {
```

mvanhorn:
Applied the fix to include `Queued` status in b546c2b. @ghansham - yes, if a queue is configured, the retry will be enqueued rather than started immediately. The wait loop now returns on either `Running` or `Queued` status.

ghansham:
Thanks a ton.

PR #1799, opened 2026-03-19
fix: harden SSE topics and dev asset versioning

yottahmd:
## Summary
- harden SSE session setup so unsupported initial topics do not fail the whole stream
- add SSE regression coverage for unsupported-topic behavior
- use a bundle-content hash for dev/local asset versioning so stale cached UI bundles do not keep subscribing to removed SSE topics

## Testing
- make fmt
- go test ./internal/service/frontend ./internal/service/frontend/sse -count=1

## Summary by CodeRabbit

* **Bug Fixes**
  * Topic subscription now filters unsupported topic types, returning HTTP 403 with `unsupported_topic` errors while allowing partial subscriptions for supported topics.
  * SSE stream initialization proceeds when requests include unsupported topics, reporting those as `unsupported_topic` without blocking supported subscriptions.

* **Improvements**
  * Frontend asset versioning uses a hash-based suffix for dev builds and respects configured release versions for cache-busting.

* **Tests**
  * Added coverage for unsupported-topic behavior across session creation, mutation, and SSE handling; added asset-version formatting tests.

### DIFF
--- a/internal/service/frontend/sse/multiplex.go
+++ b/internal/service/frontend/sse/multiplex.go
@@ -277,6 +277,8 @@ func (m *Multiplexer) applyMutation(ctx context.Context, session *streamSession,
 		}
 	}
 
+	// TODO: Extract add-topic classification into a helper when we can do a
+	// broader cleanup without mixing behavior changes into this regression fix.
 	mutationErrors := make([]TopicMutationError, 0)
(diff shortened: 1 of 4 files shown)


yohamta0:
@coderabbitai full review

PR #1851, opened 2026-03-24
fix: prevent distributed runs from remaining running after execution stops

### LINKED ISSUE #1849: Shared-nothing distributed runs can remain `running` after execution stops
## Problem

A repro built from #1837 exposed a correctness gap in shared-nothing distributed mode. When scheduled DAGs share a global queue and run on remote workers, some runs can stop executing but remain persisted as `running` indefinitely.

This creates split-brain state between run details and queue state. A DAG run can still show `running` in its details while the queue reports `runningCount=0` for the same queue. In practice, that means queue concurrency can be enforced incorrectly, remote runs may never be repaired, and operators are left with ghost `running` runs that require manual cleanup.

## Expected Behavior

Shared-nothing distributed execution should maintain a single authoritative view of distributed run state that stays correct across worker/coordinator interruptions and is used consistently by scheduling, queue capacity, run status, and recovery.

A distributed DAG assigned to a queue should:
- count against queue concurrency while it is actually in progress
- never remain `running` forever after execution has stopped
- transition deterministically to a terminal or recoverable state if its worker or coordinator disappears
- report the same state in DAG run details and queue APIs/UI

## Example

```yaml
# ~/.config/dagu/config.yaml
queues:
  enabled: true
  config:
    - name: batch
      max_concurrency: 1

---
name: distributed-report
queue: batch
schedule: "*/1 * * * *"
worker_selector:
  pool: general

steps:
  - id: generate
    command: sleep 120
  - id: publish
    command: echo "done"
    depends: generate
```

## Motivation

This blocks shared-nothing HA deployments from being reliable in practice. Teams using distributed workers and global queues need queue limits, run status, and failure recovery to stay consistent without depending on shared storage, single-node coordinators, or manual cleanup of stuck runs.

## Tests

- PR must include integration tests under intg package.

yottahmd:
## Summary
- Add a lease-based liveness mechanism for distributed DAG runs: workers now stamp `LeaseAt` on every status push, enabling the coordinator to detect stale runs even after restarts
- Introduce `detectStaleLeases` in the coordinator's zombie detector to scan persisted active runs and mark any with an expired lease as FAILED
- Update queue concurrency counting (scheduler queue processor and frontend API) to include distributed runs with fresh leases, preventing over-scheduling when workers are executing remotely
- Add per-run mutex in `ReportStatus` to serialize with `markRunFailed` and avoid race conditions during zombie cleanup

## Testing
- `make test TEST_TARGET=./internal/intg/distr/... -count=1` (new zombie recovery integration tests)
- `make test TEST_TARGET=./internal/service/coordinator/... -count=1`
- `make test TEST_TARGET=./internal/service/scheduler/... -count=1`
- `make test TEST_TARGET=./internal/service/frontend/api/v1/... -count=1`
- `make test TEST_TARGET=./internal/core/exec/... -count=1`

Closes #1849

## Summary by CodeRabbit

* **New Features**
  * Lease-based monitoring for remote workers to detect crashes/unresponsiveness and fail affected runs automatically.
  * Queue concurrency accounting now includes distributed running jobs so capacity is tracked accurately across nodes.
  * Queue APIs and running-item views now surface distributed runs alongside local ones for consistent UI/status.

* **Tests**
  * Added extensive distributed integration and unit tests for crash recovery, lease/heartbeat handling, queue consistency, and claim semantics.

### DIFF
--- a/internal/core/exec/runstatus.go
+++ b/internal/core/exec/runstatus.go
@@ -122,6 +122,9 @@ func IsLeaseActive(status *DAGRunStatus, staleThreshold time.Duration) bool {
 	if status == nil || status.LeaseAt == 0 {
 		return false
 	}
+	if staleThreshold <= 0 {
+		staleThreshold = DefaultStaleLeaseThreshold
+	}
 	return time.Since(time.UnixMilli(status.LeaseAt)) < staleThreshold
 }
 
--- a/internal/intg/distr/zombie_recovery_test.go
+++ b/internal/intg/distr/zombie_recovery_test.go
@@ -6,6 +6,7 @@ package distr_test
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"os"
 	osexec "os/exec"
 	"testing"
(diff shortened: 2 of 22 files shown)


yohamta0:
@coderabbitai full review

PR #1854, opened 2026-03-25
fix: prevent scheduler split-brain via fence token and self-fencing

### LINKED ISSUE #1849: Shared-nothing distributed runs can remain `running` after execution stops

yottahmd:
## Summary
- Add a unique fence token (written to an `owner` file inside the lock directory) at lock acquisition time to reliably detect ownership changes
- Heartbeat now verifies the fence token before refreshing the lock; on mismatch, the lock is marked as lost
- Unlock checks the fence token before removing the lock directory, preventing a stale scheduler from deleting a replacement lock held by another process
- `IsLocked()` no longer has the side effect of removing stale lock directories — stale cleanup is now only performed by `TryLock`
- `IsHeldByMe()` uses fence token comparison instead of `os.Stat` for accurate ownership detection
- Scheduler self-fences on heartbeat failure: marks the lock as not held and calls `Stop()`, ensuring it does not continue operating after losing ownership

## Testing
- `make test TEST_TARGET=./internal/cmn/dirlock/...`
- `make test TEST_TARGET=./internal/service/scheduler/...`

Closes #1849

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened lock ownership verification to reliably detect and stop using a lock if ownership is lost.
  * Scheduler now self-stops immediately when it loses lock ownership to avoid conflicting activity.
  * Lock status checks are now observe-only for stale locks; cleanup is handled during lock acquisition.

* **Tests**
  * Added and extended tests covering ownership loss, stale-lock scenarios, scheduler self-fencing, and monitor retry/flush timing.

### DIFF
--- a/internal/service/telegram/bot_test.go
+++ b/internal/service/telegram/bot_test.go
@@ -66,6 +66,16 @@ func (a *fakeTelegramAPI) textCount() int {
 	return a.messageCount
 }
 
+func assertTypingStops(t *testing.T, api *fakeTelegramAPI, msg string) int {
+	t.Helper()
+
+	stoppedAt := api.typingCount()
+	assert.Never(t, func() bool {
+		return api.typingCount() > stoppedAt
+	}, 50*time.Millisecond, 5*time.Millisecond, msg)
+	return stoppedAt
+}
+
 type fakeTelegramAgentService struct {
 	mu                 sync.Mutex
 	nextSessionID      int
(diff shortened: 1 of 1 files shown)


yohamta0:
@coderabbitai review

Issue #1252, opened 2025-09-14
Add Parameters & Ouptut Validation


yottahmd:
The two things that need validation across the entire DAG are:

**Parameters** - Validation for the DAG parameters (support key=value and JSON input)
**Step outputs** - Currently, there’s no feature to define "input" for each step directly, so the validation needs to happen at the "output" stage. Below is a rough idea.

**1) Parameter validation**
Each parameter is interpreted as a field in a JSON schema, and the values are validated accordingly. Fields without defaults are treated as required. Supporting JSON input for parameters will be the next step.

```yaml
params:
  - name: ENVIRONMENT
    type: string
    enum: [dev, staging, prod]
    default: dev
    description: Deployment environment
    required: true
```

**2) Output validation**
Allow user to define optional schema in the output field. If the validation fails, the step should fail. This would be useful to validate LLM output for example.

```yaml
steps:
  - command: python task.py
    output:
      name: OUTPUT
      schema:
        type: object
        properties:
          summary: { type: string }
          confidence: { type: number, minimum: 0.0, maximum: 1.0 }
        required: [summary, confidence]
```

ghansham:
We can have following:
1. **Date Time string format validation**: Validation on format of parameter e.g. date (YYYYMMDD, DDMMMYYYY, etc.).
We can also have validation on string length of params. There is a possibility user can be using another command output as a Param e.g.  ` date `, so that needs to be evaluated after execution of that command.
2. **Number of Parameters validation**: 
There can be a validation on number of params (minimum, maximum). 
3. **File/Directory validation**:
a. And another could be a param to be filename, that step needs to read/write/execute (permission test). 
b. Or param could be a directory name that again needs to be read or written (permission test)
4. **Parameter String Length Validation**
5. **Wildcard/Regex Pattern Validation**:
Linux wildcards in params like APOLLO__???_????
Or may be regex based.

Actually rather than keeping it in the dag yaml, should we keep it in a separate file say dagu_dagparams_rules.yaml in the dagu base path. Reason is we can have multiple dags sharing common executables or same dag using a single executable multiple times at different stages based on requirement. We can have a file which contain the list of executables along with its validation rules in a common file which can be used when we read the dag yaml.

And you can merge this issue with #1182

yohamta0:
> Actually rather than keeping it in the dag yaml, should we keep it in a separate file say dag_dagparams_rules.yaml in the dagu base path

[...]

PR #1887, opened 2026-03-31
feat: add inline JSON Schema validation for DAG params (closes #1182)

### LINKED ISSUE #1182: Add JSON Schema–based type for DAG input parameters
I’d like to add optional JSON Schema–based validation for DAG input parameters. This issue is only about the input schema (no outputs).

Proposed YAML:
```yaml
params:
  type: object
  properties:
    batch_size:
      type: integer
      default: 10
      readonly: true
    start_date:
      type: string
      format: date
    debug:
      type: boolean
      default: false
  required: ["start_date"]
```

Feedback on the schema shape and defaults would be great. Thanks!

mbprabhoo:
## Summary

Closes #1182 — adds support for defining JSON Schema **inline** in the `params` field of a DAG, enabling typed parameter validation with defaults, required-field enforcement, and type coercion — all without an external schema file.

This is part of the broader params/output validation work tracked in #1252.

## What Changed

- **`internal/core/spec/dag_params_inline_schema.go`** — new: detects inline schema format and compiles it into a `dagParamPlan` with defaults applied at build time
- **`internal/core/spec/dag_params.go`** — adds `dagParamKindInlineSchema` routing so inline schemas go through the same resolution path as external schemas
- **`internal/core/spec/dag_params_schema.go`** — minor: generalises an error message to cover both inline and external schema cases
- **`internal/core/spec/dag_params_inline_schema_test.go`** — unit tests for inline schema detection, plan building, and error cases
- **`internal/intg/params_validation_issue1182_test.go`** — 5 end-to-end integration tests covering the full lifecycle

## How to Use

Write the JSON Schema directly under `params` in your DAG YAML:

```yaml
params:
  type: object
  properties:
    batch_size:
      type: integer
      default: 10
    start_date:
      type: string
      format: date
    debug:
      type: boolean
      default: false
  required: ["start_date"]
steps:
  - name: run
    command: echo "batch=$batch_size date=$start_date debug=$debug"
```

- **Defaults** are applied automatically — `$batch_size` is `10` and `$debug` is `false` if not supplied.
- **Required fields** (`start_date`) cause the run to fail before any step executes if omitted.
- **Type mismatches** (e.g. `batch_size=not-an-int`) are rejected at startup.
- **Positional parameters** are not supported with schema-backed params.

The inline format is detected automatically when `params` is a map with a `properties` key. The existing external-schema format (`params.schema: path/to/schema.json`) is unchanged and takes precedence.

## QA

- 1472 unit/spec tests pass, 627 integration tests pass, 0 lint issues
- All 5 issue-1182 integration tests pass end-to-end
- No regressions against existing params or output-validation tests

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added JSON Schema validation for DAG parameters (supporting inline schema definitions)
  * Added JSON Schema validation for step output values
  * Parameters are validated before DAG execution begins
  * Step outputs are validated after execution completes; validation failures prevent dependent steps from running

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,5 +51,3 @@ prd.json
 # Compiled integration test binary
 intg.test
 
-# Agent workspace
-projects/


yohamta0:
LGTM, thank you very much!

PR #2028, opened 2026-04-22
fix: restore DAG status live updates

yohamta0:
## Summary
- preserve request-scoped SSE payloads while keeping browser clients on the existing multiplexed stream
- restore live updates for DAG-run list, DAG definition, and exact DAG-run status views
- recompute aggregate run status after manual step status changes so UI state reflects backend mutations

## Root Cause
The SSE multiplexer shared one topic payload/hash across sessions even when fetchers depended on request context such as auth, workspace, or remote node. Some DAG-run list topics were also moved to pure event-driven invalidation even though their visible summaries can change without a reliable event for every field. On the DAG definition page, the start/enqueue flow discarded the returned `dagRunId`, so the page stayed on the DAG definition topic instead of subscribing to the exact newly-started DAG-run.

## Validation
- `go test ./internal/service/frontend/... -count=1`
- `pnpm typecheck`
- `pnpm exec vitest run src/features/dags/components/__tests__/DAGStatus.test.tsx src/features/dags/components/dag-details/__tests__/DAGDetailsPanel.test.tsx src/features/dags/components/dag-details/__tests__/DAGDetailsSidePanel.test.tsx src/features/dags/components/common/__tests__/DAGActions.test.tsx`

## Summary by CodeRabbit

* **New Features**
  * Track newly started DAG runs and follow a specific run in the UI; expose a callback when a run is started.

* **Improvements**
  * Aggregate DAG-run status now recomputes automatically when individual step statuses change.
  * UI applies status updates optimistically and triggers targeted refreshes for affected views.
  * Real-time updates refined: per-session streaming behaves more reliably and queue views use on-demand invalidation while DAG-run topics remain polling.

* **Tests**
  * Added tests covering status recomputation, streaming/polling behavior, and related UI flows.

### DIFF
--- a/internal/service/frontend/api/v1/dagruns.go
+++ b/internal/service/frontend/api/v1/dagruns.go
@@ -2342,7 +2342,7 @@ func deriveManualDAGRunStatus(nodes []*exec.Node, fallback core.Status) core.Sta
 			hasRejected = true
 		case core.NodeFailed:
 			hasFailed = true
-			if node.Step.ContinueOn.Failure && !node.Step.ContinueOn.MarkSuccess {
+			if node.Step.ContinueOn.Failure {
 				hasContinuableFailure = true
 			} else {
 				hasUncontinuableFailure = true
@@ -2363,7 +2363,7 @@ func deriveManualDAGRunStatus(nodes []*exec.Node, fallback core.Status) core.Sta
 	case hasRunning:
 		return core.Running
 	case hasRetrying:
-		return core.Queued
+		return core.Running
 	case hasWaiting:
 		return core.Waiting
 	case hasRejected:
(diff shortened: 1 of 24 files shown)


yottahmd:
Addressed CodeRabbit feedback in `c12675a4`.

- `NodeRetrying` now derives aggregate `Running`, and `ContinueOn.Failure` is treated as continuable even with `MarkSuccess`.
- The polling-sensitive SSE test now waits long enough for the default 1s interval instead of using a non-existent helper.
- SSE metrics now emit per-fetch duration samples and at most one fetch-error metric per topic tick.
- DAG history status updates now revalidate the owned history query after a successful manual step-status PATCH.
- Added missing license headers, failure-path optimistic-update coverage, exhaustive node-status mapping, and async status-modal submit guarding.
- The DAG details page now preserves/overrides exact run IDs correctly when starting a new run or opening fullscreen.
- SSE fetchers that were using detached request contexts now use the existing bounded read timeout wrapper where applicable.

For the per-session coalescing nitpick: I did not add context-fingerprint coalescing in this PR because there is no safe stable fingerprint for all request-scoped authorization/workspace context today. The per-session fetch behavior is intentional for correctness; the concrete risk was reduced by adding bounded fetch timeouts and restoring metrics semantics without changing stream connection count.

yottahmd:
@coderabbitai review

yottahmd:
Addressed the follow-up CodeRabbit feedback in `b361916a`.

- Manual recompute now returns non-active `PartiallySucceeded` for mixed `NotStarted` + successful nodes, so manual edits do not make completed runs look actively running.
- `GetDAGRunsListData`, `GetDAGsListData`, `GetDocTreeData`, and `GetDocContentData` now run under the same bounded read timeout wrapper used by the other detached-context SSE fetchers.

Validation: `go test ./internal/service/frontend/... -count=1`.

yottahmd:
@coderabbitai review

yottahmd:
Addressed the latest CodeRabbit DAG-list default feedback in `1ece5a7f`.

- `GetDAGsListData` now uses `a.config.UI.DAGs.SortField` / `SortOrder` as defaults, with explicit SSE query params still taking precedence.
- Added regression coverage that compares the HTTP list path and SSE list path under configured `name`/`desc` defaults.

Validation: `go test ./internal/service/frontend/... -count=1`.

yottahmd:
@coderabbitai review

PR #2055, opened 2026-04-29
fix: refresh structured output examples

yohamta0:
## Summary
- refresh the built-in DAG examples to use publish-only object-form `output:` for synthetic values instead of shelling out with `echo`
- add a real JSON parsing example that shows `decode: json` and `select:` as a lightweight runtime contract for step output
- update the bundled Dagu skill references to explain string-form vs object-form output and the failure behavior for malformed decoded output

## Why
- examples should model the current structured output feature directly instead of preserving old shell-based workarounds for fixed values
- the bundled skill is user-facing guidance, so it needs to describe the actual two-mode `output:` behavior and the retry-relevant failure semantics
- a concrete built-in example makes it easier for users to discover that `decode: json` plus `select:` can validate command output structure at runtime

## Validation
- `go test ./internal/cmd -run TestExampleCommand -count=1`

## Summary by CodeRabbit

* **Documentation**
  * Clarified two output modes: string-form for flat variables and object-form for structured, step-scoped outputs (with decode/select); emphasized stdout is a log file path; added examples for JSON/YAML field extraction and validation.
  * Updated example DAGs and routing examples to use structured output references and explicit success/failure routing.
* **Bug Fix**
  * Ignore temporary/atomic-write files when listing persisted JSON files.
* **Tests**
  * Added unit test verifying temp files are excluded from persisted file lists.

### DIFF
--- a/internal/cmd/example.go
+++ b/internal/cmd/example.go
@@ -48,21 +48,24 @@ steps:
 	{
 		ID:          2,
 		Name:        "output-passing",
-		Description: "Capture step output and pass between steps",
+		Description: "Use publish-only object-form output between steps",
 		Content: `type: graph
 steps:
-  - name: get-version
-    command: echo "2.5.0"
-    output: VERSION
-  - name: get-metadata
-    command: echo '{"build":"abc123","env":"staging"}'
+  - id: publish_version
     output:
-      name: BUILD_ID
-      key: build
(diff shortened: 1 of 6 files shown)


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

PR #2076, opened 2026-04-30
fix: align command condition expansion with shell commands

### LINKED ISSUE #2057: Inconsistent behaviour between commands and preconditions
Alright another one that wasted me too much time: the way dagu handles variable expansion in commands -or other parts- differ from the preconditions.

I couldn’t find any mention of this difference in the documentation, and I’m not sure whether it’s intentional.

# Example DAG
This is a simple example dag that creates a file dagu-test.txt in the users home and echo 1 if the created file exists.

```yaml
env:
  TEST_FILE: ~/dagu-test.txt

steps:
  - command: touch $TEST_FILE
  - command: echo 1
    preconditions:
      - condition: test -f $TEST_FILE
```

## Expected behaviour
The file dagu-test.txt is created in the users home and the second step echoes a 1.

## Actual behaviour
The file is created but the second step **is skipped** with failed preconditon.

# Reason
I've managed to track the issue to the way dagu is expanding the variable -or lack thereof- in the precondition.

The expansion of the variable is being handled differently in the precondition and the **~** is not being expanded. Instead the shell is looking for a literal ~. Replacing ~ with a path or variable works. eg `TEST_FILE: $HOME/dagu-test.txt`.

yohamta0:
## Summary

- align command-form condition evaluation with shell `command` expansion semantics
- share runtime command eval options between the command executor and condition execution paths
- add regression coverage for home-relative DAG env vars in preconditions and repeat-policy conditions

## Root Cause

Command-form conditions were evaluated with `eval.OnlyReplaceVars()`, which deferred scoped variables to the shell. That diverged from shell `command:` steps, where Dagu expands the command text before execution. Values like `~/...` therefore worked in shell commands but failed in conditions because tilde expansion does not happen after shell parameter expansion.

## Testing

- `go test ./internal/runtime -run "TestEvalConditions|TestRunner_DAGPreconditions|TestRunner_RepeatPolicy" -count=1`
- `go test ./internal/intg -run "TestPreconditionWithDAGEnvVars|TestDAGLevelPreconditionWithEnvVars|TestPreconditionWithHomeRelativeDAGEnvVar|TestRepeatPolicy" -count=1`
- `go test ./internal/runtime/builtin/command -run TestCommandExecutor_GetEvalOptions -count=1`

Closes #2057

## Summary by CodeRabbit

* **Documentation**
  * Updated OpenAPI schema documentation for condition execution semantics, clarifying behavior when the `expected` parameter is omitted versus specified.

* **Tests**
  * Added integration tests for condition evaluation with environment variable expansion and repeat policy.
  * Enhanced unit tests with improved capability registration lifecycle management.

### DIFF
--- a/api/v1/api.yaml
+++ b/api/v1/api.yaml
@@ -9518,10 +9518,10 @@ components:
       properties:
         condition:
           type: string
-          description: "Expression or check to evaluate"
+          description: "Expression or check to evaluate. When `expected` is omitted, the value runs as a command check using the same variable expansion rules as shell `command` steps."
         expected:
           type: string
-          description: "Expected result of the condition evaluation"
+          description: "Expected result of the condition evaluation. When set, Dagu compares the evaluated string output instead of using command exit status."
         negate:
           type: boolean
           description: "If true, inverts the condition result (run when condition does NOT match)"
--- a/internal/core/capabilities.go
+++ b/internal/core/capabilities.go
@@ -30,8 +30,18 @@ type ExecutorCapabilities struct {
 	LLM bool
 	// Agent indicates whether the executor supports the agent field.
 	Agent bool
-	// GetEvalOptions returns eval options for command argument evaluation.
-	// If nil, default evaluation is used.
+	// GetCommandEvalOptions returns eval options for command field evaluation.
+	// Step command fields disable backtick substitution by default; this hook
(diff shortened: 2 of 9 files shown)


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

PR #2086, opened 2026-05-04
fix: sort suspended DAGs after live next runs

yohamta0:
## Summary
- Treat suspended DAG definitions as having no runnable next run when sorting by `nextRun`
- Add a regression test covering a suspended earlier schedule versus a live later schedule

## Root cause
Suspended DAGs still received projected next-run timestamps during server-side DAG listing, so the workflows page could place non-live DAGs ahead of live scheduled DAGs when sorting by next run.

## Tests
- `go test ./internal/persis/filedag -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Next-run sorting now correctly places suspended DAGs at the end of listings instead of treating them as active with projected run times.

* **Tests**
  * Added a test that verifies suspended DAGs are ordered after active DAGs when sorting by next-run time.


yohamta0:
@coderabbitai review

PR #2087, opened 2026-05-05
fix: preserve dag-run artifact paths in index

yohamta0:
## Summary
- preserve DAG-run artifact archive paths in the day index used by paginated run listings
- restore `ArchiveDir` when list responses are built from indexed summaries
- add regression coverage for index-backed list results so artifact availability remains visible in Cockpit

## Root cause
The DAG-run day index omitted `ArchiveDir`, so once enough runs existed for `.dagrun.index` to be used, `/dag-runs` list responses reconstructed statuses without the artifact directory. Detail and artifact endpoints still read `status.jsonl`, but Cockpit cards use the list response and saw `artifactsAvailable: false`.

## Tests
- `go test -count=1 ./internal/persis/filedagrun/... ./internal/service/frontend/api/v1`

## Summary by CodeRabbit

* **Enhancements**
  * DAG runs now track and preserve archive directory information throughout persistence and retrieval operations.

* **Tests**
  * Added test coverage ensuring archive directory is preserved when listing DAG runs with history.

### DIFF
--- a/internal/persis/filedagrun/dagrun.go
+++ b/internal/persis/filedagrun/dagrun.go
@@ -76,6 +76,7 @@ type DAGRunSummary struct {
 	AutoRetryMaxInterval time.Duration
 	ProcGroup            string
 	SuspendFlagName      string
+	ArchiveDir           string
 }
 
 // DAGRun represents a dag-run with its associated timestamp and run ID.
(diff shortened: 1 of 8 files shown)


yohamta0:
@coderabbitai review

PR #2127, opened 2026-05-08
fix: keep history JSON output valid for empty results

### LINKED ISSUE #2092: `dagu history` should output its error messages to `STDERR` instead of `STDOUT`, especially when using `--format json`
At the moment, when getting an empty result from `dagu history`, the error message is written to stdout, even when using `--format json`.
This becomes a problem when trying to pipe the output into some other JSON-parsing tool. 

```
sh-5.3# dagu history --run-id FOO --format json 2>/dev/null
No DAG runs found matching the specified filters. # Message written to STDOUT
```

I would suggest that in this situation, the output should either be empty or an appropriate message should be written to `STDERR` and an empty JSON-Object to `STDOUT`:

```
sh-5.3# dagu history --run-id FOO --format json
[] # Empty JSON array on STDOUT
No DAG runs found matching the specified filters. # On STDERR
```

yohamta0:
## Summary
- route empty history result notices to stderr instead of stdout
- render structured empty results for json and csv output paths
- avoid mutating shared agent web tool provider config while applying defaults, fixing a CI race detected in the Ubuntu shard
- add regression tests for empty json history output and non-mutating web tool config resolution

## Testing
- git diff --check
- go test ./internal/intg -run TestHistoryCommand -count=1
- go test ./internal/cmd -count=1
- go test ./internal/agent -run TestResolveWebToolsConfig_DoesNotMutateInput -count=1
- go test -race ./internal/agent -run 'TestRegisteredTools_FactoriesProduceValidTools|TestResolveWebToolsConfig' -count=1
- go test -race ./internal/agent -count=1

Closes #2092

## Summary by CodeRabbit

* **New Features**
  * History command now produces format-aware empty results (JSON/CSV render as empty outputs while other formats show a message).

* **Bug Fixes**
  * Web tools configuration processing no longer mutates the caller's input.

* **Tests**
  * Added tests covering empty JSON history output and ensuring web tools config normalization does not mutate inputs.

### DIFF
--- a/internal/agent/web_tools.go
+++ b/internal/agent/web_tools.go
@@ -336,6 +336,9 @@ func newConfiguredWebToolProvider(cfg WebToolsConfig) (webToolProvider, bool) {
 
 // ResolveWebToolsConfig applies runtime defaults and trimming for web tool config.
 func ResolveWebToolsConfig(cfg WebToolsConfig) WebToolsConfig {
+	if cloned := cloneWebToolsConfig(&cfg); cloned != nil {
+		cfg = *cloned
+	}
 	if cfg.Backend == "" {
 		cfg.Backend = WebToolsBackendTavily
 	}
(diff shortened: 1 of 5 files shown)


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

PR #2187, opened 2026-05-21
fix: recompute env variables before launching scheduled DAGs

spicyneuron:
## Summary

Fixes a bug where scheduled DAGs skip env evaluation on launch, causing backticks to output as literal strings.

```yaml
env:
  - TODAY: "`date +%Y-%m-%d`"
  - DATA: "data/${TODAY}"
```

Directly invoking `dagu start ...` works as expected - `${DATA}` evaluates to the date stamp.

But scheduled runs incorrectly evaluate `${DATA}` to the string: `` data/`date +%Y-%m-%d` ``.

## Changes

- Add `RecomputeEnv` flag to `ResolveEnvOptions` so callers can opt for re-evaluation.
- Scheduler passes `RecomputeEnv: true` when preparing a DAG for subprocess launch, ensuring env is fresh at execution time.
- Deliberately does not evaluate during metadata scan.
- Add test covering backtick evaluation plus chained env variable resolution.

## Related Issues

n/a

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Environment variables in DAG execution are now properly re-evaluated during subprocess preparation, ensuring dynamic values and expressions are correctly resolved instead of potentially using cached values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dagucloud/dagu/pull/2187?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/core/spec/runtime_env.go
+++ b/internal/core/spec/runtime_env.go
@@ -14,7 +14,8 @@ import (
 // ResolveEnvOptions controls how a DAG is reloaded to recover resolved env
 // values for subprocess launchers.
 type ResolveEnvOptions struct {
-	BaseConfig string
+	BaseConfig   string
+	RecomputeEnv bool
 }
 
 // QuoteRuntimeParams quotes persisted params so values containing spaces survive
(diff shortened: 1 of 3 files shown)


yohamta0:
Thanks for the report and for putting together a fix.

I’m going to close this PR because I don’t think this is a bug in Dagu’s current behavior. Dynamic evaluation is intentionally scoped, and metadata/scheduler paths must avoid evaluating arbitrary backtick expressions while loading DAG definitions.

The relevant docs are here:

- https://docs.dagu.sh/writing-workflows/template-variables#command-substitution
- https://docs.dagu.sh/writing-workflows/environment-variables#command-substitution
- https://docs.dagu.sh/writing-workflows/parameters#dynamic-defaults-with-eval

For DAG params specifically, dynamic values should use inline `eval:`. Literal defaults and runtime overrides remain literal by design.

Thanks again for the contribution. I’m closing this as intended behavior rather than merging a scheduler-side recomputation change.

spicyneuron:
Maybe I'm misreading it, but https://docs.dagu.sh/writing-workflows/template-variables#command-substitution clearly states that `env` _does_ support dynamic evaluation?

> Execute commands and use their output in env: blocks using backticks. This runs the command at DAG load time and stores the result:

```
env:
  - TODAY: "`date +%Y-%m-%d`"
  - HOSTNAME: "`hostname -f`"
  - GIT_COMMIT: "`git rev-parse HEAD`"

steps:
  - run: echo "Deploy on ${TODAY} from ${HOSTNAME}"
```

As of now, this is only true if run via `dagu start`, but not during a scheduled run. Is that inconsistency deliberate?

It creates a rather frustrating debugging loop where scheduled runs always fail, but manual debugging runs always succeed. After triple checking my own logic, it took finally inspecting dagu source to understand the issue — so even if not actually a bug, I think situationally dynamic `env` is a pretty confusing convention.

PR #2190, opened 2026-05-21
fix(mail): forward cfg.Attachments to the mailer instead of an empty slice

bitsmashtrader:
## Summary

The `mail` step decodes `attachments:` from YAML into `cfg.Attachments` but
then hands a hardcoded `[]string{}` to `mailer.Send`, silently dropping every
attachment. The schema, docs, and config field all suggest the feature works;
only the one call site that hands the slice to the mailer ignores it.

## Changes

- `internal/runtime/builtin/mail/mail.go`: pass `e.cfg.Attachments` instead of `[]string{}` to `mailer.Send`.
- `internal/runtime/builtin/mail/mail_test.go`: add `TestMailRunSendsAttachments`, an end-to-end regression test that drives the executor against an in-process fake SMTP server and asserts the captured DATA payload contains `Content-Disposition: attachment; filename=<basename>` and the base64-encoded file body.

The existing `TestMail` only asserts that `cfg.Attachments` decodes correctly from YAML; it never invokes `Run()`, which is why this slipped through. The new test fails against the previous code with a clear diagnostic and passes with the fix.

## Repro (before the fix)

```yaml
smtp:
  host: smtp.example.com
  port: "587"

steps:
  - name: send
    executor:
      type: mail
      config:
        from: a@example.com
        to: b@example.com
        subject: hello
        message: see attached
        attachments:
          - /tmp/report.xlsx
```

This step succeeds and logs "sending email succeed.", but `report.xlsx` is never on the outgoing message.

## Related Issues

(none filed — happy to open one first if maintainers prefer)

## Checklist

- [ ] Documentation has been updated as needed (feature was already documented as working)
- [x] Changes have been tested locally (go test ./internal/runtime/builtin/mail/...)

## Summary by CodeRabbit

* **Bug Fixes**
  * Email attachments configured in mail steps are now correctly included and transmitted with outgoing messages.

* **Tests**
  * Added an end-to-end regression test that verifies attachments are sent and their encoded content and filenames appear in outgoing emails.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dagucloud/dagu/pull/2190?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/runtime/builtin/mail/mail.go
+++ b/internal/runtime/builtin/mail/mail.go
@@ -105,7 +105,7 @@ func (e *mail) Run(ctx context.Context) error {
 		toAddresses,
 		e.cfg.Subject,
 		e.cfg.Message,
-		[]string{},
+		e.cfg.Attachments,
 	)
 	if err != nil {
 		_, _ = e.stderr.Write([]byte("error occurred."))
--- a/internal/runtime/builtin/mail/mail_test.go
+++ b/internal/runtime/builtin/mail/mail_test.go
@@ -208,17 +208,22 @@ func TestMailRunSendsAttachments(t *testing.T) {
 	require.NoError(t, err)
 
 	dataCh := make(chan string, 1)
-	serverDone := make(chan error, 1)
+	// serverErr signals fake-SMTP failures only; a clean run leaves it empty
+	// so the select below can use it as an unambiguous fail-fast case without
+	// racing the happy-path completion.
+	serverErr := make(chan error, 1)
 
 	go func() {
 		conn, err := listener.Accept()
 		if err != nil {
-			serverDone <- err
+			serverErr <- err
 			return
 		}
 		defer func() { _ = conn.Close() }()
 		_ = conn.SetDeadline(time.Now().Add(10 * time.Second))
-		serverDone <- runFakeSMTP(conn, dataCh)
+		if err := runFakeSMTP(conn, dataCh); err != nil {
+			serverErr <- err
+		}
 	}()
 
 	step := core.Step{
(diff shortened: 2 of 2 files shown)


bitsmashtrader:
@coderabbitai review

bitsmashtrader:
@coderabbitai review

Re: the nitpick on the wait loop — applied with one tweak. Your patch as written introduced a race: in the happy path the goroutine sends `nil` to `serverDone` after the QUIT exchange, so by the time the select fires, both `dataCh` and `serverDone` have values and Go picks pseudo-randomly. The unconditional `t.Fatal` after `require.NoError` then fails the test ~half the runs (caught it on the first re-run locally).

Fixed by changing `serverDone` (renamed `serverErr`) to signal errors only — a clean run leaves it empty, so the select branch is unambiguously a failure case. Verified with `go test -count=20` locally.

bitsmashtrader:
@yohamta — apologies for the ping; this is a small but user-visible fix and I wanted to flag it in case it slips by.

The `mail.send` step decodes `attachments:` from YAML into `cfg.Attachments` but then passes a hardcoded `[]string{}` to `mailer.Send`, so attachments are silently dropped on every send (the step logs "sending email succeed.", the recipient just gets the body with no file). The existing `TestMail` only asserts the YAML decodes onto `cfg.Attachments`; it never calls `Run()`, which is how this slipped through.

The PR is a one-line fix in `mail.go` plus an end-to-end regression test that drives the executor against an in-process fake SMTP server and asserts the captured DATA payload contains `Content-Disposition: attachment; filename=<basename>` and the base64-encoded file body. CodeRabbit's pre-merge checks are green and its one nitpick has been addressed.

Happy to make any changes you'd like.

yohamta0 (approved):
LGTM, thank you very much for fixing the bug!

PR #2219, opened 2026-05-28
refactor(persis/store): backend-neutral distributed stores — replace file lock with CompareAndSwap + Create

yohamta0:
## Summary

Close the first of the two **neutrality gaps** flagged in `refactor_persis_layer.html`: the distributed control-plane stores (`DAGRunLeaseStore`, `ActiveDistributedRunStore`, `DispatchTaskStore`) previously depended on a runtime type-assert into `WithLockOptions`, which only the file backend satisfies. SQL/etcd backends would have failed at runtime.

This PR moves all three stores to **pure optimistic concurrency** on the `persis.Collection` interface:

- Adds one new primitive: `Collection.Create(ctx, *Record) error`.
- Migrates lease/active to a `Get → Create or CompareAndSwap` retry shape via a new `retryCAS` helper.
- Drops the coarse `locks/dispatch_tasks` lock from dispatch — `CompareAndDelete(pending)` already provided per-task atomicity; count/has methods become eventually consistent.
- `internal/persis/store/` no longer imports `internal/cmn/dirlock`.

**Outcome:** doc's gap count goes 2 → 1 (only backend-specific `DAGRunStore` remains). SQL backend is unblocked for the distributed-store path.

## Commits

| Commit | What |
|---|---|
| `a8020e6f0` | Add `Create` primitive to `persis.Collection`. File backend uses `O_EXCL\|O_CREATE\|O_WRONLY`; memory backend is a map-existence check. Three contract subtests + 16-goroutine atomicity race test. |
| `fda7a3115` | New `retryCAS` helper: exponential full-jitter backoff 5ms→5s cap, matching the existing distributed-lock latency curve. Only `ErrConflict` retries; everything else propagates. |
| `00b01758d` | Migrate `DAGRunLeaseStore` (`Upsert`/`Touch`/`Delete`) and `ActiveDistributedRunStore` (`Upsert`/`Delete`). Caller-visible semantics preserved. Strengthen the existing concurrent-Touch test with distinct timestamps. Add concurrent-Touch+Upsert clobber test and concurrent-active-Upsert test. Remove the three obsolete file-lock-path tests. |
| `a5e03916c` | Drop coarse `withDispatchLock` from `DispatchTaskStore`'s three locked methods. Rename `*Locked` helpers. Collapse `distributed_key.go` to sha256 helper only — `dirlock` import gone from the package. |
| `8267da384` | Wiring: lease/active drop `NewCollectionWithLockRoot` for plain `NewCollection` (no more lock files written by these stores). Dispatch wiring unchanged. |
| `22709f9f2` | Doc: mark Phase 2 step 4 done; gap count 2→1; method count 8→9; mermaid diagram step D flips to green. |

### `Collection.Create`

`CompareAndSwap` returns `ErrNotFound` when the record is missing — it cannot atomically create. Distributed Upsert needs "create if absent, replace if present" semantics. `Create` returns `ErrConflict` if the ID exists. Maps cleanly to SQL `INSERT ... ON CONFLICT DO NOTHING`, etcd `Txn(version=0)`, and file `O_EXCL`.

`WriteFileAtomic` was unsuitable for the file impl — its rename step is unconditional and would silently replace. `O_EXCL|O_CREATE|O_WRONLY` is the POSIX primitive that atomically fails on a pre-existing target, mapped by Go's runtime to `fs.ErrExist` on both Unix and Windows.

### `retryCAS`

Backend-neutral, package-internal. Total time is bounded by `ctx` (not an attempt cap) so callers control the deadline — matches today's `WithLockOptions` ctx-bound wait. Full-jitter spread is the standard mitigation against thundering-herd from heartbeat-aligned writers.

### Dispatch is different

`ClaimNext` already uses `CompareAndDelete(pending)` as the per-task atomicity point — the coarse lock was redundant for correctness. The inner loop's "Put claim → CompareAndDelete pending; on conflict, clean up claim and continue" handles racing pollers without any outer serialization. `CountOutstandingByQueue` and `HasOutstandingAttempt` become eventually consistent: a task in flight between `Put(claim)` and `CompareAndDelete(pending)` may be transiently counted twice. Acceptable for observability — callers (`handler_test.go` invocations included) read between dispatch operations, not during.

### Touch race "bug" doesn't exist

The exploration phase flagged a potential race in `Touch` where `s.Get` looked like it was outside the lock. After verifying the code, the `Get` is inside the `withLeaseLock` closure (invoked while the lock is held). No bug. Migration is purely about backend neutrality, not bug fixing.

## On-disk impact

- No format change. Lease, active-run, dispatch records keep their existing JSON layout.
- Lock directories (`<distributedDir>/locks/...`) are no longer created. Stale lock dirs from old releases become orphans — empty, harmless, no migration needed.
- `dirlock.Unlock` already removed lock files on shutdown; only the empty directory remains.

## Test coverage

- File + memory `Create` contract: 3 subtests (insert, conflict-when-present, after-delete) run via shared `RunCollectionContract`.
- `TestFileCollectionCreateIsAtomicAcrossGoroutines`: 16 goroutines race on the same ID under `-race`; exactly one wins, 15 see `ErrConflict`.
- `TestDAGRunLeaseStore_ConcurrentTouchPreservesLatestHeartbeat`: strengthened to use distinct timestamps so the assertion catches silent fold-to-one.
- `TestDAGRunLeaseStore_ConcurrentTouchAndUpsertNoClobber`: new — documents the CAS-retry semantic when Touch and Upsert race.
- `TestActiveDistributedRunStore_ConcurrentUpsertSerializes`: new — previously zero concurrent-Upsert coverage on the active store.
- All `make test` packages pass with `-race`. `make lint` clean.

## Out of scope

- SQL/etcd backend implementations.
- `proc_lock.go`, `dagstate.go`, `queue.go` `WithLock` callsites — they have legitimate non-CAS lock needs (kill-by-group, single-record locks, queue mutations) and will be ported separately.
- `file.NewCollectionWithLockRoot` constructor stays exported in case other callers need lock-root scoping.

---
## Summary by cubic
Switch distributed stores to backend-neutral optimistic concurrency (`CompareAndSwap` + new `Create`), remove file-lock dependency, and simplify `persis` to JSON-only. Dispatch keeps throughput with an in-process mutex; SQL/etcd backends are unblocked.

- New Features
  - Added `persis.Collection.Create(ctx, *Record) error` for atomic inserts (returns `ErrConflict`); file uses `O_EXCL`, memory mirrors; contract + atomicity tests added.
  - Added `retryCAS` with full-jitter backoff (5ms→5s); retries only on `ErrConflict`.

- Refactors
  - `DAGRunLeaseStore` and `ActiveDistributedRunStore` now use Get → Create|`CompareAndSwap` with `retryCAS`; `Touch` returns `ErrDAGRunLeaseNotFound` when missing; deletes are idempotent.
  - `DispatchTaskStore`: removed file-based `locks/dispatch_tasks` lock; restored an in-process mutex to serialize recycle/scan/claim; `ClaimNext` relies on `CompareAndDelete`; count/has are eventually consistent.
  - Simplified `persis`: dropped `Record.Encoding` and `ExpiresAt`; `Encode` now returns JSON bytes; removed `Collection.Claim` and `WithLockOptions`; deleted `file.NewCollectionWithLockRoot`; `distributed_key.go` now only provides the SHA-256 record key; wiring switches to `file.NewCollection`. Docs/tests updated; no on-disk format changes.
  - CI: pin Go to `1.26.3` in `.github/workflows/ci.yaml`.

<sup>Written for commit 27ab7348c39973b8fc7c5c726954a4373c782226. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2219?utm_source=github">Review in cubic</a>

## Summary by CodeRabbit

* **New Features**
  * Added atomic record creation mechanism to prevent concurrent insertion conflicts.

* **Improvements**
  * Enhanced concurrency handling in distributed lease and active-run stores for improved performance under concurrent access.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dagucloud/dagu/pull/2219?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,7 +46,7 @@ permissions:
   contents: read
 
 env:
-  GO_VERSION: "1.26"
+  GO_VERSION: "1.26.3"
   NODE_VERSION: "20"
 
 jobs:
--- a/internal/cmd/context.go
+++ b/internal/cmd/context.go
@@ -329,10 +329,11 @@ func NewContext(cmd *cobra.Command, flags []commandLineFlag) (*Context, error) {
 	}
 	drs := file.NewDAGRunStore(cfg, hrOpts...)
 	distributedDir := filepath.Join(cfg.Paths.DataDir, "distributed")
-	// Records live in store-specific subdirectories, but locks stay under the
-	// shared distributed root to preserve mixed-version coordinator exclusion.
-	leaseCollection := file.NewCollectionWithLockRoot(filepath.Join(distributedDir, "leases"), distributedDir)
-	activeRunCollection := file.NewCollectionWithLockRoot(filepath.Join(distributedDir, "active-runs"), distributedDir)
+	// Lease and active-run stores use CompareAndSwap-based optimistic
+	// concurrency, so plain collections suffice — the previous lockRoot
+	// scoping for file flock is no longer needed.
+	leaseCollection := file.NewCollection(filepath.Join(distributedDir, "leases"))
+	activeRunCollection := file.NewCollection(filepath.Join(distributedDir, "active-runs"))
 	dagRunLeaseStore := store.NewDAGRunLeaseStore(leaseCollection)
 	activeDistributedRunStore := store.NewActiveDistributedRunStore(activeRunCollection)
 	drm := runtime.NewManager(drs, ps, cfg)
--- a/internal/persis/backend.go
+++ b/internal/persis/backend.go
@@ -16,14 +16,6 @@ import (
 	"time"
 )
 
-// Encoding identifies the serialization format of [Record.Data].
-type Encoding string
-
-const (
-	EncodingJSON  Encoding = "json"
-	EncodingProto Encoding = "proto3"
-)
-
 // Record is the universal storage primitive for all control-plane data.
 //
 // ID uses "/" as a hierarchy separator so that a [ListQuery.Prefix] of
@@ -39,17 +31,8 @@ const (
 type Record struct {
 	ID        string
 	Data      []byte
-	Encoding  Encoding
 	CreatedAt time.Time
 	UpdatedAt time.Time
-	// ExpiresAt is an optional, non-authoritative GC/TTL hint: when non-nil it
-	// tells the backend the record MAY be purged after this time. Correctness
-	// must not depend on it — adapters that need real expiry (heartbeats,
-	// leases) keep the authoritative timestamp inside Data. It is also NOT part
-	// of record identity for CompareAndSwap / CompareAndDelete. The file
-	// backend does not persist it; a SQL backend MAY honor it via an
-	// expires_at column and index.
-	ExpiresAt *time.Time
 }
 
 // ListQuery controls what [Collection.List] returns.
@@ -110,11 +93,6 @@ type Collection interface {
 	// bytes equal expected. Returns [ErrConflict] when they do not match.
 	// Used for optimistic concurrency on DAGRunStatus updates.
 	CompareAndSwap(ctx context.Context, id string, expected, next []byte) error
-
-	// Claim atomically removes one record matching q and returns it.
-	// Returns [ErrNotFound] when no matching record exists.
-	// Used exclusively by queue adapters to implement atomic dequeue.
-	Claim(ctx context.Context, q ListQuery) (*Record, error)
 }
 
 // Backend is the factory for storage [Collection]s and the sole interface
--- a/internal/persis/backend.go
+++ b/internal/persis/backend.go
@@ -90,6 +90,11 @@ type Collection interface {
 	// Put creates or replaces a record.
 	Put(ctx context.Context, rec *Record) error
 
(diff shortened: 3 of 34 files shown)


yohamta0:
@coderabbitai Both findings applied in `4107cfdda`:

1. **`cas_retry.go` — `ctx.Err()` pre-call guard**: added exactly as suggested. The op callbacks (Get/Create/CompareAndSwap) do honor ctx via the underlying Collection methods, but a pre-call guard avoids relying on that and skips one round-trip when the deadline already expired.

2. **`refactor_persis_layer.html` — interface snippet**: the embedded Collection block was missing both `Create` (just added) AND `CompareAndDelete` (pre-existing gap from earlier in the doc's history). Added both. Updated the count text to "Collection: 8 methods, Backend: 2 methods" — more precise than a single conflated total. Status panel's "interface surface" stat bumped to 10 (was 9 in the prior commit — undercounted because I forgot CompareAndDelete was also missing from the snippet).

Verified locally: `go test ./internal/persis/store/... -race` and `make lint` pass.

PR #2262, opened 2026-06-04
fix: pre-filter retry scanner DAG runs at index level (#2261)

### LINKED ISSUE #2261: [FIX] Memory leak: RetryScanner loads all DAG runs into memory before filtering by status
## Root Cause

The file-based DAG run store's `loadDayRuns` allocates `*DAGRun` objects for **all** runs in the retry window before `resolveStatus` filters by status. The day index already contains the status — but it's not used for pre-filtering.

## Impact (user report with 9 hello-world DAGs every minute)

- **9 DAGs × every minute** = 540 runs/hour = **12,960 runs/day**
- RetryScanner (24h window, every 30s): **~12,960 `DAGRun` + 12,960 `DAGRunStatus` objects ≈ 12 MB garbage per scan**
- Over 10 hours: 1,200 scans → **14.4 GB total allocated**
- Go GC collects objects but `MADV_FREE` never returns address space to OS
- Heap fragments over thousands of scan cycles → RSS grows linearly from 15 MB to 250 MB+ over 36h
- **Deleting `data/dag-runs/*` does NOT free memory** — confirms the memory is in Go's retained/fragmented heap, not live objects
- Bug affects all versions (confirmed on v2.3.0, v2.3.1, v2.6.10)
- Also causes sustained 10-30% CPU during "idle" periods from the scan loop

### Fix 1: Pre-filter at index level (stops the bleeding)
Pass the status filter to `loadDayRuns`. When the day index is present and a status filter is active, check `indexEntry.Status` **before** calling `NewDAGRun`. Skip entries that don't match — avoiding 12,960 object creations when 0 runs are failed.

**Files:** `internal/persis/file/dagrun/pagination.go` (`loadDayRuns`, `loadDay`, `listStatusesOrdered`)
**Expected:** ~99.9% reduction in per-scan allocations for installations where most runs succeed

### Fix 2: Periodic `debug.FreeOSMemory()` (cleans the wound)
Add a periodic `debug.FreeOSMemory()` call in the scheduler main loop (~every 5 minutes) to force Go to return retained heap pages to the OS.

**Expected:** Resolves existing bloat and prevents future accumulation

## Acceptance Criteria
1. RetryScanner no longer creates DAGRun objects for runs that don't match the status filter when index is present
2. All existing tests pass
3. No behavior change for unfiltered listing
4. Scheduler RSS stabilizes (no linear growth over days)

Related: #546

four-bytes-robby:
## Summary

Single fix for the memory leak reported in #546, root-caused in #2261.

The file-based DAG run store's `loadDayRuns` function creates `DAGRun` objects for ALL runs in a day before `resolveStatus` filters them by status. The day index already contains the status — use it for pre-filtering.

## Changes

**`internal/persis/file/dagrun/pagination.go`** — Pass the status filter to `loadDayRuns`. When the day index is present and a status filter is active, check each index entry's `Status` before calling `NewDAGRun`. Skip entries whose known status doesn't match the filter.

This avoids creating 12,960+ `DAGRun` + `DAGRunStatus` objects per retry-scanner scan for installations with many historical runs (9 DAGs × every minute = 12,960 runs/day in a 24h window).

## Design Decision: No periodic `FreeOSMemory()`

We considered adding a periodic `debug.FreeOSMemory()` call to the scheduler's cronLoop to return accumulated heap to the OS. **Rejected** — two reasons:

1. **The pre-filter eliminates the root cause.** With ~0 allocations per scan (instead of ~12MB), the heap doesn't grow in the first place. No wound to bandage.
2. **FreeOSMemory() adds runtime overhead.** It triggers a full stop-the-world GC + madvise syscall every 5 minutes, introducing scheduling jitter. Unnecessary once the allocation rate is fixed.

No new imports, no runtime penalty.

Closes #2261
Related: #546

### DIFF
--- a/internal/persis/file/dagrun/pagination.go
+++ b/internal/persis/file/dagrun/pagination.go
@@ -250,7 +250,7 @@ func (it *dagRunStatusIterator) loadDay(ctx context.Context, dayPath string) ([]
 		return nil, fmt.Errorf("read day directory %s: %w", dayPath, err)
 	}
 
-	runs, err := loadDayRuns(dayPath, entries)
+	runs, err := loadDayRuns(dayPath, entries, it.statusesFilter, it.hasStatusFilter)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


yohamta0:
LGTM, thank you very much for the improvement!

yohamta0 (approved):

PR #2320, opened 2026-06-23
fix: resolve shared-nothing secret refs via coordinator

### LINKED ISSUE #2318: feat: Support secret resolution for remote workers in shared nothing mode

When using Dagu in **shared nothing mode**, secrets defined in the registry are not resolved on remote workers, making it impossible to use them in distributed DAG executions.

The following error is returned at runtime:
```
failed to resolve secrets: failed to resolve secret "MY_SECRET" from registry ref "prod/my-secret": secret not found
```

This appears to happen because the remote worker does not have access to the secret registry available on the scheduler side. As a result, any DAG step relying on a secret fails immediately when dispatched to a remote worker.

Secrets should be resolvable by remote workers in shared nothing mode. Possible approaches could include:

- Resolving secrets on the scheduler side before dispatching the task to the worker, and forwarding them securely as part of the execution context.
- Allowing the remote worker to query the secret registry at runtime (e.g. via a shared endpoint or a forwarded credential).
- Documenting a supported pattern if one already exists but is undocumented.

The expected behavior is that a DAG using secrets runs identically whether executed locally or on a remote worker in shared nothing mode.

## Describe alternatives you've considered

Injecting secrets manually via the Dagu API before dispatching the task — but this requires significant additional orchestration logic on the caller side, which defeats the purpose of having a centralized secret registry.

## Use case

Running DAGs on **remote workers in shared nothing mode**, where workers do not share a filesystem or configuration with the scheduler. This is a common deployment pattern for scalable or multi-host Dagu setups.

## Additional context

- Dagu version: *2.8.2*
- Deployment: shared nothing mode with remote workers
- The secret is defined in the registry under `prod/my-secret` and works correctly in local (non-remote) execution
- No additional configuration was found in the documentation regarding secret forwarding to remote workers

yohamta0:
## Summary
- add coordinator-backed secret reference resolution for shared-nothing workers
- route DAG registry refs through one ReferenceResolver abstraction while preserving local/shared-volume behavior
- add regression coverage for coordinator-owned secrets with an empty worker-local store

## Root cause
Shared-nothing workers reloaded the dispatched DAG and resolved secrets[].ref through their own local SecretStore. That store does not contain coordinator-owned registry secrets, so valid refs failed with secret not found.

## Testing
- go test ./internal/runtime/agent ./internal/service/worker ./internal/service/coordinator -run 'TestAgent_OutputSecretMasking|TestAgent_RegistryRefSecretResolution|TestRemoteTaskHandlerResolvesRegistrySecretsViaCoordinator|TestResolveSecretReference' -count=1
- make test TEST_TARGET="./internal/service/worker ./internal/service/coordinator ./internal/runtime/agent ./internal/proto/convert ./internal/cmd"

Closes #2318

---
## Summary by cubic
Route DAG secret registry refs through the coordinator for shared‑nothing workers to fix “secret not found” failures. Resolution is scoped to active attempts so workers can resolve or check refs without local store access. Closes #2318.

- **New Features**
  - Added `ResolveSecretReference` RPC on `CoordinatorService` with check‑only mode and run scoping (worker ID, attempt key/ID).
  - Injected a coordinator‑backed `SecretReferenceResolver` into remote workers; local runs keep using the local store.

- **Bug Fixes**
  - Shared‑nothing workers no longer read an empty local secret store; registry refs resolve via the coordinator.
  - Coordinator enforces active‑attempt checks before returning values; notification route tests are now deterministic to avoid flakiness.

<sup>Written for commit 3373043c9a08852d505e56f6e906a3089c85a93b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added secret reference resolution for shared-nothing worker deployments through the coordinator
  * Workers can now resolve and validate managed secrets, including workspace-scoped references
  * Added verification mode for access validation without exposing plaintext values

* **Tests**
  * Added comprehensive test coverage for secret resolution flows

### DIFF
--- a/internal/cmd/coord.go
+++ b/internal/cmd/coord.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"os"
 
+	cmdprocess "github.com/dagucloud/dagu/internal/cmd/process"
 	"github.com/dagucloud/dagu/internal/cmn/config"
 	"github.com/dagucloud/dagu/internal/cmn/logger"
 	"github.com/dagucloud/dagu/internal/cmn/logger/tag"
@@ -225,6 +226,7 @@ func newCoordinator(
 	}
 
 	// Create handler with DAGRunStore for status persistence and LogDir for log streaming
+	agentStores := cmdprocess.NewRuntimeAgentStores(ctx.Context, cfg)
 	handler := coordinator.NewHandler(coordinator.HandlerConfig{
 		DAGRunStore:               dagRunStore,
 		StateStore:                stateStore,
@@ -237,6 +239,7 @@ func newCoordinator(
 		DAGRunLeaseStore:          dagRunLeaseStore,
 		ActiveDistributedRunStore: activeDistributedRunStore,
 		DAGStore:                  dagStore,
+		SecretStore:               agentStores.SecretStore,
 		EventService:              ctx.EventService,
 		EventSourceInstance:       ctx.EventSourceInstance,
 	})
--- a/internal/runtime/agent/agent.go
+++ b/internal/runtime/agent/agent.go
@@ -237,6 +237,8 @@ type Agent struct {
 	profileResolvedAt string
 	// profileEntries records non-secret injected key metadata for status/history.
 	profileEntries []exec.RuntimeProfileEntry
+	// secretReferenceResolver resolves registry refs without requiring local store access.
+	secretReferenceResolver secrets.ReferenceResolver
 	// secretMasker redacts resolved secret values from status/history snapshots.
 	secretMasker *masking.Masker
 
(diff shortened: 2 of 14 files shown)


l-we:
An error occurred when I configured two sets of keys.

```
secrets:
  - name: DB_PASSWORD
    ref: dev/db-password
  - name: DB_PW_IT
    ref: dev/db-pw-it
```

```
"err":"failed to resolve secrets: failed to resolve secret \"DB_PW_IT\" from registry ref \"dev/db-pw-it\": resolve secret reference failed: rpc error: code = NotFound desc = secret not found"
```

yohamta0:
Thanks, let me check and fix it if it's a bug.

PR #2383, opened 2026-07-12
perf(ui): defer heavy route dependencies

yohamta0:
## Summary

- split protected UI pages into route-level lazy chunks while keeping login and setup in the bootstrap
- load the Scalar API reference only after its document is ready and load Monaco only when the DAG editor renders
- cache the versioned main bundle only when its query version matches the embedded bundle hash

## Why

Every page was synchronously reachable from `App.tsx`, which pulled Scalar, Vue, CodeMirror, Monaco, terminal, charting, and administration code into the startup bundle even when those routes were not visited.

## Impact

The production entry bundle drops from 18.02 MB to 1.03 MB raw and from 4.63 MB to about 303 KB with the server's gzip level. Total generated assets remain essentially flat because the heavy code moves into content-hashed deferred chunks.

## Validation

- `pnpm typecheck`
- `pnpm test` (104 files, 419 tests)
- focused App and API docs tests
- `go test ./internal/service/frontend -count=1`
- `pnpm build`
- `git diff --check`

---
## Summary by cubic
Defers heavy UI route dependencies and adds version-aware caching for the main `bundle.js`, dropping the production entry bundle from 18.02 MB to 1.03 MB raw (4.63 MB → ~303 KB gzipped). Adds a lazy-route error boundary with a reload prompt and tests to validate bundle caching and stabilize distributed cancellation checks.

- **Refactors**
  - Lazy-load protected pages and heavy editors (`ScalarViewer`, `DAGEditor`) behind `React.Suspense` with lightweight loaders; keep login and setup in the bootstrap.
  - Add an error boundary for lazy routes that shows a reload action when a chunk fails to load.
  - Cache `bundle.js` for 1 year when `?v` matches `currentAssetVersion()`; otherwise, disable JS caching.
  - Add tests for current-version bundle caching via asset routes and for the lazy-load failure UI; stabilize distributed lifecycle checks by reading running tasks from the coordinator.

<sup>Written for commit 9d2f0a7f232115cc5d2a30e843321532e4ea32c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance**
  * Improved long-lived caching for the main `bundle.js` only when it’s the current client-referenced version.
  * Refined cache-control for JavaScript assets, workers, and other static files for more accurate browser behavior.
* **User Experience**
  * Implemented lazy loading across most pages/components with a loading indicator.
  * Added a user-facing “Unable to load this page” message and a “Reload” button on lazy-load failures.
* **Bug Fixes**
  * Corrected frontend asset cache-control for both versioned and non-versioned requests.
* **Tests**
  * Updated cache-header assertions and expanded lazy-load failure coverage.

### DIFF
--- a/internal/intg/distr/lifecycle_test.go
+++ b/internal/intg/distr/lifecycle_test.go
@@ -81,7 +81,7 @@ worker_selector:
 steps:
   - name: long-sleep
     run: %s
-`, test.ShellQuote(test.Sleep(30*time.Second))), withLabels(map[string]string{"foo": "bar"}))
+`, test.ShellQuote(test.Sleep(1000*time.Second))), withLabels(map[string]string{"foo": "bar"}))
 		defer f.cleanup()
 
 		require.NoError(t, f.start())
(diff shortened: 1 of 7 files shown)


yohamta0:
Addressed both CodeRabbit nitpicks in 409317e5c:\n\n- use the existing lowercase basename for the versioned main-bundle comparison\n- add a lazy-route error boundary with a reload recovery path and a regression test for rejected chunk imports\n\nValidation: full frontend test suite (104 files / 420 tests), TypeScript typecheck, full internal/service/frontend Go tests, and production Webpack build.\n\nThe generic docstring-coverage warning was not applied: these are internal implementation helpers, and this repository's documentation conventions explicitly avoid implementation-oriented godoc.

yohamta0:
@coderabbitai review

yohamta0:
Pushed `9d2f0a7f2` to address the failed CI shards properly:

- the asset-route test now supplies a controlled in-memory bundle, so it does not depend on whether the standalone Go job built `ui/dist`
- the distributed cancellation test now synchronizes on the coordinator worker registry and keeps the child alive until cancellation, instead of racing a transient parent-status snapshot

Validated with the full frontend server package, the focused cancellation test across repeated runs, and the full distributed integration package.

@coderabbitai review

PR #2394, opened 2026-07-19
perf: avoid bcrypt on API key requests

yohamta0:
## Summary

- add a versioned SHA-256 digest and an in-memory digest-to-ID index for API key lookup
- lazily promote existing bcrypt-only keys after their next successful credential check
- serialize legacy verification and skip migrated records so concurrent API requests do not run bcrypt in parallel
- coalesce API key `LastUsedAt` writes to at most once per minute

## Root cause

Every API-key-authenticated REST and MCP request listed all stored keys and ran a cost-12 bcrypt comparison for the matching prefix. Concurrent clients therefore consumed CPU in proportion to request rate, and every successful request also performed a synchronous atomic file write for `LastUsedAt`.

## Compatibility and migration

The stored `key_digest` field is optional. New keys populate it immediately, while existing keys are upgraded automatically on first valid use. The bcrypt hash remains stored for rollback compatibility. No configuration, public API, key format, or startup migration is required.

Metadata updates preserve the latest credential fields and `LastUsedAt`, preventing concurrent promotion or usage tracking from being overwritten.

## Validation

- `make lint`
- `go test -race ./internal/auth ./internal/persis/store ./internal/service/auth -count=1`
- `go test ./internal/service/frontend/auth ./internal/service/frontend/api/v1 -count=1`
- `go test ./... -run '^$' -count=1`
- parallel digest-path benchmark with a CPU profile; bcrypt and Blowfish had no samples on the steady-state path

---
## Summary by cubic
Avoided cost-12 bcrypt on API key requests by switching to a fast, indexed validation path and throttling `LastUsedAt` writes. Existing keys migrate automatically on first successful use; no config or API changes required.

- **Refactors**
  - Store: added `GetByDigest` and `PromoteDigest`, in-memory digest→ID index, and `UpdateLastUsed` coalesced to once per minute; updates preserve stored credential fields and newer `LastUsedAt`.
  - Service: validation computes a versioned SHA-256 digest and uses digest lookup; legacy bcrypt path is serialized and promotes the digest on success; last-used updates are throttled and non-fatal.
  - Data model and tests: introduced `KeyDigest` on `APIKey`/`APIKeyForStorage` (optional, populated for new keys), indices rebuilt on startup with uniqueness enforcement; added tests for migration, throttling, concurrency, and a steady-state benchmark.

<sup>Written for commit 6a2cec381680c87ee9bfa4c5ea02bc1b2fb0540b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2394?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * API keys now support faster digest-based validation and lookup.
  * Existing legacy API keys are automatically migrated when successfully validated.
  * API key credentials and usage timestamps are preserved during updates.
  * “Last used” activity updates are throttled to reduce unnecessary writes.
  * Duplicate credential digests are detected and rejected.

* **Bug Fixes**
  * Improved consistency of API key indexes after creation, updates, deletion, and restart.
  * Validation remains successful when recording usage metadata fails.

### DIFF
--- a/internal/auth/apikey.go
+++ b/internal/auth/apikey.go
@@ -65,6 +65,9 @@ type APIKey struct {
 	// KeyHash is the bcrypt hash of the API key secret.
 	// Excluded from JSON serialization for security.
 	KeyHash string `json:"-"`
+	// KeyDigest is the versioned digest used for API key lookup.
+	// Excluded from JSON serialization for security.
+	KeyDigest string `json:"-"`
 	// KeyPrefix stores the first 8 characters of the key for identification.
 	KeyPrefix string `json:"key_prefix"`
 	// CreatedAt is the timestamp when the API key was created.
@@ -108,7 +111,7 @@ func NewAPIKey(name, description string, role Role, keyHash, keyPrefix, createdB
 }
 
 // APIKeyForStorage is used for JSON serialization to persistent storage.
-// It includes the key hash which is excluded from the regular APIKey JSON.
+// It includes credential fields excluded from the regular APIKey JSON.
 type APIKeyForStorage struct {
 	ID                       string                 `json:"id"`
 	Name                     string                 `json:"name"`
@@ -123,6 +126,7 @@ type APIKeyForStorage struct {
 	ServiceAccountName       string                 `json:"service_account_name,omitempty"`
 	MigratedAsServiceAccount bool                   `json:"migrated_as_service_account,omitempty"`
 	KeyHash                  string                 `json:"key_hash"`
+	KeyDigest                string                 `json:"key_digest,omitempty"`
 	KeyPrefix                string                 `json:"key_prefix"`
 	CreatedAt                time.Time              `json:"created_at"`
 	UpdatedAt                time.Time              `json:"updated_at"`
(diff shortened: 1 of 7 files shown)


yohamta0:
Reviewed the CodeRabbit legacy-path nitpick. No code change is appropriate in this PR:

- The process-wide migration mutex deliberately bounds legacy bcrypt to one comparison at a time, while new and promoted keys never enter this path.
- The auth service has no trustworthy client/IP identity. A global limiter here would let invalid traffic consume the migration allowance and reject or starve legitimate one-time legacy migrations.
- Existing bcrypt hashes cannot be background-migrated into a digest without receiving the plaintext key.

A separate request-level limiter at the HTTP/MCP boundary could be designed with the correct client identity and response semantics, but adding a global service limiter or a bypassable negative cache here would be a availability band-aid rather than a sound fix. Dormant keys intentionally remain untouched per this PR's migration scope.

PR #2401, opened 2026-07-21
feat: add fallback pull policy

dodyirawan85:
## Summary

Add a new pull_policy: fallback that attempts to pull the image from the registry first but falls back to the existing local image if the pull fails (e.g., registry down). Fails with an error if no local image is available.

## Changes

- Add PullPolicyFallback enum to core.Container
- Extract hasLocalImage helper from shouldPullImage in docker client
- Handle ImagePull errors with local image fallback in startNewContainer
- Map fallback to PullIfNotPresent in Kubernetes config

## Checklist

---
## Summary by cubic
Add a new image pull policy `fallback` that tries to pull first and, if the pull fails, uses a local, platform-compatible image; errors if none exists. Supported in Docker; Kubernetes rejects this policy.

- **New Features**
  - Added `PullPolicyFallback` to `core.Container` and schema; accepts "fallback".
  - Docker: parses the ImagePull JSON stream to detect registry errors and falls back to a local image on failure; warns and errors if no local image is found.
  - Kubernetes: returns `ErrUnsupportedFallbackPolicy` for "fallback" (use `Always`, `IfNotPresent`, or `Never`).

- **Bug Fixes**
  - Docker: variant-tolerant platform matching using `platforms.OnlyStrict` (treats arm64 and arm64/v8 as equivalent; does not match incompatible variants like arm64/v7).

<sup>Written for commit cce6aaf7e3eda852bd98231050b218e35d3896fc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a `fallback` image pull policy for Docker and Kubernetes workloads.
  * The policy attempts to pull the image first, then uses a matching local image if the pull fails.
  * Docker validates local image platform compatibility before falling back.
* **Bug Fixes**
  * Improved handling and reporting of image pull failures.
* **Tests**
  * Added coverage for parsing, validation, and execution of the new policy.

### DIFF
--- a/internal/runtime/builtin/docker/client.go
+++ b/internal/runtime/builtin/docker/client.go
@@ -6,6 +6,7 @@ package docker
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -759,16 +760,38 @@ func (c *Client) startNewContainer(ctx context.Context, name string, cli *client
 
 		logger.Debug(ctx, "Docker: startNewContainer calling ImagePull")
 		reader, err := cli.ImagePull(ctx, c.cfg.Image, pullOpts)
+
+		// Check for API-level error
 		if err != nil {
 			logger.Error(ctx, "Docker: startNewContainer ImagePull failed", tag.Error(err))
-			return "", err
+		} else {
+			// Check for stream-level error (registry errors reported in JSON)
+			logger.Debug(ctx, "Docker: startNewContainer ImagePull returned, checking stream for errors")
+			err = checkImagePullStream(reader)
+			_ = reader.Close()
+			if err != nil {
+				logger.Error(ctx, "Docker: startNewContainer image pull stream error", tag.Error(err))
+			}
 		}
-		logger.Debug(ctx, "Docker: startNewContainer ImagePull returned, draining reader")
-		// Output pull-image log to stderr instead of stdout
-		_, _ = io.Copy(io.Discard, reader)
-		_ = reader.Close()
-		logger.Debug(ctx, "Docker: startNewContainer image pull completed")
-		logger.Infof(ctx, "Successfully pulled the image %q", c.cfg.Image)
+
+		// Handle pull failure with unified fallback logic
+		if err != nil {
+			if c.cfg.Pull == core.PullPolicyFallback {
+				hasLocal, checkErr := c.hasLocalImage(ctx, cli, &c.platform)
+				if checkErr != nil {
+					return "", fmt.Errorf("image pull failed and local image check failed: %w (original pull error: %v)", checkErr, err)
+				}
+				if !hasLocal {
+					return "", fmt.Errorf("image pull failed and no local image available: %w", err)
+				}
+				logger.Warnf(ctx, "Pull failed for %q, falling back to local image: %v", c.cfg.Image, err)
+			} else {
+				return "", err
+			}
+		} else {
+			logger.Debug(ctx, "Docker: startNewContainer image pull completed")
+			logger.Infof(ctx, "Successfully pulled the image %q", c.cfg.Image)
+		}
 	}
 
 	ctCfg := *c.cfg.Container // Copy to avoid mutating original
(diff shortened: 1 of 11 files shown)


internal/runtime/builtin/docker/client.go
```diff
@@ -761,14 +761,26 @@ func (c *Client) startNewContainer(ctx context.Context, name string, cli *client
...
+					return "", fmt.Errorf("image pull failed and no local image available: %w", err)
+				}
+				logger.Warnf(ctx, "Pull failed for %q, falling back to local image: %v", c.cfg.Image, err)
+			} else {
+				return "", err
+			}
+		} else {
+			logger.Debug(ctx, "Docker: startNewContainer ImagePull returned, draining reader")
+			// Output pull-image log to stderr instead of stdout
+			_, _ = io.Copy(io.Discard, reader)
```
yohamta0: `ImagePull` can report registry errors in the JSON response stream even when the call itself returns no error. Discarding the stream here treats those failures as successful pulls, so the fallback path is skipped. Could we decode the stream and handle any reported error through the same fallback logic?

internal/runtime/builtin/kubernetes/config.go
```diff
@@ -1385,6 +1385,11 @@ func (cfg *Config) toK8sImagePullPolicy() (corev1.PullPolicy, error) {
 		return corev1.PullNever, nil
 	case "ifnotpresent":
 		return corev1.PullIfNotPresent, nil
+	case "fallback":
```
yohamta0, line 1389: `IfNotPresent` does not try the registry when the image is cached, so this can run a stale image instead of using the pull-first behavior promised by `fallback`. Since Kubernetes has no equivalent policy, should we reject `fallback` for this executor?

yohamta0 (changes requested):
Thanks for adding this. I left two inline comments about pull error handling and the Kubernetes behavior. CI also needs `gofmt` on `internal/core/container.go`.

internal/runtime/builtin/docker/client.go
dodyirawan85: Done

internal/runtime/builtin/kubernetes/config.go
dodyirawan85: Sure, now it return value to reject the fallback

internal/runtime/builtin/docker/client.go
```diff
@@ -1320,16 +1343,51 @@ func getPlatform(ctx context.Context, cli *client.Client, cfg *Config) (specs.Pl
...
+	}
+
+	for {
+		var msg pullMessage
+		if err := decoder.Decode(&msg); err != nil {
+			if err == io.EOF {
+				break
+			}
+			// Ignore decode errors in the stream, continue reading
+			continue
```
yohamta0: Should this return the decode error? A truncated stream could otherwise loop here forever.

yohamta0 (commented):
One remaining issue in the updated pull stream handling.

internal/runtime/builtin/docker/client.go
dodyirawan85: Done

dodyirawan85:
Added new commit (62fdbc6a950f8ebf885ce33d754ba67c68c38018), due to docker behaviour on arm64 system.
It somehow keeps pulling "v8" variant automatically while the checks are empty ("" != "v8")
What do you think about this workaround ?

yohamta0:
@dodyirawan85 Thanks for testing this on arm64. The `""` and `"v8"` mismatch is a normalization issue, but treating an empty variant as a wildcard is too broad. Could you compare the platforms with `platforms.OnlyStrict` instead? That should treat `arm64` and `arm64/v8` as equivalent without matching incompatible ARM variants.

```go
localPlatform := specs.Platform{
	OS:           inspect.Os,
	Architecture: inspect.Architecture,
	Variant:      inspect.Variant,
}

if platforms.OnlyStrict(*platform).Match(localPlatform) {
	return true, nil
}
```

dodyirawan85:
> @dodyirawan85 Thanks for testing this on arm64. The `""` and `"v8"` mismatch is a normalization issue, but treating an empty variant as a wildcard is too broad. Could you compare the platforms with `platforms.OnlyStrict` instead? That should treat `arm64` and `arm64/v8` as equivalent without matching incompatible ARM variants.
> 
> ```go
> localPlatform := specs.Platform{
> 	OS:           inspect.Os,
> 	Architecture: inspect.Architecture,
> 	Variant:      inspect.Variant,
> }
> 
> if platforms.OnlyStrict(*platform).Match(localPlatform) {
> 	return true, nil
> }
> ```

Done, tested and works on arm64 platform

yohamta0 (approved):
LGTM 🚀🚀🚀 Thank you for adding support for fallback pull policy!

PR #2408, opened 2026-07-23
feat: reconnect worker to coordinator after heartbeat timeout

pirm-in:
## Summary

Workers can get stuck talking to a dead cached coordinator connection when a
coordinator restarts or moves to a new address under the same ID. This drops
and re-dials the cached gRPC client on transient failures, and bounds the
worker heartbeat with a configurable timeout so a stalled coordinator cannot
block the heartbeat loop.

## Changes

- Reset (close + drop) the cached coordinator client on transient gRPC errors (Unavailable / deadline) so the next call re-dials
- Add configurable `HeartbeatTimeout` (default 10s) bounding the worker heartbeat
- Health check reuses the already-resolved client instead of re-resolving it

## Related Issues

_none_

## Checklist

- [x] Documentation has been updated as needed (N/A - config not user-facing)
- [x] Changes have been tested locally (`make test`, `make lint`)

---
## Summary by cubic
Workers now reliably reconnect to coordinators after heartbeat timeouts and address/instance changes. Heartbeat failover is connection-safe and won’t tear down active streams.

- **Bug Fixes**
  - Make the discovered coordinator address and `StartedAt` authoritative; refresh the cached client only on address/instance changes under the same ID. Applied to dispatch, stream failover, workspace bundle uploads, status reporting, and worker listing.
  - Skip health checks when the context is done and check health on the resolved client to avoid evicting healthy connections; transient RPC errors no longer interrupt active log/artifact streams.
  - Stop failover when the request context ends; pre-warm the pinned state coordinator client to avoid late failures.

- **New Features**
  - Added `HeartbeatTimeout` to `coordinator.Config` (default 10s). The budget is split across members during failover, order is randomized, and caller deadlines are honored.

<sup>Written for commit 7ec212599dbf9e27482114e6b47ef64cd8107449. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Improved coordinator recovery after transient failures and expired deadlines.
  * Prevented concurrent operations from discarding newer coordinator connections.
  * Added configurable heartbeat timeouts, defaulting to 10 seconds.
  * Improved failover for heartbeats, stream connections, and workspace uploads.

* **Bug Fixes**
  * Heartbeats now reconnect successfully after coordinator deadline failures.
  * Stale or unhealthy connections are safely removed before retrying.

### DIFF
--- a/internal/service/coordinator/client.go
+++ b/internal/service/coordinator/client.go
@@ -317,12 +317,13 @@ func (cli *clientImpl) attemptCall(ctx context.Context, members []exec.HostInfo,
 		}
 
 		// Check if the coordinator is healthy
-		if err := cli.isHealthy(ctx, member); err != nil {
+		if err := cli.isHealthy(ctx, client); err != nil {
 			logger.Warn(ctx, "Failed to check coordinator health",
 				slog.String("coordinator-id", member.ID),
 				tag.Host(member.Host),
 				tag.Port(member.Port),
 				tag.Error(err))
+			cli.resetClientOnTransientError(member, client, err)
 			cli.recordFailure(err)
 			lastErr = err
 			continue
@@ -341,6 +342,7 @@ func (cli *clientImpl) attemptCall(ctx context.Context, members []exec.HostInfo,
 			if errors.Is(err, backoff.ErrPermanent) {
 				return err
 			}
+			cli.resetClientOnTransientError(member, client, err)
 
 			lastErr = err
 		} else {
@@ -523,13 +525,7 @@ func (cli *clientImpl) callMemberWithTimeout(ctx context.Context, member exec.Ho
 	return cli.callMember(callCtx, member, callback)
 }
 
-func (cli *clientImpl) isHealthy(ctx context.Context, member exec.HostInfo) error {
-	// Get or create client for this coordinator
-	client, err := cli.getOrCreateClient(member)
-	if err != nil {
-		return fmt.Errorf("failed to get coordinator client: %w", err)
-	}
-
+func (cli *clientImpl) isHealthy(ctx context.Context, client *client) error {
 	// Check health
 	req := &grpc_health_v1.HealthCheckRequest{
 		Service: "", // Check overall server health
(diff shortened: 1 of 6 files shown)


internal/service/coordinator/client.go
```diff
@@ -547,6 +542,16 @@ func (cli *clientImpl) isHealthy(ctx context.Context, member exec.HostInfo) erro
 	return nil
 }
 
+// checkHealthy verifies the coordinator is serving and evicts the cached
+// client when the failure indicates a stale connection.
+func (cli *clientImpl) checkHealthy(ctx context.Context, member exec.HostInfo, client *client) error {
+	if err := cli.isHealthy(ctx, client); err != nil {
```
yohamta0: Could we bail out before the health check when `ctx` is already done?

`Heartbeat` shares one timeout across all coordinator attempts. If one coordinator uses the full timeout, the next healthy coordinator gets `DeadlineExceeded` immediately and this code closes its cached `ClientConn`, canceling any active log or artifact stream. I reproduced this with two coordinators.

```suggestion
	if err := ctx.Err(); err != nil {
		return err
	}
	if err := cli.isHealthy(ctx, client); err != nil {
```

yohamta0 (commented):
See inline comment.

internal/service/coordinator/client.go
pirm-in: Good catch, thanks! I reproduced it too. I did not try with two coordinators, just a single one, so will remember that. Coding agent left additional feedback: 
  
>  Added the `ctx.Err()` bail-out at the top of `checkHealthy` (rather than inline at one site), so it covers every failover path that routes through it (`attemptCall`, `openStreamWithFailover`, `PutWorkspaceBundle`), plus a regression test (`TestClientHeartbeatExpiredContextDoesNotEvictUntriedCoordinator`) that stands up two coordinators with live log streams and asserts only the coordinator that actually failed gets evicted.
> 
>  One thing I want to flag honestly: the guard closes the reported case (a coordinator whose turn starts with an already-expired context), but it's a point-in-time check at `checkHealthy` entry. If a coordinator enters with a small-but-nonzero sliver of the shared heartbeat budget, the guard passes and expiry during the health-check RPC, or during the `Heartbeat` callback itself, can still evict that coordinator's `ClientConn` and cancel its streams - the same class of bug in a narrower window.
> 
> The more complete fix would be a per-attempt timeout inside `Heartbeat`'s callback, the way `Dispatch` already does it, so each coordinator gets its own full budget and `DeadlineExceeded` only ever means "this coordinator was genuinely slow." Trade-off is worst-case total latency of roughly N × `HeartbeatTimeout`, which is the same trade-off `Dispatch` accepts today. (Would need to stay in the `Heartbeat` callback specifically - `Poll` shares `attemptCall` and intentionally has no per-attempt timeout for long-polling.)
>
> Happy to go either way: keep the guard as the minimal fix for now, or fold in the per-attempt timeout in this PR. Let me know which you'd prefer.

yohamta0 (approved):
LGTM 🚀🚀🚀 Thanks for catching and fixing the bug!

yohamta0:
Found some issues, I'll push to the PR.

PR #2415, opened 2026-07-24
fix: reuse succeeded parallel child runs

yohamta0:
## Summary

- reuse persisted successful child DAG-run results during parallel step retries
- avoid creating redundant child attempts in both local and distributed execution
- preserve stored child outputs for parallel result aggregation

## Root cause

Parallel child run IDs are deterministic, but the local and coordinator-backed child runners treated every existing child run as retryable. Retrying a failed parallel parent step therefore created a new attempt even for child runs that had already succeeded.

## Impact

Succeeded parallel children now return their persisted result without another attempt. Failed and otherwise incomplete children continue through the existing retry path. Non-parallel sub-DAG retry behavior is unchanged.

## Testing

- `make fmt`
- `go test ./internal/subflow ./internal/runtime/executor ./internal/runtime/builtin/dag -count=1`
- `go test ./internal/intg -run '^TestParallelExecution_' -count=1`

---
## Summary by cubic
Reuse successful parallel child runs during external step retries to avoid extra attempts and keep outputs intact in both local and distributed runners.

- **Bug Fixes**
  - On `ExternalStepRetry`, if a child run is already `Succeeded`, return its stored status and outputs; do not dispatch or create a new attempt.
  - Applies to both local and coordinator-backed runners; the original child attempt ID is reused.
  - Added tests for local and distributed paths verifying no dispatch, attempt ID reuse, and output preservation.

<sup>Written for commit 46880a86829dfc80c40c7ef1e22ce27dfe6e496e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2415?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * External workflow step retries now reuse previously completed successful child runs.
  * Prevented unnecessary workflow re-execution and duplicate attempts during retries.
  * Preserved the original run status and output values when reusing a successful result.


yohamta0:
Addressed CodeRabbit’s output-preservation nitpick in `46880a86`: the local reuse regression test now seeds a persisted child output and verifies it is returned while retaining the original attempt assertion.

PR #2417, opened 2026-07-24
feat: support variable substitution in worker_selector keys and values

### LINKED ISSUE #2410: feat: Support variable / parameter substitution in `worker_selector` values
## Summary
 
I would like `worker_selector` values to support variable and parameter substitution, so a selector label can be driven by a variable (e.g. `${WORKLOAD}`) instead of being hardcoded in every DAG. Today the value is treated literally, so `workload: ${WORKLOAD}` never matches a worker.
 
## Motivation
 
Our main goal is **routing by workspace**. We run several workspaces, and each workspace should execute on its own dedicated set of workers. We want to define the target worker label **once per workspace** and have every DAG in that workspace inherit it — rather than repeating the same `worker_selector` label in 200+ individual DAG files.
 
Dagu already lets us set workspace-wide values (e.g. `WORKLOAD=intraday`) in a per-workspace base config, and every DAG in that workspace inherits it. The only missing piece is that `worker_selector` doesn't resolve that inherited variable. If it did, workspace-based routing would work with zero per-DAG configuration.
 
Without substitution we're forced to either hardcode the worker label in every DAG (which defeats the point of workspace-level config) or generate the DAGs programmatically. The same limitation also blocks other parameter-driven routing, e.g. selecting a specific node.
 
## Our setup
 
- We run **200+ facility pipelines**. Each facility has its own sub-DAG, stored in a per-facility subfolder. These files are synced onto the server.
- A **parent DAG** reads a list of facilities and fans them out in parallel, running each facility's sub-DAG via a parallel `dag.run` step over that list.
- The **parent runs locally** (no `worker_selector` on the parent). This is required so it can resolve the sub-DAG file paths from the local DAG store — if the parent were dispatched to a worker, it wouldn't find the child DAGs.
- The **child sub-DAGs** should run on the workers, selected via `worker_selector`.
- Workspace-wide settings live in a per-workspace base config, so all children in a workspace share the same environment.
Simplified parent DAG:
 
```yaml
name: intraday-prod
labels:
  - workspace=intraday
  - run_mode=prod
 
env:
  - WORKLOAD: intraday   # inherited from the workspace base config
 
steps:
  - id: load_subdags
    # produces the list of facility sub-DAGs to run
    outputs:
      - SUBDAG_ITEMS
 
  - id: run_downstreams
    run: ${JOBS_ROOT}/facilities/${ITEM}
    parallel:
      items: ${load_subdags.outputs.SUBDAG_ITEMS}
    worker_selector:
      workload: ${WORKLOAD}   # want: resolves to "intraday" and routes children to intraday workers
    continue_on:
      failure: true
```
 
Because `${WORKLOAD}` in `worker_selector` isn't substituted, we currently have to write `workload: intraday` literally in the parent — which brings the per-DAG hardcoding back and prevents us from managing the routing label centrally per workspace.
 
## What I'd like
 
`worker_selector` keys/values should be evaluated the same way other step fields (`command`, `args`, `env`, `run`, sub-DAG name, …) already are — resolving variables from the workspace/DAG environment and from `parallel` item context. That way we set the worker label once per workspace and every DAG routes correctly with no per-DAG duplication.
 
## Environment
 
- Dagu version: v2.10.7 
- Deployment: Docker Compose, distributed coordinator + multiple workspace-tagged worker pools
- OS: Linux

Giebisch:
## Summary
- resolve `${VAR}` references in `worker_selector` keys and values so routing labels can be defined once (e.g. in workspace base-config env) instead of hardcoded in every DAG
- DAG-level selectors resolve at build time after base-config composition, with the same semantics as `env:` (undefined variables stay literal, skipped under `BuildFlagNoEval`)
- step-level selectors resolve at runtime before the sub-DAG run/enqueue request is created; parallel children can route per item via `${ITEM}`
- add spec, loader (base-config inheritance), and runtime regression coverage

## Changes
- add a post-compose `resolveWorkerSelector` build step for the root selector in `internal/core/spec/dag.go`, reusing the `evaluatePairs` resolver construction
- add a runtime `resolveWorkerSelector` helper in `internal/runtime/builtin/dag` and wire it into the `dag.run`, parallel, and enqueue executors (parallel children resolve with the item context as `${ITEM}`)
- keep the string form `worker_selector: local` literal; evaluate selector values with the strict workflow field policy, so no command substitution runs in routing metadata
- update the value-resolution matrix in `specs/003-value-resolution.md`, the JSON schema descriptions, and `README_SCHEMA.md`

## Impact
Selectors flow through dispatch and coordinator matching already resolved; no changes to the dispatch decision, proto conversion, or matching logic. DAGs with literal selectors behave exactly as before, and undefined variables stay literal (the label simply never matches, the same failure mode as a typo today). No configuration or API changes.

## Related Issues
Closes #2410

## Testing
- `go test ./internal/core/spec/... -count=1` (includes new `TestWorkerSelectorEvaluation`, `TestLoad_WorkerSelectorFromBaseConfigEnv`)
- `go test ./internal/runtime/builtin/dag/... -count=1` (includes new `TestResolveWorkerSelector`)
- full `make test` (12k+ tests) and `make lint` pass
- verified live on a local distributed stack (coordinator + two labeled workers): DAG-level selector follows base-config env, step-level selector dispatches the child correctly, and a parallel step over `[intraday, batch]` with `workload: ${ITEM}` lands one child on each worker

### Manual verification
```bash
make bin
mkdir -p /tmp/wsdemo/{dags,data,logs}

cat > /tmp/wsdemo/config.yaml <<'EOF'
host: 127.0.0.1
port: 8080
auth: { mode: none }
coordinator: { enabled: true, host: 127.0.0.1, advertise: 127.0.0.1, port: 50055, health_port: 0 }
queues: { enabled: true }
paths:
  data_dir: /tmp/wsdemo/data
  dags_dir: /tmp/wsdemo/dags
  log_dir: /tmp/wsdemo/logs
  base_config: /tmp/wsdemo/base.yaml
EOF

# the workspace-level routing label — the only place it is defined
printf 'env:\n  WORKLOAD: intraday\n' > /tmp/wsdemo/base.yaml

cat > /tmp/wsdemo/dags/route-demo.yaml <<'EOF'
worker_selector:
  workload: ${WORKLOAD}
steps:
  - name: report
    run: echo "routed via workspace env"
EOF

# run each in its own terminal
.local/bin/dagu coordinator -c /tmp/wsdemo/config.yaml
.local/bin/dagu server -c /tmp/wsdemo/config.yaml
.local/bin/dagu worker -c /tmp/wsdemo/config.yaml --worker.id=w-intraday \
  --worker.labels workload=intraday --worker.coordinators=127.0.0.1:50055 --worker.health-port=0

# trigger via the API and check the assigned worker
curl -s -X POST localhost:8080/api/v1/dags/route-demo/start -H 'Content-Type: application/json' -d '{}'
sleep 3
curl -s 'localhost:8080/api/v1/dag-runs?name=route-demo' | grep -o '"workerId":"[^"]*"'
# -> "workerId":"w-intraday"; the worker log shows: worker-selector=map[workload:intraday]
# flip /tmp/wsdemo/base.yaml to a label with no worker and re-trigger:
# dispatch fails with "no workers match the required selector" showing the RESOLVED value
```

I've not tested this in real distributed environments yet, will report back though.

## Checklist

---
## Summary by cubic
Adds variable substitution for `worker_selector` keys and values. Resolution happens at DAG build (with base-config) and at sub-DAG runtime (env, params, and `${ITEM}`), and flows via `executor.RunParams.WorkerSelector` across `dag.run`, parallel, and enqueue.

- **New Features**
  - Resolve root `worker_selector` after base-config composition; honor `BuildFlagNoEval`; add `BuildFlagDeferWorkerSelector`; keep `worker_selector: local` literal.
  - Resolve step-level selectors during sub-DAG expansion with `${ITEM}`; if no step override, resolve the child DAG’s selector against runtime params before dispatch/retry and persist on enqueue; `dag.run`, parallel, and enqueue pass selectors via `executor.RunParams.WorkerSelector`; docs and schema updated.

- **Bug Fixes**
  - Validate empty or duplicate resolved keys; error when deduplicated parallel items produce the same sub-run ID with different selectors.
  - Fail fast if a child selector requires missing params; surface the error on the parent before dispatch.
  - Keep the parallel item as runtime-only routing context and do not persist it.

<sup>Written for commit 09b28e8754ab719f8b535de56769fb6a7068f260. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2417?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added variable substitution support for worker selector keys and values.
  * Worker selectors can now reference environment variables, parameters, and parallel items using expressions such as `${VAR}` and `${ITEM}`.
  * Parallel sub-DAG runs preserve their original item context while honoring explicitly provided parameters.
* **Bug Fixes**
  * Added validation for unresolved empty or duplicate selector keys.
* **Documentation**
  * Updated schema, guide, and value-resolution documentation to describe worker selector substitution behavior.


internal/runtime/transform/node.go
```diff
@@ -25,11 +25,19 @@ func ToNodeWithStep(n *exec.Node, step core.Step) *runtime.Node {
...
+		}
 	}
 	childrenRepeated := make([]runtime.SubDAGRun, len(n.SubRunsRepeated))
 	for i, r := range n.SubRunsRepeated {
-		childrenRepeated[i] = runtime.SubDAGRun(r)
+		childrenRepeated[i] = runtime.SubDAGRun{
+			DAGRunID: r.DAGRunID,
+			Params:   r.Params,
+			DAGName:  r.DAGName,
+		}
```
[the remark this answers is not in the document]
Giebisch: @coderabbitai Ah good catch!

I've now added the ParallelItem field to the persisted exec.SubDAGRun model (as parallelItem,omitempty for backward compatibility) and propagated it through the enqueue/action sub-run recording paths, which also restores the direct struct conversions in transform/node.go, agent.go, and step_executor.go so the parallel item context now survives persistence for retry/restart. 

Does that solve your concerns ?

Giebisch:
@coderabbitai review

Giebisch:
@yohamta0 What do you think ? :)

internal/core/exec/node.go
```diff
@@ -88,6 +88,10 @@ type SubDAGRun struct {
 	// For chat tool calls, this is the tool DAG name.
 	// This field enables UI drill-down when step.call is not set.
 	DAGName string `json:"dagName,omitempty"`
+	// ParallelItem contains the original item for a parallel sub-DAG run.
+	// It is kept separate from Params because explicit sub-DAG parameters may
+	// replace the item-derived parameters.
+	ParallelItem *string `json:"parallelItem,omitempty"`
```
yohamta0: Why does the parallel item need to be persisted? It is available while expanding the parallel step, which is also when `${ITEM}` should be resolved for `worker_selector`. This looks like runtime-only execution context, not part of the sub-DAG run status.

internal/runtime/builtin/dag/dag.go
```diff
@@ -62,11 +64,16 @@ func newDAGExecutor(ctx context.Context, step core.Step) (executor.Executor, err
...
+	if err != nil {
+		_ = child.Cleanup(context.WithoutCancel(ctx))
+		return nil, err
+	}
+	if err := validateSubDAG(child.DAG, step.SubDAG.Name, workerSelector); err != nil {
 		_ = child.Cleanup(context.WithoutCancel(ctx))
 		return nil, err
 	}
-	child.SetWorkerSelector(step.WorkerSelector)
+	child.SetWorkerSelector(workerSelector)
```
yohamta0: When the step selector is empty, this falls back to `child.DAG.WorkerSelector`, which was built before `runParams.Params` is applied. A `with.params` override therefore cannot affect worker selection. The parallel path has the same issue.

internal/core/spec/dag.go
```diff
@@ -731,6 +734,81 @@ func (s *dagBuildState) composeInheritedContext() {
...
+	var paramDeclarations cmnvalue.Values
+	if s.ctx.envScope != nil {
+		consts = s.ctx.envScope.consts
+		params = s.ctx.envScope.params
+		paramsJSON = s.ctx.envScope.paramsJSON
+		paramDeclarations = s.ctx.envScope.paramDeclarations
+	}
+	resolver := cmnvalue.NewResolver(
+		cmnvalue.StaticScope{Consts: cmnvalue.Values(consts), Params: paramDeclarations},
+		cmnvalue.RuntimeScope{Consts: cmnvalue.Values(consts), Params: params, ParamsJSON: paramsJSON, Env: scope},
```
yohamta0: `s.result.Params` includes params inherited from the base config, but `${params.*}` is resolved from the pre-composition scope here. `${params.REGION}` therefore stays unresolved when `REGION` is declared only in the base config.

internal/core/spec/dag.go
```diff
@@ -731,6 +734,81 @@ func (s *dagBuildState) composeInheritedContext() {
 	s.result = merged
 }
 
+// resolveWorkerSelector expands variable references in worker_selector keys and
+// values. It must run after composeInheritedContext so labels defined in
+// base-config env resolve correctly (base env is only merged into the result
+// during composition, not during field stages).
+func (s *dagBuildState) resolveWorkerSelector() {
```
yohamta0, line 743: A selector inherited from the base config has already been resolved while building `baseDAG`. If the child overrides an env value used by that selector, the override has no effect because the expression is already gone.

yohamta0 (changes requested):
I found a few issues with when `worker_selector` is resolved. I left comments inline.

internal/core/exec/node.go
Giebisch: After verifying it again, you are right. I'm new to the codebase and coderabbit suggested it, so I thought I missed it and relied on its feedback 😅

`BuildSubDAGRuns` is re-run, so nothing has to be persisted. I'll push a fix in the upcoming days

Giebisch:
Resolved DAG-level `worker_selector` against the composed base config (so base-config env/params and `${params.*}` work) and made the empty step-selector fall back to the child's selector re-resolved with `with.params` overrides. Also made the parallel `${ITEM}` runtime-only instead of persisting it in the sub-DAG run status, since it's re-derived on every run.

yohamta0:
Sorry for taking time. I'll look into this PR today.

yohamta0:
Found some issues, I'll push to this PR.

yohamta0 (approved):
LGTM, thank you very much!

PR #2437, opened 2026-07-25
fix: finalize failed queued dispatches

### LINKED ISSUE #2436: bug: queued run that fails before startup is never dequeued and retries forever
## Describe the bug

A queued DAG-run that fails **before its subprocess starts** is never removed from the queue, so the scheduler re-dispatches it on every tick and it fails again, forever. One unrunnable run produces `ERROR` + `WARN` lines every queue interval for as long as the scheduler lives.

The dispatcher treats "could not start" as a retry problem and gives up after its backoff budget, but giving up does not dequeue the item. The next tick selects the same item again with a fresh budget.

## To Reproduce

1. Enqueue a DAG-run that cannot possibly build. Anything that fails during DAG preparation works, for example a field that the current binary rejects:

   ```yaml
   llm:
     provider: ${params.PROVIDER}   # rejected by the provider whitelist
   steps:
     - name: noop
       run: echo hi
   ```

2. Start the scheduler (or `start-all`).
3. Watch the log. The same run ID is dispatched, fails permanently, and reappears every tick.

Deleting the DAG file afterwards produces the same loop through a different error:

```
Error: failed to open log file: ... no such file or directory
```

## Expected behavior

A run that fails before startup with a non-retriable error should be dequeued and its status moved to a terminal state, exactly once, with one error logged. A poison item must not be able to occupy a queue slot indefinitely.

## Actual behavior

The same two run IDs repeat every 30s indefinitely:

```
INFO  Processing batch of items          queue=Default count=2 max-concurrency=6 alive=0
DEBUG Processing queue item              queue=Default run-id=033vfYF5mVGW5dtsZhv0D2 name=llmvar_bad
ERROR Failed to execute DAG              err="failed to prepare DAG env for subprocess: ... invalid provider ... (value: ${params.PROVIDER})"
WARN  DAG execution failed before startup was observed
WARN  Retryable operation failed with non-retriable error   attempt=2
ERROR Failed to execute DAG after retries err="permanent error: ..."
```

`count=2` is the same two items on every pass.

## Root cause

`DequeueByDAGRunID` has only four call sites, and none of them covers a failed dispatch:

| Call site | Trigger |
|---|---|
| `internal/cmd/dequeue.go` | operator runs `dagu dequeue` |
| `internal/service/frontend/api/v1/dagruns.go` | dequeued from the UI/API |
| `internal/service/scheduler/queue_dispatcher.go:777` | the DAG is **suspended** |
| — | nothing for a run that fails to start |

Normally the launched run removes its own queue item once it is alive. When preparation fails before that point, nobody removes it.

In `queue_dispatcher.go` around line 1003, the permanent error is logged and conditions are recorded, but the item is left queued:

```go
if err := backoff.Retry(ctx, operation, policy, nil); err != nil {
    logger.Error(ctx, "Failed to execute DAG after retries", tag.Error(err))
    if shouldRecordStartupCondition(err) {
        ...
        conditionStage.flush(ctx)
    }
}
```

## Why deleting the queue directory does not help

`QueueStore.List` is served from a cached index (`internal/persis/store/queue_index.go`, `cachedQueueIndexLocked` / `cacheQueueIndexLocked`). Emptying `~/.dagu/data/queue/<name>/` on disk leaves a running scheduler looping on its cached copy. Only restarting the process clears it, which makes the problem look unfixable from the outside.

## Suggested fix

In the dispatch failure path, when the error is non-retriable and the run was never observed to start:

1. move the DAG-run to a terminal status with the failure reason, then
2. `DequeueByDAGRunID`, so the slot is released and the error is reported once.

Retriable startup errors should keep their current behaviour.

## Environment

- Dagu version: built from `main` at `2166de4ba`
- OS: macOS 15 (Darwin 25.5.0)
- Go version: 1.26
- Installation method: `go run ./cmd start-all`

yohamta0:
## Summary

- mark queued DAG runs as failed when a permanent local launch or startup error proves the subprocess never started
- persist the failure reason and finish time, then delete the failed queue item by its exact item ID
- preserve retries published concurrently with failed-item cleanup
- keep distributed worker availability, selector mismatch, transient startup-observation, and liveness failures queued for later recovery
- avoid duplicate scheduler error logs for the same local pre-start failure

## Root cause

The local startup retry path recorded a queued condition after a permanent pre-start failure but left the queue item intact. Because each scheduler tick created a fresh retry budget, the same poison item was selected and failed indefinitely.

The fix uses a compare-and-swap from the exact queued attempt before deleting the selected queue item. Exact-ID deletion prevents cleanup from removing a new retry published after the failed-status transition. Distributed dispatch failures remain queued because a compatible worker can register later.

## Impact

Permanently unrunnable local queued runs now reach a terminal `failed` state once and release their queue slot. Concurrently published retries and recoverable coordinator, worker-selection, liveness, and status-store errors retain their queued recovery behavior.

## Validation

- `make fmt`
- `make bin`
- `gotestsum --format pkgname -- ./internal/service/scheduler -count=1` (299 tests)
- `go test ./internal/intg/distr -run 'TestExecution_QueuedDispatch_RecoversWhen.*WorkerRegistersLater' -count=1`

Fixes #2436

### DIFF
--- a/internal/service/scheduler/queue_conditions_external_test.go
+++ b/internal/service/scheduler/queue_conditions_external_test.go
@@ -427,6 +427,65 @@ func TestQueueProcessorFinalizesLaunchFailure(t *testing.T) {
 	require.Empty(t, items)
 }
 
+func TestQueueProcessorPreservesRetryPublishedDuringFailureCleanup(t *testing.T) {
+	t.Parallel()
+
+	var hookedQueueStore *queueConditionQueueStore
+	f := newQueueConditionFixtureWithConfig(
+		t,
+		config.ExecutionModeLocal,
+		nil,
+		&queueConditionDispatcher{},
+		queueConditionFixtureConfig{
+			executable: filepath.Join(t.TempDir(), "missing-dagu"),
+			procStore: func(base exec.ProcStore) exec.ProcStore {
+				return &queueConditionProcStore{
+					ProcStore:       base,
+					isRunAliveDelay: 50 * time.Millisecond,
+				}
+			},
+			queueStore: func(base exec.QueueStore) exec.QueueStore {
+				hookedQueueStore = &queueConditionQueueStore{QueueStore: base}
+				return hookedQueueStore
+			},
+		},
+	)
+	f.enqueueRun("waiting-run", nil)
+	items, err := f.queueStore.List(f.ctx, f.dag.Name)
+	require.NoError(t, err)
+	require.Len(t, items, 1)
+	originalItemID := items[0].ID()
+
+	runRef := exec.NewDAGRunRef(f.dag.Name, "waiting-run")
+	hookedQueueStore.beforeDelete = func(ctx context.Context) error {
+		attempt, err := f.dagRunStore.FindAttempt(ctx, runRef)
+		if err != nil {
+			return err
+		}
+		status, err := attempt.ReadStatus(ctx)
+		if err != nil {
+			return err
+		}
+		queued, err := exec.EnqueueRetry(ctx, f.dagRunStore, f.queueStore, f.dag, status, exec.EnqueueRetryOptions{})
+		if err != nil {
+			return err
+		}
+		if !queued {
+			return errors.New("retry was not queued")
+		}
+		return nil
+	}
+
+	f.processor.ProcessQueueItems(f.ctx, f.dag.Name)
+
+	status := f.readStatus("waiting-run")
+	require.Equal(t, core.Queued, status.Status)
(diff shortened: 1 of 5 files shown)


yohamta0:
Addressed the CodeRabbit nitpick in dc31f076: `TestQueueProcessorFinalizesLaunchFailure` now starts with a non-empty queued condition and verifies terminal normalization clears it.

The same commit also fixes the CI regression by limiting terminal finalization to proven local pre-start failures. Distributed selector/worker-availability failures remain queued so a matching worker can register later.

PR #2485, opened 2026-08-03
misc: Optimize docker configs and build

dodyirawan85:
## Summary

Optimize Docker build caching and remove unused duplicate deployment Dockerfiles.

## Changes

- Cache UI dependencies before copying the remaining UI sources
- Cache Go modules before copying the remaining repository sources
- Remove redundant node_modules cleanup from the Docker builds
- Keep Dockerfile.dev and Dockerfile.alpine at the repository root as the canonical image recipes
- Update deploy/docker/README.md to reference the root Dockerfiles

## Testing

- go test ./internal/packaging


Dockerfile.alpine
```diff
@@ -7,10 +7,10 @@ RUN rm -f /usr/local/bin/yarn /usr/local/bin/yarnpkg && \
     npm install -g corepack@latest && corepack enable
 
 WORKDIR /app
+COPY ui/package.json ui/pnpm-lock.yaml ./
```
yohamta0, line 10: The new deploy aliases point to this Dockerfile and `Dockerfile.dev` through symlinks. On checkouts with `core.symlinks=false` (common on Windows), each alias becomes a regular file containing only its target path. I reproduced it on this commit: `go test ./internal/packaging` fails, and Docker cannot parse the files. If these aliases are needed, can we keep regular files and add a sync check instead?

yohamta0 (commented):
One inline portability concern.

dodyirawan85:
> One inline portability concern.

Done, i reverted the symlink change and added sync checks instead

yohamta0:
I checked the repo and these duplicate Dockerfiles are not used by the workflows, Compose files, docs, or website. I will remove them and the sync test, update deploy/docker/README.md to point to the root Dockerfiles, and push the change to this PR.

yohamta0 (approved):
LGTM 🚀🚀🚀  Thank you very much for improving the docker build configs!

PR #2503, opened 2026-08-05
fix: avoid DAG-run list refreshes for progress updates

### LINKED ISSUE #2501: Cockpit causes sustained backend CPU from repeated DAG-run index scans
## Summary

Opening the Overview page's Cockpit tab can drive sustained backend CPU usage when the current day contains a large DAG-run history and a workflow is actively updating.

This is backend filesystem/index work, not React rendering. Leaving Cockpit stops the load, and the browser renderer remains idle while the Dagu process consumes CPU.

The issue reproduces on v2.11.4 at commit `8c073ea97060ce1c4fabcdc60472d9781ab7db06`. The relevant hot-path files are unchanged on current `main` (`0e4f321301e333063515465afd25e60d1183996d`), where the same reproduction also occurs.

## Reproduction

1. Create at least 10,000 successful DAG-run statuses for one DAG on the current day.
2. Add a DAG that runs every minute and persists frequent updates. The profiling fixture used a chain of 47 sequential `sleep 1` steps, producing a roughly 48-second run.
3. Start Dagu and open **Overview → Cockpit**.
4. Keep Cockpit open while the scheduled DAG executes.
5. Observe CPU usage for the Dagu server process.

The 48-second fixture persisted 98 lifecycle/update events:

- 1 `dag.run.running`
- 96 `dag.run.updated`
- 1 `dag.run.succeeded`

## Observed behavior

External one-second process sampling observed approximately five CPU cores at minute boundaries while the current-day index was stale. The process then continued using approximately one core while the chain emitted step updates, returning below one percent after the run completed.

A 108.84-second Go CPU profile collected against the exact v2.11.4 binary contained 151.68 CPU-seconds, or about 1.39 cores on average. Cumulative hot-path costs were:

| Function | Cumulative CPU |
| --- | ---: |
| `dagrun.(*Store).ListStatusesPage` | 88.44% |
| `sse.(*multiplexTopic).fetchPayload` | 88.07% |
| `dagrunindex.TryLoadForDay` | 76.73% |
| `dagrunindex.RebuildForDay` | 48.62% |
| `dagrunindex.findLatestAttempt` | 30.03% |
| `dagrunindex.validateIndex` | 27.12% |
| `dagrunindex.parseStatusFile` | 18.13% |

Separately, `syscall.rawsyscalln` accounted for 84.84% of flat samples.

Within `validateIndex`, 19.66 seconds were attributed to `os.Stat(runDir)` and 21.28 seconds to `os.Stat(statusPath)`.

## Root cause

Cockpit creates five independent live DAG-run queries per date, one for each status bucket. Today and yesterday are both marked live, so the page registers 10 query-scoped DAG-run SSE topics.

Each DAG-run event returned by the invalidation reader calls `WakeTopicType(TopicTypeDAGRuns)`. Each active topic independently refetches its payload through `GetDAGRunsListData → Store.ListStatusesPage`.

For the five current-day topics:

1. Starting a run changes the day-directory entry count, invalidating the existing index.
2. All five queries independently enter `RebuildForDay`; there is no per-day shared snapshot or singleflight.
3. Each rebuild discovers the latest attempt and parses the status for every run in the day.
4. A successfully parsed active run prevents `RebuildForDay` from persisting the rebuilt current-day index.
5. The rebuilt result is returned to that request but is not reusable, so subsequent events repeat the full scan while the run remains active.
6. Even a valid index is checked by calling `os.Stat` on both the run directory and status file for every indexed entry, making a cache hit O(number of runs) per query.

The result is a query-scoped SSE thundering herd against an exhaustive, non-reusable current-day index.

## Why the existing index does not protect this path

The index behaves as a fully validated snapshot rather than a mutation-maintained secondary index:

- A readable active entry prevents the snapshot from being saved.
- Every cache hit validates every indexed run.
- Every invalidation rebuilds the entire day.
- Concurrent query topics do not share validation or rebuild work.

Consequently, its normal and fallback paths are both linear in historical run count, and the fallback remains active for the duration of a running workflow.

## Required correctness contract

Any replacement must preserve the file store's canonical behavior across separate DAG-run writer, server/reader, scheduler, and retention processes.

- `status.jsonl` remains canonical. A derived-index failure must not corrupt canonical state or turn an otherwise successful DAG run into a failure.
- After a canonical mutation returns success, a subsequent list read must observe it. If the secondary update cannot be published, the affected DAG root/day must remain detectably dirty and the list path must use canonical data.
- An interruption between canonical persistence and secondary-index publication must be detected and repaired automatically. It must not leave indefinitely stale list results.
- Mutations concurrent with a full scan or compaction must be included through a fenced generation/watermark and catch-up protocol, or equivalent locking. They must not disappear at the scan/checkpoint boundary.
- Base/checkpoint publication and delta retirement must not create a gap. Replaying a delta after interruption must be idempotent.
- Processes caching an older base snapshot must detect a newer generation.
- Duplicate and out-of-order updates must be safe. An older-attempt update must not replace a newer retry, and a delayed upsert must not resurrect a deleted run.
- Ordering must not depend on wall-clock agreement between processes.
- Both direct deletion and retention deletion must publish equivalent tombstone semantics.
- Shared-file append safety must not be assumed across processes. Unique files plus atomic rename is one viable design, but the implementation may use another protocol that establishes the same invariants.
- Index state is scoped by DAG root and day.

## Suggested direction

The steady-state design should make work proportional to changed runs rather than total historical runs:

- Allow active-run summaries to be represented in reusable index state.
- Record a small durable upsert/tombstone when a run status, attempt, or existence changes.
- Share one immutable DAG-root/day snapshot across all status-filter queries.
- Apply only unapplied changes before filtering and pagination.
- Compact changes into a versioned base checkpoint atomically outside the request hot path.
- Use process-local singleflight for recovery and an appropriately fenced cross-process protocol for rebuild/compaction.
- Fall back to canonical filesystem data whenever completeness cannot be established.
- Reconcile unsupported out-of-band filesystem changes outside the normal request hot path, with a documented reconciliation interval.

A missing, corrupt, partial, or version-incompatible index may require a full rebuild, but recovery work should happen once for a given DAG-root/day state and must not be multiplied by Cockpit topics.

## Related event-delivery observation

The file event reader can return an inbox event in one poll and return the same event again from the committed log in a later poll because deduplication is currently scoped to one read. The collector appends the event to the committed log and then removes the inbox file.

Changing this behavior requires its own crash/replay delivery contract; it is not required to implement the index fix. This issue requires duplicate event delivery to be tolerated or coalesced so it cannot multiply expensive historical scans.

### Correctness

- Existing workspace, label, name, status, time-range, ordering, and cursor-pagination behavior remains unchanged.
- Queued, running, waiting, terminal, retried, and deleted runs remain fresh while workflows are active.
- A successful mutation made by another store/writer process is visible to the next completed list read.
- A retry that adds an attempt to an older run cannot be overwritten by a delayed update for an older attempt.
- Direct and retention deletion cannot be undone by delayed upserts.
- Missing, corrupt, truncated, partially written, duplicate, out-of-order, and old-version secondary-index data falls back to canonical storage without returning incorrect results.
- Fault-injection tests interrupt before and after canonical persistence, secondary mutation publication, base/checkpoint publication, and delta retirement. Recovery happens automatically on the next completed read or within the documented reconciliation interval.

### Concurrency

- Tests use at least two real writer processes and a separate reader/compactor process.
- A mutation occurring during full recovery or compaction is present after the resulting checkpoint and catch-up complete.
- Multiple readers using an old cached base detect and load the new generation.
- Multiple sessions and query topics cannot multiply full historical scans for unchanged DAG-root/day state.

### Scaling

- After warm-up, one active-run update inspects a bounded changed set rather than all historical runs.
- Repeated updates to one active run do not repeatedly parse every historical status in that day.
- Scaling benchmarks at `N` and `10N` instrument full scans, status parses, and filesystem metadata operations; steady-state work scales with the changed set rather than `N`.
- Wall-clock CPU is supporting evidence, not the sole regression assertion.

## Relevant code

- `ui/src/features/cockpit/hooks/useDateKanbanData.ts`
- `ui/src/features/cockpit/components/DateKanbanSection.tsx`
- `internal/service/frontend/sse/dagrun_invalidator.go`
- `internal/service/frontend/sse/multiplex.go`
- `internal/persis/file/dagrun/pagination.go`
- `internal/persis/file/dagrun/dagrunindex/dagrunindex.go`
- `internal/persis/file/eventstore/notifications.go`

yohamta0:
## Summary

- keep same-status progress events refreshing exact DAG-run detail subscriptions
- skip aggregate DAG-run list invalidation for progress-only events
- retain aggregate list refreshes for lifecycle transitions
- cover both paths through the SSE multiplexer

## Root cause

`dag.run.updated` events are emitted to keep active run details current. The invalidator also woke every aggregate `dagruns` topic for each of those events, so Cockpit's status buckets repeatedly rebuilt the same date index while a run progressed without changing lifecycle status.

## Scope

This PR fixes the sustained CPU load reported in #2501 by removing aggregate list invalidation from progress-only updates. Broader index architecture improvements can be tracked separately.

## Testing

- `go test ./internal/service/frontend/sse -count=1`
- `go test -race ./internal/service/frontend/sse -count=3`
- `go test ./internal/service/frontend/... -count=1`
- native and Windows `golangci-lint run --timeout=10m ./...`

Closes #2501

### DIFF
--- a/internal/service/frontend/sse/dagrun_invalidator.go
+++ b/internal/service/frontend/sse/dagrun_invalidator.go
@@ -97,7 +97,9 @@ func wakeTopicsForDAGRunEvent(mux *Multiplexer, event *eventstore.Event) {
 		}
 	}
 
-	mux.WakeTopicType(TopicTypeDAGRuns)
+	if event.Type != eventstore.TypeDAGRunUpdated {
+		mux.WakeTopicType(TopicTypeDAGRuns)
+	}
 	mux.WakeTopicType(TopicTypeQueues)
 	mux.WakeTopicType(TopicTypeDAGsList)
 
--- a/internal/service/frontend/sse/dagrun_invalidator_test.go
+++ b/internal/service/frontend/sse/dagrun_invalidator_test.go
@@ -1,0 +1,91 @@
+// Copyright (C) 2026 Yota Hamada
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package sse
+
+import (
+	"context"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/dagucloud/dagu/v2/internal/core"
+	"github.com/dagucloud/dagu/v2/internal/core/exec"
+	"github.com/dagucloud/dagu/v2/internal/service/eventstore"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDAGRunInvalidatorRefreshesListsOnlyForLifecycleEvents(t *testing.T) {
+	tests := []struct {
+		name              string
+		eventType         eventstore.EventType
+		wantListRefreshes int64
+	}{
+		{
+			name:              "progress update",
+			eventType:         eventstore.TypeDAGRunUpdated,
+			wantListRefreshes: 0,
+		},
+		{
+			name:              "lifecycle update",
+			eventType:         eventstore.TypeDAGRunRunning,
+			wantListRefreshes: 1,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mux := NewMultiplexer(StreamConfig{}, nil)
+			t.Cleanup(mux.Shutdown)
+
+			var detailRefreshes atomic.Int64
+			mux.RegisterFetcher(TopicTypeDAGRun, func(_ context.Context, _ string) (any, error) {
+				detailRefreshes.Add(1)
+				return nil, nil
+			})
+			mux.SetRefreshMode(TopicTypeDAGRun, TopicRefreshModeOnDemand)
+
+			var listRefreshes atomic.Int64
+			mux.RegisterFetcher(TopicTypeDAGRuns, func(_ context.Context, _ string) (any, error) {
+				listRefreshes.Add(1)
+				return nil, nil
+			})
+			mux.SetRefreshMode(TopicTypeDAGRuns, TopicRefreshModeOnDemand)
+
+			result, err := mux.createSession(
+				context.Background(),
+				httptest.NewRecorder(),
+				[]string{"dagrun:test/run-1", "dagruns:status=running"},
+				0,
+			)
+			require.NoError(t, err)
+			require.NotNil(t, result.session)
+			defer mux.removeSession(result.session)
+
+			status := &exec.DAGRunStatus{
+				Name:      "test",
+				DAGRunID:  "run-1",
+				AttemptID: "attempt-1",
+				Status:    core.Running,
+			}
+			event := eventstore.NewDAGRunEvent(
+				eventstore.Source{Service: eventstore.SourceServiceServer},
+				tt.eventType,
+				status,
+				nil,
+			)
+
+			wakeTopicsForDAGRunEvent(mux, event)
+
+			require.Eventually(t, func() bool {
+				return detailRefreshes.Load() == 1 && listRefreshes.Load() == tt.wantListRefreshes
+			}, time.Second, 10*time.Millisecond)
+			if tt.wantListRefreshes == 0 {
+				require.Never(t, func() bool {
+					return listRefreshes.Load() != 0
+				}, 200*time.Millisecond, 20*time.Millisecond)
+			}
+		})
+	}
+}


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

PR #2525, opened 2026-08-07
fix: omit redundant params from parallel child labels

### LINKED ISSUE #2498: feat: Notification templates - expose the steps that failed
# Notification templates: expose the steps that failed

## What is missing

Message templates can render the run as a whole — name, run ID, status, error, link —
but there is no way to render *which parts of the run failed*.

For a workflow with many steps, the notification says the run failed without saying
what failed. Finding that out means opening the UI for every alert.

## What I would like

A token per status, holding the list of steps that ended in that status, so the message
can be composed as needed:

```
{{run.failed_steps}}
{{run.partially_succeeded_steps}}
{{run.aborted_steps}}
{{run.succeeded_steps}}
```

Example template:

```
Run {{run.status}} for {{dag.name}}
Failed: {{run.failed_steps}}
{{run.link}}
```

Rendering as a simple comma or newline separated list of step names would already be
enough. An empty token when no step has that status keeps templates simple.

## One detail that matters

When a step fans out over a list of items, or runs child workflows in parallel, the step
name alone is not useful — every failure reports the same single step name, not which of
the items or child workflows failed.

For those steps the list should identify the individual failing runs, for example by
item value or child workflow name, so a fan-out over many child workflows reports the
specific ones that failed rather than just the name of the step that launched them.

## Why this helps

The list of what failed is the first thing anyone needs when an alert arrives. Having it
in the notification itself makes the message actionable on its own, which matters most
for scheduled runs where nobody is watching the UI.

yohamta0:
## Summary

- render a parallel child DAG's resolved name without its parameter list when that name uniquely identifies the child
- retain parameters when sibling child runs share the same resolved name
- preserve parameter and run-ID fallbacks when a child name is unavailable

## Root cause

Parallel status-detail labels always combined the resolved child DAG name with every parameter. Notification templates then embedded that verbose label even when sibling child names were already distinct.

## Impact

Notifications now produce concise labels such as `run_downstreams[intraday-bnci]` while same-child fan-outs remain distinguishable.

Closes #2498.

## Validation

- `go test ./internal/runtime/... -count=1`
- `go test ./internal/service/notification -count=1`
- `go test -race ./internal/runtime/builtin/dag ./internal/service/notification -count=1`
- golangci-lint on changed files for native and Windows targets

---
## Summary by cubic
Makes parallel child status labels concise by omitting params when the resolved child name is unique. Notifications now show shorter labels like intraday-bnci while still disambiguating same‑name children.

- **Bug Fixes**
  - Use resolved child names to check uniqueness; include params only for duplicates. Tests cover resolved-name cases and missing-name fallbacks.
  - Preserve fallbacks to params-only or run ID when no name is available, and stabilize `chatbridge` notification monitor tests on Windows with a platform-aware Eventually timeout.

<sup>Written for commit 49b4801be0e70170da33d26665d08d030bbcaeda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parallel workflow status labels when multiple runs share the same child name.
  * Labels now include parameter values only when needed to distinguish duplicate runs.
  * Preserved fallback labels using available names, values, or run IDs when details are missing.

* **Tests**
  * Expanded coverage for unique names, duplicate names, and missing child-name scenarios.

### DIFF
--- a/internal/runtime/builtin/dag/parallel.go
+++ b/internal/runtime/builtin/dag/parallel.go
@@ -310,6 +310,14 @@ func (e *parallelExecutor) GetStatusDetails() []exec1.NodeStatusDetail {
 	e.lock.Lock()
 	defer e.lock.Unlock()
 
+	nameCounts := make(map[string]int, len(e.runParamsList))
+	for _, params := range e.runParamsList {
+		name := parallelRunName(params, e.results[params.RunID])
+		if name != "" {
+			nameCounts[name]++
+		}
+	}
+
 	details := make([]exec1.NodeStatusDetail, 0, len(e.runParamsList))
 	for _, params := range e.runParamsList {
 		result := e.results[params.RunID]
@@ -320,7 +328,7 @@ func (e *parallelExecutor) GetStatusDetails() []exec1.NodeStatusDetail {
 			status = parallelNodeStatus(result.Status)
 		}
 		details = append(details, exec1.NodeStatusDetail{
-			Label:  parallelRunLabel(params, result),
+			Label:  parallelRunLabel(params, result, nameCounts[parallelRunName(params, result)] > 1),
 			Status: status,
 		})
 	}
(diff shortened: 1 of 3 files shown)


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai full review

yohamta0:
@coderabbitai review

PR #2557, opened 2026-08-13
fix(docker): decode embedded HostConfig.Resources fields in host config

### LINKED ISSUE #2558: bug(docker): decode embedded HostConfig.Resources fields in host config
I'm building a project with Dagu and found a bug that blocks a user from managing the resources given to containerized Dagu runs. 

A fix is already up as [PR #2557](https://github.com/dagucloud/dagu/pull/2557) (one-line Squash: true + a regression test that fails red against the current decoder: expected: 536870912, actual: 0), made with Fable + Claude Code.

---

## Describe the bug

The docker executor's host: block silently drops every field that lives in container.HostConfig's embedded Resources struct — Memory, CPUShares, NanoCPUs, PidsLimit, Devices, and the other cgroup fields — while sibling direct fields (NetworkMode, SecurityOpt, Binds) decode fine. The flat shape shown in the docs (host: {Memory: 536870912}) decodes to zero values with no warning anywhere, so containers run without their configured resource limits.

Root cause: LoadConfigFromMapWithWorkDir (internal/runtime/builtin/docker/config.go) decodes host: with mapstructure but without DecoderConfig.Squash: true. Moby's container.HostConfig embeds container.Resources, so mapstructure only fills those fields from a nested host: {resources: {...}} key — which is undocumented, but works (verified live; that asymmetry pinpoints the missing squash).

## To Reproduce

1. Create a DAG with the configuration below (it's the docs' own resource-limits example, plus a long-running command so the container can be inspected).
2. Run the workflow.
3. While the step runs: docker inspect <container> --format 'Memory={{.HostConfig.Memory}} CPUShares={{.HostConfig.CpuShares}} PidsLimit={{.HostConfig.PidsLimit}}'

## Expected behavior

The created container carries the configured limits: Memory=536870912 CPUShares=512 PidsLimit=128.

## Actual behavior

Memory=0 CPUShares=0 PidsLimit=<nil>

No error, no log warning — the limits are silently absent. Same result for Devices (Devices=[] for a /dev/fuse mapping). Meanwhile direct HostConfig fields in the same block (NetworkMode, SecurityOpt: ["seccomp=unconfined"], Binds with :ro) all land correctly, and the undocumented nested form host: {resources: {Memory: 536870912, PidsLimit: 128}} fills everything — confirming the embedded-struct decode is the issue.

## Environment

- Dagu version: 2.13.0 (ghcr.io/dagucloud/dagu:2.13.0, digest sha256:8f69fcff…); also reproduced against source at the v2.13.0 tag and current main (decoder unchanged)
- OS: Linux x86_64 (kernel 6.6, Docker Engine; dagu in the official container with the host socket)
- Go version: n/a for the container repro; source repro built with Go 1.26
- Installation method: Docker (official image)

## DAG configuration

steps:
  - id: with_resource_limits
    action: docker.run
    with:
      image: alpine:3
      auto_remove: false
      host:
        Memory: 536870912   # per docs example — decodes to 0
        CPUShares: 512      # decodes to 0
        PidsLimit: 128      # decodes to nil
      container:
        Entrypoint: ["sleep", "60"]

fidecastro:
## The bug

The docker executor's `host:` block silently drops every field that lives in `container.HostConfig`'s **embedded `Resources` struct** — `Memory`, `CPUShares`, `NanoCPUs`, `PidsLimit`, `Devices`, and the rest of the cgroup fields — while sibling *direct* fields (`NetworkMode`, `SecurityOpt`, `Binds`) decode fine.

The docs' own example does not work today:

```yaml
steps:
  - id: with_resource_limits
    action: docker.run
    with:
      image: alpine:3
      host:
        Memory: 536870912   # <- decodes to 0
        CPUShares: 512      # <- decodes to 0
```

Measured on **v2.13.0** against a real daemon (`docker inspect` on the created container): `Memory=0`, `CPUShares=0`, `PidsLimit=nil`, `Devices=[]` — no warning anywhere, so containers run **without their configured resource limits**.

## Root cause

`LoadConfigFromMapWithWorkDir` decodes with mapstructure but without `DecoderConfig.Squash`. mapstructure therefore treats the embedded `Resources` struct as a *nested* field: only the (undocumented) `host: {resources: {Memory: ...}}` shape reaches those fields — verified live on v2.13.0, that nested form lands everything, which pinpoints the missing squash.

## Fix

`LoadConfigFromMapWithWorkDir` now decodes the map twice. The first pass keeps the previously working nested `host.resources` shape. The second pass enables `Squash`, which applies Docker's flat JSON shape and gives flat fields precedence when both forms set the same field.

Regression coverage is consolidated in the existing `LoadConfigFromMap` table. It covers the embedded fields, including `NanoCPUs`, the `Devices` object array, and the `*int64 PidsLimit`; direct host fields; the legacy nested form; and mixed-shape precedence.

Validation:

```text
make test TEST_TARGET=./internal/runtime/builtin/docker/...
make check
```

**Behavior note:** Both the documented flat form and the legacy nested `resources:` form remain accepted. Flat fields win when both forms set the same option.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Docker executor configuration parsing so flat host resource settings are correctly applied.
  * Memory, CPU, PID limits, devices, and other host settings now load as expected.

* **Tests**
  * Added coverage to verify decoding of embedded Docker host resource configuration.

Closes #2558

### DIFF
--- a/internal/runtime/builtin/docker/config.go
+++ b/internal/runtime/builtin/docker/config.go
@@ -92,6 +92,14 @@ func LoadConfigFromMapWithWorkDir(workDir string, data map[string]any, registryA
 	md, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
 		Result:           &ret,
 		WeaklyTypedInput: true,
+		// Squash embedded structs so the documented flat `host:` shape
+		// works: container.HostConfig EMBEDS container.Resources, which
+		// holds Memory, CPUShares, NanoCPUs, PidsLimit, Devices and every
+		// other cgroup field. Without Squash, mapstructure fills them only
+		// from a nested `host: {resources: {...}}` key — the flat form the
+		// docs show (`host: {Memory: 536870912}`) decoded to zero values
+		// silently, so containers ran without their resource limits.
+		Squash: true,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to create decoder: %w", err)
--- a/internal/runtime/builtin/docker/config.go
+++ b/internal/runtime/builtin/docker/config.go
@@ -89,23 +89,19 @@ func LoadConfigFromMapWithWorkDir(workDir string, data map[string]any, registryA
 		Shell      []string `mapstructure:"shell"`
 	}{}
 
-	md, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
-		Result:           &ret,
-		WeaklyTypedInput: true,
-		// Squash embedded structs so the documented flat `host:` shape
-		// works: container.HostConfig EMBEDS container.Resources, which
-		// holds Memory, CPUShares, NanoCPUs, PidsLimit, Devices and every
-		// other cgroup field. Without Squash, mapstructure fills them only
-		// from a nested `host: {resources: {...}}` key — the flat form the
-		// docs show (`host: {Memory: 536870912}`) decoded to zero values
-		// silently, so containers ran without their resource limits.
-		Squash: true,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create decoder: %w", err)
-	}
-	if err := md.Decode(data); err != nil {
-		return nil, fmt.Errorf("failed to decode config: %w", err)
+	// Decode legacy nested Resources first, then overlay Docker's flat JSON shape.
+	for _, squash := range []bool{false, true} {
+		md, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
+			Result:           &ret,
+			WeaklyTypedInput: true,
+			Squash:           squash,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create decoder: %w", err)
+		}
+		if err := md.Decode(data); err != nil {
+			return nil, fmt.Errorf("failed to decode config: %w", err)
+		}
 	}
 
 	var autoRemove bool
--- a/internal/runtime/builtin/docker/config_test.go
+++ b/internal/runtime/builtin/docker/config_test.go
@@ -154,3 +154,38 @@ func TestApplyResourceLimits(t *testing.T) {
 	require.Equal(t, int64(1_500_000_000), cfg.Host.Resources.NanoCPUs)
 	require.Equal(t, int64(512*1024*1024), cfg.Host.Resources.Memory)
 }
+
+func TestLoadConfigFromMap_HostEmbeddedResources(t *testing.T) {
+	// container.HostConfig EMBEDS container.Resources (Memory, CPUShares,
+	// NanoCPUs, PidsLimit, Devices, …). The docs show the flat form —
+	// host: {Memory: 536870912} — which requires DecoderConfig.Squash:
+	// without it, every Resources field silently decoded to its zero value
+	// while sibling direct fields (NetworkMode, SecurityOpt) worked, so
+	// containers ran WITHOUT their configured resource limits.
+	cfg, err := LoadConfigFromMap(map[string]any{
+		"image": "alpine",
+		"host": map[string]any{
+			"NetworkMode": "bridge",
+			"SecurityOpt": []string{"seccomp=unconfined"},
+			"Memory":      536870912,
+			"CPUShares":   512,
+			"PidsLimit":   128,
+			"Devices": []map[string]any{{
+				"PathOnHost":        "/dev/fuse",
+				"PathInContainer":   "/dev/fuse",
+				"CgroupPermissions": "rwm",
+			}},
+		},
+	}, nil)
+	require.NoError(t, err)
+	require.NotNil(t, cfg.Host)
+	require.Equal(t, int64(536870912), cfg.Host.Memory)
+	require.Equal(t, int64(512), cfg.Host.CPUShares)
+	require.NotNil(t, cfg.Host.PidsLimit)
+	require.Equal(t, int64(128), *cfg.Host.PidsLimit)
+	require.Len(t, cfg.Host.Devices, 1)
+	require.Equal(t, "/dev/fuse", cfg.Host.Devices[0].PathOnHost)
+	// direct HostConfig fields keep working alongside the squash
+	require.Equal(t, "bridge", string(cfg.Host.NetworkMode))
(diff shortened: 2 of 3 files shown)


internal/runtime/builtin/docker/config.go
```diff
@@ -92,6 +92,14 @@ func LoadConfigFromMapWithWorkDir(workDir string, data map[string]any, registryA
...
 		Result:           &ret,
 		WeaklyTypedInput: true,
+		// Squash embedded structs so the documented flat `host:` shape
+		// works: container.HostConfig EMBEDS container.Resources, which
+		// holds Memory, CPUShares, NanoCPUs, PidsLimit, Devices and every
+		// other cgroup field. Without Squash, mapstructure fills them only
+		// from a nested `host: {resources: {...}}` key — the flat form the
+		// docs show (`host: {Memory: 536870912}`) decoded to zero values
+		// silently, so containers ran without their resource limits.
+		Squash: true,
```
rflpazini: **[suggestion]** `Squash: true` removes `Resources` from the matchable field set, so the only shape that actually applied limits on v2.13.0 — `host: {resources: {Memory: ...}}` — now decodes to zeros with no error (`ErrorUnused` is unset). Anyone who measured `docker inspect` and switched to the nested form to make limits stick will silently lose them after upgrade. The documented flat form is the right contract, but this is the same silent-drop failure mode the PR is fixing, inverted.

**Suggestion:** Lift a nested `resources`/`Resources` map onto the flat `host` keys before decode, with flat keys winning if both are present. Mutating only missing keys avoids clobbering the documented shape:

```go
if host, ok := data["host"].(map[string]any); ok {
    var nested map[string]any
    for _, key := range []string{"resources", "Resources"} {
        if m, ok := host[key].(map[string]any); ok {
            nested = m
            break
        }
    }
    for k, v := range nested {
        if _, exists := host[k]; !exists {
            host[k] = v
        }
    }
}
```

Add a test that `host.resources.Memory` still lands on `cfg.Host.Memory`.

internal/runtime/builtin/docker/config.go
```diff
@@ -92,6 +92,14 @@ func LoadConfigFromMapWithWorkDir(workDir string, data map[string]any, registryA
 	md, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
 		Result:           &ret,
 		WeaklyTypedInput: true,
+		// Squash embedded structs so the documented flat `host:` shape
```
rflpazini: **[suggestion]** The seven-line comment restates the bug history (which fields live on `Resources`, that the nested form used to work, that zeros were silent). That belongs in the PR body, not next to the decoder.

**Suggestion:** Keep a one-line why:

```go
// HostConfig embeds Resources; squash so flat host.Memory matches Docker JSON.
Squash: true,
```

internal/runtime/builtin/docker/config_test.go
```diff
@@ -154,3 +154,38 @@ func TestApplyResourceLimits(t *testing.T) {
 	require.Equal(t, int64(1_500_000_000), cfg.Host.Resources.NanoCPUs)
 	require.Equal(t, int64(512*1024*1024), cfg.Host.Resources.Memory)
 }
+
+func TestLoadConfigFromMap_HostEmbeddedResources(t *testing.T) {
+	// container.HostConfig EMBEDS container.Resources (Memory, CPUShares,
```
rflpazini: **[suggestion]** The test comment repeats the same design narrative already in the function name and assertions (`Memory`/`CPUShares` vs `NetworkMode`/`SecurityOpt`).

**Suggestion:** Drop the block comment. The test name `TestLoadConfigFromMap_HostEmbeddedResources` plus the assertions already document the contract.

internal/runtime/builtin/docker/config_test.go
```diff
@@ -154,3 +154,38 @@ func TestApplyResourceLimits(t *testing.T) {
...
+				"CgroupPermissions": "rwm",
+			}},
+		},
+	}, nil)
+	require.NoError(t, err)
+	require.NotNil(t, cfg.Host)
+	require.Equal(t, int64(536870912), cfg.Host.Memory)
+	require.Equal(t, int64(512), cfg.Host.CPUShares)
+	require.NotNil(t, cfg.Host.PidsLimit)
+	require.Equal(t, int64(128), *cfg.Host.PidsLimit)
```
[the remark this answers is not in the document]
yohamta0: Added `NanoCPUs` to the consolidated flat-resource case in 25493a380, alongside `Memory`, `CPUShares`, `PidsLimit`, and `Devices`.

internal/runtime/builtin/docker/config.go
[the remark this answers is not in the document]
yohamta0: Fixed in 25493a380. The decoder now runs the legacy nested pass first and the squashed flat pass second, so both shapes work and flat fields win when both set the same option. Tests cover nested input and mixed-shape precedence.

yohamta0: Fixed in 25493a380. The decoder now runs the legacy nested pass first and the squashed flat pass second, so both shapes work and flat fields win when both set the same option. Tests cover nested input and mixed-shape precedence.

internal/runtime/builtin/docker/config.go
yohamta0: Fixed in 25493a380. The bug-history block is now a single line that states the decode order and precedence.

internal/runtime/builtin/docker/config_test.go
yohamta0: Fixed in 25493a380. The standalone test and narrative comment were removed; the coverage now lives in the existing table-driven `TestLoadConfigFromMap`.

yohamta0:
@rflpazini Thanks for catching the compatibility regression. The decoder now reads the nested form first, then applies the flat fields, so existing `host.resources` configs keep working. The existing table test covers the nested form and flat-field precedence. It also checks `NanoCPUs`. I removed the extra test comments too.

rflpazini:
@yohamta0 happy to help!

yohamta0 (approved):
LGTM, thank you very much for catching the bug and making the fix! 🚀

PR #2564, opened 2026-08-14
refactor: separate DAG-run work directories from history

yohamta0:
## Summary

- store new file-backed DAG-run work directories outside DAG-run history using a shallow, bounded layout
- keep existing nested work directories usable through lazy fallback and remove separate work trees during run deletion and retention
- add the default path, configuration, schema, Helm, README, and behavior-focused test coverage

## Why

DAG-run history and mutable execution work have different storage lifecycles. Separating them makes each root independently configurable while keeping retry behavior and explicit `working_dir` behavior unchanged. The shallow hashed layout also avoids embedding timestamps and long run IDs in working-directory paths.

## User impact

- new local runs use `{paths.data_dir}/dag-run-work` by default
- `paths.dag_run_work_dir` or `DAGU_DAG_RUN_WORK_DIR` can override the root
- existing runs reuse legacy nested work directories when present
- workflows using `DAG_RUN_WORK_DIR` continue unchanged; workflows hardcoding the old physical path must use the environment variable instead

## Validation

- `make check`
- `make test`
- `go test ./internal/cmn/config ./internal/cmn/schema ./internal/persis ./internal/persis/file/dagrun ./internal/test`
- persistence and common-package import boundary checks
- documentation build and social metadata validation in the separate docs repository

---
## Summary by cubic
Separates per-run work directories from DAG-run history and hardens path isolation. Previously work lived under each run; now it lives under a separate hashed root with lazy fallback to legacy locations. Deleting or retaining runs removes both new and legacy work dirs, including sub-DAGs.

- Default root is `{paths.data_dir}/dag-run-work`; configure via `paths.dag_run_work_dir` or `DAGU_DAG_RUN_WORK_DIR`. `dagu config` now prints this path.
- Layout: `<work-root>/<dag-dir>/<root-run-hash>/{root|<child-run-hash>}`. DAG names are sanitized and run IDs use base32-encoded hashes; retries reuse the same directory and checks prevent path traversal or non-directory collisions.
- Required migration: when using Helm, use `/data/dag-run-work` (or set `paths.dag_run_work_dir`). Processes that share runs must use the same work root. Workflow code must read the runtime `DAG_RUN_WORK_DIR` for its assigned directory. Backups that target subdirectories must include both `paths.dag_runs_dir` and `paths.dag_run_work_dir`. If using a shared durable work root, drain/stop old processes and upgrade together to avoid mixed versions writing to different paths.
- Legacy compatibility: if a legacy nested work dir exists under a run, the store uses it; otherwise it creates the new location.
- Helm: chart bumped to `2.0.2`; ConfigMap sets `paths.dag_run_work_dir: /data/dag-run-work`.
- Tests/docs: schema and environment bindings added; repository and workdir tests updated; incident monitor test stabilized.

<sup>Written for commit d31aeac840f7c87b57c45f98f976fcd59903e518. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2564?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable DAG-run working directories via `paths.dag_run_work_dir` and `DAGU_DAG_RUN_WORK_DIR`.
  * Added a default work directory under the configured data directory.
  * Configuration output now displays the resolved work directory.
* **Improvements**
  * Improved work-file isolation, cleanup, path safety, and compatibility with existing layouts.
  * Updated Helm deployments to use `/data/dag-run-work`.
* **Documentation**
  * Documented backup requirements, shared work-root configuration, and upgrade considerations.


README.md
```diff
@@ -926,8 +926,18 @@ The table lists the most common commands. The binary ships 31 in total, includin
...
 | `DAGU_BASE_CONFIG` | — | Shared base configuration applied to all DAGs |
 
+Set the per-run work root in `config.yaml`, or use the corresponding environment variable above:
+
+```yaml
+paths:
+  dag_run_work_dir: /mnt/dagu/dag-run-work
+```
+
+Processes sharing DAG runs must use the same work root.
```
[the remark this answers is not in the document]
yohamta0: Fixed in d95a8ee1eb. The README now distinguishes the Dagu process configuration root from the runtime DAG_RUN_WORK_DIR value and tells workflows not to construct work paths under DAG-run history.

yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

internal/persis/file/dagrun/workdir_store.go
```diff
@@ -18,21 +19,41 @@ import (
...
+			return "", fmt.Errorf("inspect legacy work directory %s: %w", legacyDir, err)
+		}
+		if exists {
+			return legacyDir, nil
+		}
+	} else if !errors.Is(err, dagrun.ErrDAGRunIDNotFound) {
 		return "", err
 	}
+
 	if err := fileutil.MkdirAll(dir, 0o750); err != nil {
```
rflpazini, line 57: This is the write-path flip. New binaries create `<work-root>/.../root` here when neither hashed nor legacy exists. Old binaries still only look at `<history-run>/work`.

On a rolling Helm deploy (`deploymentMode: distributed`, shared RWX), a parent can land on a new worker and a child/retry on an old one, and they write different dirs. Preferring legacy when it exists doesn't help if the new process creates the hashed dir first.

Mind not switching writers in this release? Two options:
- Expand/contract: this PR still *creates* legacy, next release creates new
- Or symlink the legacy path to the new dir so old binaries follow

At least worth a hard upgrade note: drain, stop the fleet, upgrade, start. "Same work root" doesn't cover mixed binaries.

README.md
```diff
@@ -926,8 +926,20 @@ The table lists the most common commands. The binary ships 31 in total, includin
...
+Set the per-run work root in `config.yaml`, or use the corresponding environment variable above:
+
+```yaml
+paths:
+  dag_run_work_dir: /mnt/dagu/dag-run-work
+```
+
+`DAGU_DAG_RUN_WORK_DIR` configures this root for Dagu processes. Workflow code should use the runtime `DAG_RUN_WORK_DIR` variable for its assigned per-run directory instead of constructing paths under DAG-run history.
+
+Processes sharing DAG runs must use the same work root.
```
rflpazini, line 941: True, and I'd also call out the backup/upgrade contract here.

Work is no longer inside `dag-runs`. A backup of only `paths.dag_runs_dir` restores history with empty work dirs on retry.

A few specifics that would help:
1. Default Helm: no extra volume — `/data/dag-run-work` already lives on the existing PVC
2. Anyone backing up `dag-runs` alone now needs `dag-run-work` too
3. Mixed-version workers are unsupported unless you keep writers on the legacy path (or symlink)

internal/persis/file/dagrun/workdir_store.go
```diff
@@ -43,22 +64,27 @@ func (*WorkDirStore) Snapshot(context.Context, dagrun.WorkDirRef, string) error
...
 		return fmt.Errorf("remove work directory %s: %w", dir, err)
 	}
 	return nil
 }
 
-func (s *WorkDirStore) workDir(ctx context.Context, ref dagrun.WorkDirRef) (string, error) {
-	root := NewDataRoot(s.baseDir, ref.RootDAGRun.Name)
+func (s *WorkDirStore) workDir(ref dagrun.WorkDirRef) string {
+	rootDir := filepath.Join(s.rootDir, dagDirName(ref.RootDAGRun.Name), runIDHash(ref.RootDAGRun.ID))
+	if ref.DAGRun == ref.RootDAGRun {
```
rflpazini, line 80: This treats "same ID + same name" as root. `normalizeWorkDirRef` copies the root name when IDs match and the child name is missing, so `OpenChildAttempt` (which only sets `{ID: childRunID}`) can come back here as `.../root`.

`TestStoreWritesCurrentDAGRunFileCompatibilityLayout` documents store-level `<hash>/<hash>` for a same-ID child, but the production path doesn't honor that.

An explicit `IsRoot bool` (or always passing the child name) would be safer than inferring from "missing name + same ID". A test through `OpenChildAttempt` would lock it down.

internal/persis/file/dagrun/workdir_store.go
```diff
@@ -80,10 +106,26 @@ func workDirForDAGRunDir(dagRunDir string) string {
...
+}
+
+func runIDHash(runID string) string {
+	sum := sha256.Sum256([]byte(runID))
+	return strings.ToLower(base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(sum[:8]))
+}
+
+func directoryExists(path string) (bool, error) {
+	info, err := fileutil.Stat(path)
+	if errors.Is(err, os.ErrNotExist) {
```
rflpazini, line 127: Children already used `sha256[:8]` + base32. Roots used the full history path (`dag-run_<ts>_<id>`). Roots now share this 8-byte hash.

Collision odds are low at typical volumes, but two root IDs under the same DAG would silently share a work dir. 16 bytes, or a `run.json` sidecar with the real run ID, would make that detectable and easier to debug.

internal/persis/file/dagrun/workdir_store.go
```diff
@@ -43,22 +64,27 @@ func (*WorkDirStore) Snapshot(context.Context, dagrun.WorkDirRef, string) error
...
-func (s *WorkDirStore) Remove(ctx context.Context, ref dagrun.WorkDirRef) error {
-	dir, err := s.workDir(ctx, ref)
-	if err != nil {
-		if errors.Is(err, dagrun.ErrDAGRunIDNotFound) {
-			return nil
-		}
-		return err
+func (s *WorkDirStore) Remove(_ context.Context, ref dagrun.WorkDirRef) error {
+	dir := s.workDir(ref)
+	if ref.DAGRun == ref.RootDAGRun {
```
rflpazini, line 69: `filepath.Dir(<...>/root)` assumes the last component is always `root`. That's true today, but a layout tweak would delete the wrong tree or leave children behind.

A small `runTree(ref)` helper would make the intent explicit. Also worth deleting the empty `<work-root>/<dag>/` leftover after the last run.

internal/persis/file/dagrun/workdir_store.go
```diff
@@ -18,21 +19,41 @@ import (
...
+	if err == nil {
+		exists, err = directoryExists(legacyDir)
+		if err != nil {
+			return "", fmt.Errorf("inspect legacy work directory %s: %w", legacyDir, err)
+		}
+		if exists {
+			return legacyDir, nil
+		}
+	} else if !errors.Is(err, dagrun.ErrDAGRunIDNotFound) {
 		return "", err
```
rflpazini, line 54: First materialize of a new run still walks history (`FindByDAGRunID`) just to confirm `<run>/work` is missing. Same cost as before, so not a regression — but the new layout doesn't need the history path to *create* a dir.

Also, any non-`ErrDAGRunIDNotFound` from `FindSubDAGRun` fails Materialize and never creates the new dir. Treating "can't resolve legacy" as a miss (except permission/IO) would be safer. Optionally skip the scan when the caller already knows the history run dir.

internal/persis/file/dagrun/workdir_store.go
yohamta0: Addressed in d31aeac840 and docs commit ec6b268. The README and operational docs now require draining/stopping processes that share a durable work root before this layout transition, then upgrading them together. The Helm docs also clarify that standalone already uses Recreate and that default distributed workers use their existing ephemeral data volumes.

README.md
yohamta0: Addressed in d31aeac840 and docs commit ec6b268. The README now states that selective backups must include both paths and that a complete data-dir backup includes both defaults. The Helm README confirms that /data/dag-run-work uses the existing volume and needs no extra mount.

internal/persis/file/dagrun/workdir_store.go
yohamta0: Fixed in d31aeac840. Root classification is captured before normalization, so a same-ID child with no name is no longer rewritten as the root. The existing compatibility test now exercises this through NewRunStateStore.OpenChildAttempt and verifies that the child directory is returned.

internal/persis/file/dagrun/workdir_store.go
yohamta0: Fixed in d31aeac840. New root and child paths now use 16 SHA-256 bytes (128 bits). Legacy nested child lookup deliberately retains its existing 8-byte encoding so upgrades can still find prior work directories.

internal/persis/file/dagrun/workdir_store.go
yohamta0: The path assumption is removed in d31aeac840 with an explicit runTree helper used by both lookup and root removal. I did not remove the DAG parent directory: concurrent run creation can race an empty-check/removal, while retaining an empty parent is harmless and avoids adding locking solely for cosmetic cleanup.

internal/persis/file/dagrun/workdir_store.go
yohamta0: No code change here after tracing the error cases. The initial scan is the compatibility mechanism for existing runs and is not a regression; the ref does not carry a history path or a safe cutover marker that would let new runs skip it. Current handling already treats ErrDAGRunIDNotFound as a miss while preserving invalid-ID, permission, and I/O failures from FindByDAGRunID/FindSubDAGRun, which should remain actionable errors.

yohamta0:
@coderabbitai review

yohamta0:
Thanks @rflpazini for the thorough review. Catching the same-ID child bug, hash-collision risk, and upgrade/backup implications materially improved this change. I really appreciate the detailed and actionable feedback!

PR #2566, opened 2026-08-14
fix(docker): stop containers and execs on step timeout

rflpazini:
## Summary

`timeout_sec` only cancels the step context. Native `Client.Run` then polled with `context.Background()` and never called `ContainerStop`, so a `sleep 60` with a 2s timeout waited for the sleep. Shared-container `Exec` left the process running because `TerminateOnCancel` defaulted to false.

- After `attachAndWait` returns, honor ctx: stop the container once and join until it is gone.
- Native execs (DAG-level `container:`, multi-command steps, exec-into-existing) now pass `TerminateOnCancel` plus a pid file, matching harness.

## Why

Harness already documented and worked around this hang. The native docker path did not. A cancelled step could leak a running container or an orphan exec inside a shared DAG container.

## Validation

- `go test ./internal/runtime/builtin/docker/ -run 'TestWaitUntilContainerStopped|TestNativeExecOptions'`
- `go test -race` on the same tests
- Live daemon: `alpine sleep 60` with a 2s context finished in **5.63s** (not 60s); container was auto-removed

`cancel.go` line coverage: `waitUntilContainerStopped` 100%, `nativeExecOptions` 100%.

## User impact

Step `timeout_sec` now stops docker `Run` containers. Shared-container steps kill only the exec process, not the whole keepalive container.

---
## Summary by cubic
Stop Docker containers and in‑container execs when a step times out, and bound all post‑cancel cleanup to a single context. Previously `timeout_sec` only canceled the step; `Client.Run` kept waiting and shared‑container execs continued running.

- `Client.Run` now stops the container by ID on cancel using `StopSignal` and `StopGrace` (from step `SignalOnStop` and DAG `MaxCleanUpTime`) and waits with a bounded join (~10s). The stop, inspect, and remove phases share one cleanup context so a stalled daemon cannot hang the run; `auto_remove` behavior is preserved.
- Shared‑container execs terminate on cancel via an in‑container helper marked by `DAGU_EXEC_TOKEN`. It runs as the step user, works with remote daemons and Windows, avoids PID‑file wrapping, and falls back once without the helper if install fails. Cancellation returns the step’s exit status plus any post‑wait errors.
- Fixed exec completion and log follow races by waiting for the attach stream to close before inspect and joining copy goroutines after readers close.
- Helper provisioning adapts to the daemon: bind‑mount locally; copy for remote/in‑container daemons. `Close()` cleans up temp copies.
- Misc: shell resolution uses `sh` via PATH; container cgroup detection is broader. Harness relies on client cancellation and returns exit code 124 on timeout. Build/tests: keepalive assets moved to `internal/runtime/builtin/docker/assets`, `ZIG` is configurable, container tests gated by `DAGU_TEST_DOCKER`, and a frontend test timeout is relaxed to 5s.

<sup>Written for commit ba2c3e520d09ccbd1b6096680f6f09207520ebb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2566?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker task cancellation by stopping running containers promptly when execution is canceled.
  * Ensured canceled processes terminate reliably.
  * Improved handling of stopped, missing, or unavailable containers.
  * Docker execution now reports polling and inspection errors instead of hanging.
  * Added safer fallback naming for unnamed execution steps.


internal/runtime/builtin/docker/cancel.go
```diff
@@ -0,0 +1,63 @@
...
+	if poll <= 0 {
+		poll = defaultPollInterval
+	}
+
+	var stopOnce sync.Once
+	requestStop := func() {
+		stopOnce.Do(func() {
+			if stop != nil {
+				_ = stop()
+			}
```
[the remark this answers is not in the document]
rflpazini: Already addressed in d2d254dd.

`waitUntilContainerStoppedWithGrace` now stores `stopErr = stop()` and returns it immediately (`stop container after cancel: %w`). A no-op stop that leaves the container running is bounded by `maxAfterCancel` (default 10s) so `Client.Run` cannot hang.

Coverage:
- `TestWaitUntilContainerStopped_ReturnsStopError_WhenCancelAndStopFails`
- `TestWaitUntilContainerStopped_Returns_WhenStopIsNoOpAndContainerKeepsRunning`

internal/runtime/builtin/docker/cancel.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func nativeExecOptions(stepName string) ExecOptions {
+	safeStepName := fileutil.SafeName(stepName)
+	if safeStepName == "" {
+		safeStepName = "step"
+	}
+	return ExecOptions{
+		TerminateOnCancel: true,
+		PIDFile:           fmt.Sprintf("/tmp/dagu-exec-%s-%d.pid", safeStepName, time.Now().UnixNano()),
```
yohamta0: This makes `/bin/sh` and a writable `/tmp` mandatory for every native exec. Read-only or distroless `container.exec` targets now fail before the command starts; read-only BusyBox exits 125. Can we keep cancellation from adding those requirements?

yohamta0 (commented):
First pass: one compatibility concern.

internal/runtime/builtin/docker/cancel.go
rflpazini: will try to fix it! ty!

rflpazini: dropped the PID wrap from native execs.

`nativeExecOptions()` is now only `TerminateOnCancel: true`. The step command runs as-is (no `/bin/sh`, no `/tmp`). On Linux, cancel signals the PID from `ExecInspect`. Harness can still set `PIDFile` when it already has a shell.

`TestNativeExecOptions_DoesNotRequireShellOrTmp` locks the unwrapped argv.

internal/runtime/builtin/docker/client.go
```diff
@@ -1096,23 +1098,36 @@ func terminateExecProcess(cli *client.Client, execID string, pidFile string) err
...
+	if inspectResp.PID <= 0 {
+		return fmt.Errorf("exec %s has no pid for targeted cancellation", execID)
 	}
-	if err := signalContainerPIDFileProcess(cli, inspectResp.ContainerID, pidFile, "KILL"); err != nil {
-		return err
-	}
-	if execStoppedWithin(cli, execID, 2*time.Second) {
-		return nil
+	if runtime.GOOS == "linux" {
+		_ = syscall.Kill(inspectResp.PID, syscall.SIGTERM)
```
yohamta0: `ExecInspect.PID` belongs to the Docker host, so `syscall.Kill` only works when Dagu shares the daemon's PID namespace. With remote Docker or containerized Dagu, it can signal an unrelated local process and leave the exec running. I reproduced the leak on Docker Desktop, and this also breaks Windows builds because `syscall.Kill` is undefined. Can cancellation stay on the daemon side?

internal/runtime/builtin/docker/cancel.go
```diff
@@ -0,0 +1,102 @@
...
+			stopErr = stop()
+		})
+	}
+
+	for {
+		if err := ctx.Err(); err != nil {
+			if canceledAt.IsZero() {
+				canceledAt = time.Now()
+			}
+			requestStop()
```
yohamta0: `maxAfterCancel` doesn't cover `requestStop` or `inspect`; the deadline is checked only after they return. Since both Docker calls use `context.Background()`, a stalled daemon can still hang `Client.Run` forever. Can they share a cleanup context with this deadline?

yohamta0 (commented):
Second pass: I found two remaining cancellation issues.

internal/runtime/builtin/docker/client.go
rflpazini: Addressed in cf04b62e.

Dropped local `syscall.Kill` on `ExecInspect.PID`. Cancelable native execs now get a `DAGU_EXEC_TOKEN` env marker (argv unchanged — still no `/bin/sh` or `/tmp` wrap). On cancel we signal that process from inside the container via a daemon-side `docker exec` helper (same pattern as the harness PID-file path).

Verified on Docker Desktop: `sleep 60` is gone after timeout while the keepalive container stays running. Windows package also builds again (`syscall.Kill` is gone).

internal/runtime/builtin/docker/cancel.go
rflpazini: Addressed in cf04b62e.

After cancel, `waitUntilContainerStoppedWithGrace` creates a cleanup context with `maxAfterCancel` and passes it to both `stop` and `inspect`. `stopContainerByID` / `ContainerStop` now take that context too, so a stalled daemon cannot hang past the cleanup deadline.

Coverage: `TestWaitUntilContainerStopped_CancelsStopAndInspect_WhenTheyBlock` blocks in stop+inspect until the cleanup ctx fires.

yohamta0:
Thanks for this PR! I found a few Docker cleanup edge cases during final review and added follow-up commits with tests. The core implementation is still yours. I'll push the fixes here shortly.

yohamta0 (approved):
LGTM 🚀 thank you very much for fixing the issue of container step timeout!

PR #2603, opened 2026-08-22
fix(schema): allow output, form and files on action http.request

### LINKED ISSUE #2602: bug: schema validation incorrect for action http.request, property output
## Describe the bug

action `http.request`: property `output` is not allowed according to the JSON schema validation.
But is correct according to the documentation and step executes correct when ignoring the validation
warning.

## To Reproduce

1. Create a DAG similar to this one: https://docs.dagu.sh/step-types/http#file-output-mode
2. In the editor you'll see scribles under `output:`
3. Hovering over `output` shows: `Property output is not allowed....`

## Expected behavior

`output` should be validated as accepted property for action `http.request`

## Actual behavior

Validation fails on `output` property for action `http.request`

## Environment

- Dagu version: 2.14.0
- OS: Ubuntu 24.04
- Go version: N/A
- Installation method: Helm chart, 5 nodes, 5 worker nodes

## DAG configuration

N/A

## Screenshots

Image

## Additional context

Although schema validation is failing, ignoring the warning shows it still works fine as per documentation.

yzxcj797:
Fixes #2602.

## Summary

The HTTP executor accepts `output` (file output mode), `form` and `files` (multipart upload) — all three documented `with` fields of `action: http.request` (llms.txt / docs.dagu.sh, and the runtime struct in `internal/runtime/builtin/http/http.go`) — but the schema's `httpExecutorConfig` listed none of them while setting `additionalProperties: false`. The editor squiggle reported *Property output is not allowed* and schema validation rejected valid DAGs even though execution worked.

The three properties are added to `httpExecutorConfig`:

- `output` — string, path to write the response body to;
- `form` — map of string→string, multipart form fields;
- `files` — map of string→string, multipart upload field→local path.

## Testing

Added `TestDAGSchemaHTTPRequestWithFields` to the schema test suite: a step combining `form` + `files` + `output` (the multipart upload example from the docs, plus file output mode) and a minimal `output`-only step both validate.

- With the fix: the new tests pass.
- Differential (schema reverted, tests kept): both fail validation — the reported behavior.
- Note: `TestDAGSchemaRepoCopyMatchesEmbeddedSchema` fails on this Windows checkout both with and without the change — `schemas/dag.schema.json` is a relative symlink, which git materializes as a plain text file here (environment, unrelated).

---
## Summary by cubic
Allow `output`, `form`, and `files` in `action: http.request` and enforce that `body` cannot be combined with multipart fields, matching executor behavior and docs. Previously, the schema rejected these fields and allowed `body` with multipart (failing at runtime); now valid DAGs pass and invalid combinations fail during validation.

- Add `output` (string path), `form` (map string→string), and `files` (map string→string) to `httpExecutorConfig`.
- Reject `body` when `form` or `files` is present via a schema `if/then` rule.
- Add tests for multipart + `output`, `output`-only, `body`-only, and invalid `body` + `form`.

<sup>Written for commit 7e24c24e23bf6c9643a2ed464e4111ce45f1ed5b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * HTTP requests now support multipart form fields and file uploads.
  * Response bodies can be saved directly to a file.
  * Uploaded file paths are resolved relative to the step’s working directory.
  * Form and file uploads cannot be combined with a request body.

### DIFF
--- a/internal/cmn/schema/dag.schema.json
+++ b/internal/cmn/schema/dag.schema.json
@@ -3558,6 +3558,20 @@
     "httpExecutorConfig": {
       "type": "object",
       "additionalProperties": false,
+      "allOf": [
+        {
+          "$comment": "body cannot be combined with the multipart fields; the executor rejects that combination at runtime.",
+          "if": {
+            "anyOf": [
+              { "required": ["form"] },
+              { "required": ["files"] }
+            ]
+          },
+          "then": {
+            "not": { "required": ["body"] }
+          }
+        }
+      ],
       "properties": {
         "method": {
           "type": "string",
--- a/internal/cmn/schema/dag.schema.json
+++ b/internal/cmn/schema/dag.schema.json
@@ -3586,6 +3586,16 @@
           "type": "string",
           "description": "Request body content. Can be JSON, XML, or plain text depending on Content-Type header."
         },
+        "form": {
+          "type": "object",
+          "additionalProperties": { "type": "string" },
+          "description": "Multipart form fields mapping server field names to values. Cannot be combined with body."
+        },
+        "files": {
+          "type": "object",
+          "additionalProperties": { "type": "string" },
+          "description": "Files to upload as multipart form data, mapping server field names to local paths. Relative paths resolve from the step working directory. Cannot be combined with body."
+        },
         "silent": {
           "type": "boolean",
           "description": "When true, suppresses response headers and status in output on success."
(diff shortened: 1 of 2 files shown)


yzxcj797:
Adopted in 7e24c24: the schema now rejects `body` combined with `form`/`files` via an `if/then` guard, mirroring the executor's runtime check. Two new cases cover it: `RejectHTTPRequestBodyWithForm` (invalid) and `HTTPRequestBodyOnly` (valid). The new tests all pass (`TestDAGSchemaHTTPRequestWithFields`).

yohamta0 (approved):
LGTM 🚀 thank you very much for fixing the schema issue.

PR #2609, opened 2026-08-23
fix: align file dependency workspaces across execution modes

### LINKED ISSUE #2581: feat: step-level file dependencies that Dagu transfers to the worker

When using Dagu with distributed workers, DAG definitions can be executed remotely, but files referenced by a step are not transferred to the worker.

For example, a DAG may contain:

```yaml
steps:
  - name: Backup
    run: |
      & ./scripts/install.ps1
      node ./scripts/backup.mjs
    shell: powershell
```

If the selected worker does not already have those files in its local filesystem, the step fails.

This becomes especially problematic when the scripts are implementation details of a specific DAG. They logically belong next to the DAG, but currently there is no simple way to declare that the step depends on those files and have Dagu send them to the worker.

I would like steps to support declaring file dependencies that are automatically transferred to the worker together with the step/DAG execution.

For example:

```yaml
steps:
  - name: Backup
    run: |
      & ./scripts/install.ps1
      node ./scripts/backup.mjs
    shell: powershell

    dependencies:
      - ./scripts/install.ps1
      - ./scripts/backup.mjs
```

When the step is dispatched to a worker, Dagu would:

1. Resolve the dependencies relative to the DAG file.
2. Transfer those files to the worker.
3. Preserve their relative directory structure.
4. Make them available from the step's working directory.

Conceptually:

```text
Coordinator:

dags/backup/
├── dag.yaml
└── scripts/
    ├── install.ps1
    └── backup.mjs

        ↓ dispatch

Worker temporary directory:

├── scripts/
│   ├── install.ps1
│   └── backup.mjs
└── ...
```

It would also be useful if directories or glob patterns could eventually be supported:

```yaml
dependencies:
  - ./scripts/**
  - ./config/**
```

But support for individual files alone would already solve the main problem.

### Git checkout inside the DAG

The DAG can clone or pull a repository before executing the actual step.

This works, but adds infrastructure/bootstrap logic to every DAG and requires Git credentials and repository access on workers.

The files may also be very specific to a single DAG, so having the DAG restore its own implementation before every execution feels unnecessary.

### Dagu Actions

Packaged Actions can contain additional files, but they are better suited for reusable functionality.

In this case the scripts are implementation details of a particular DAG, so converting the DAG-specific scripts into reusable Actions introduces another abstraction only to make the files transferable.

### Inline scripts

Scripts can be embedded directly inside the DAG.

This is practical for small shell scripts, but not for complex scripts or programs implemented in PowerShell, Node.js, Python, compiled applications, or projects containing multiple files.

### Shared filesystem

All workers could access the same shared filesystem.

This creates an additional infrastructure dependency and is not ideal for distributed workers running on different machines or operating systems.

### Pre-synchronizing files to every worker

The files could be synchronized independently before Dagu starts.

This works, but requires maintaining a separate file distribution system outside Dagu and ensuring that every worker has the correct version before execution.

## Use case

I use Dagu as a distributed automation system with multiple workers on different machines.

Some DAGs depend on scripts or programs that are specific to that DAG, for example:

```text
backup/
├── dag.yaml
└── scripts/
    ├── prepare.ps1
    ├── backup.mjs
    └── cleanup.ps1
```

These files are part of the DAG implementation and should ideally live together with the DAG.

The coordinator can choose any compatible worker to execute the DAG, so I cannot rely on a particular absolute path or assume that every worker already contains those files.

Being able to declare:

```yaml
dependencies:
  - ./scripts/**
```

would make a DAG self-contained from an execution perspective while still allowing the implementation to remain in normal source files.

This would also make distributed execution behave more naturally: the DAG describes both **what needs to run** and **which local resources are required to run it**.

## Additional context

I am not specifically looking for Git integration here.

The important capability is transferring DAG-local files to the worker selected for execution.

For example, something like:

```yaml
steps:
  - name: Backup
    shell: powershell

    run: |
      & ./scripts/install.ps1
      node ./scripts/backup.mjs

    dependencies:
      - ./scripts/install.ps1
      - ./scripts/backup.mjs
```

could internally result in Dagu packaging those dependencies, sending them along with the execution request, extracting them into the worker's temporary working directory, and then executing the step normally.

This would be especially useful for distributed/shared-nothing worker setups where workers intentionally do not share a filesystem with the coordinator.

yohamta0:
## Summary
- source file dependencies from the DAG working directory and materialize them for local runs
- use the same workspace snapshot semantics for coordinator/worker execution, including dynamically resolved working directories
- add built-in worker OS and architecture labels so the platform selectors reported in the #2581 follow-up can match compatible workers
- update the file dependency contract, schema descriptions, authoring guidance, and conformance fixtures

## Testing
- make fmt
- make test
- go test ./internal/runtimeenv ./internal/runtime/executor ./internal/runtime/agent ./internal/service/coordinator ./internal/intg/distr -count=1
- file dependency regression verified red against the old raw-path behavior and green with the resolver

Closes #2581

## Summary by CodeRabbit

- **New Features**
  - File dependencies now work consistently for local and distributed runs.
  - Dependencies are resolved from the DAG working directory and materialized into the execution workspace.
  - Inline and source-less DAGs support configured working directories.
  - Workers now advertise immutable `os` and `arch` platform labels.

- **Bug Fixes**
  - Improved workspace handling for retries, embedded DAGs, hidden files, and dependencies outside the DAG source directory.

- **Documentation**
  - Updated dependency, working-directory, worker-label, API, and CLI documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -901,7 +901,7 @@ OIDC variables: `DAGU_AUTH_OIDC_CLIENT_ID`, `DAGU_AUTH_OIDC_CLIENT_SECRET`, `DAG
 | `DAGU_WORKER_ID` | — | Worker instance ID |
 | `DAGU_WORKER_MAX_ACTIVE_RUNS` | `100` | Max concurrent runs per worker |
 | `DAGU_WORKER_HEALTH_PORT` | `8092` | Worker health check port |
-| `DAGU_WORKER_LABELS` | — | Worker labels (`key=value,key=value`) |
+| `DAGU_WORKER_LABELS` | — | Worker labels (`key=value,key=value`); `os` and `arch` are built in and cannot be overridden |
 | `DAGU_COORDINATOR_ADVERTISE` | auto-detected hostname | Address advertised in the service registry |
 | `DAGU_WORKER_COORDINATORS` | — | Explicit coordinator addresses for shared-nothing mode |
 
--- a/api/v1/api.yaml
+++ b/api/v1/api.yaml
@@ -13586,7 +13586,7 @@ components:
         dependencies:
           type: array
           minItems: 1
-          description: "DAG-local files, directories, or glob patterns to snapshot for distributed execution"
+          description: "Files, directories, or glob patterns relative to the DAG working directory to snapshot for local or distributed execution"
           items:
             type: string
         call:
--- a/conformance/spec035_file_dependencies/testdata/input.txt
+++ b/conformance/spec035_file_dependencies/testdata/input.txt
@@ -1,0 +1,1 @@
+local dependency
--- a/conformance/spec035_file_dependencies/testdata/local.yaml
+++ b/conformance/spec035_file_dependencies/testdata/local.yaml
@@ -1,13 +1,6 @@
 name: local-file-dependencies
 
 steps:
-  - id: scalar
-    run: test ! -e missing-scalar.txt
-    dependencies: missing-scalar.txt
-
-  - id: array
-    depends: scalar
-    run: test ! -e missing-array.txt
-    dependencies:
-      - missing-array.txt
-      - missing-dir/**
+  - id: consume
+    run: test "$(cat input.txt)" = "local dependency"
+    dependencies: input.txt
--- a/internal/cmn/schema/dag.schema.json
+++ b/internal/cmn/schema/dag.schema.json
@@ -1781,7 +1781,7 @@
               }
             }
           ],
-          "description": "DAG-local files, directories, or glob patterns to snapshot for distributed execution."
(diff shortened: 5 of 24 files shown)


yohamta0:
@coderabbitai review

yohamta0:
Review warning triage:

- The worker os/arch labels are in scope: the linked issue follow-up (https://github.com/dagucloud/dagu/issues/2581#issuecomment-5378357957) reports that worker_selector os:linux fails because no worker advertises the platform label, and the behavior was confirmed as a bug.
- The new exported API, WithSourceWorkDir, has Go documentation. The remaining touched helpers are private and follow the existing repository convention; adding non-contract comments only to raise an automated coverage percentage would add noise.

CodeRabbit reported no actionable comments or inline threads on commit 46c57e81c.

PR #2657, opened 2026-08-29
Fix/named child file dependencies

### LINKED ISSUE #2656: bug: Distributed named child DAGs with file dependencies fail when dispatched from remote workers
## Describe the bug

A DAG running on a remote worker cannot dispatch a separately stored named child DAG when that child declares file `dependencies`.

The worker can fetch the named child's YAML from the coordinator and can dispatch named children that do not have file dependencies. However, the `GetDAG` response contains only YAML, so the worker does not know the child's coordinator-local source file or working directory. The coordinator client then tries to build the dependency bundle on the worker and fails before sending the child dispatch RPC.

Root DAG file dependencies and inline multi-document child dependencies work because those paths already have access to a source workspace or an inherited bundle.

## To Reproduce

1. Run a coordinator/server with `distributed-parent.yaml`, `dependency-child.yaml`, and `scripts/test.sh` in its configured DAG directory.
2. Start a shared-nothing worker that connects to the coordinator but does not have a local copy of the coordinator's DAG directory.
3. Configure distributed execution so `distributed-parent` runs on that worker.
4. Run `distributed-parent`.
5. Observe that the worker successfully fetches `dependency-child` through the coordinator fallback.
6. Observe that dispatching `dependency-child` fails before the coordinator receives its dispatch RPC.

## Expected behavior

The coordinator should prepare a workspace bundle for the named child from the coordinator's authoritative DAG source and dependency files. The selected worker should then download and extract that bundle through the existing workspace-bundle mechanism and successfully execute `scripts/test.sh`.

This should work at every named-child boundary, regardless of DAG nesting depth or which matching worker eventually claims the task.

## Actual behavior

The intermediate worker receives only the named child's YAML. Because it has neither the child's source file nor its source working directory, client-side workspace preparation fails with:

```text
DAG file dependencies require a source file or working directory
```

Representative worker logs:

```text
No local DAG store, trying remote fallback
DAG loaded from remote fallback
Executing sub DAG via injected sub-workflow runner
Dispatching child workflow task
Distributed child workflow dispatch failed
failed to dispatch task: DAG file dependencies require a source file or working directory
```

There is no corresponding `Handler Dispatch called` entry on the coordinator for the failed child. This shows that the failure occurs during client-side bundle preparation before the child dispatch RPC.

Removing `dependencies` from the named child allows the same nested dispatch path to succeed.

## Environment

- Dagu version: `v2.15.3` (`93c42c7f705b7df225f34c457ef1fc1deaf27901`)
- OS: Windows 10 coordinator/server; Ubuntu on WSL2 shared-nothing worker
- Go version: `go1.27.0 windows/amd64` for the source build
- Installation method: binary built from source

## DAG configuration

`distributed-parent.yaml`:

```yaml
name: distributed-parent

worker_selector:
  os: linux
  distro: ubuntu
  environment: wsl

steps:
  - id: run_child
    action: dag.run
    with:
      dag: dependency-child
    worker_selector:
      os: linux
      distro: ubuntu
      environment: wsl
```

`dependency-child.yaml`:

```yaml
name: dependency-child

shell: bash

steps:
  - id: run_script
    run: ./scripts/test.sh
    dependencies:
      - ./scripts/test.sh
```

`scripts/test.sh`:

```sh
#!/usr/bin/env sh
set -eu

echo "named child dependency was transferred"
```

## Screenshots

Image

## Additional context

The workspace-bundle format, coordinator upload/download RPCs, content-addressed bundle store, task descriptor fields, and worker extraction logic already exist and work for root distributed DAGs.

The missing behavior is source-ownership handoff for a remotely fetched named child:

```text
remote worker fetches child YAML
  -> remote worker does not own child files
  -> client attempts to package child dependencies locally
  -> packaging fails before dispatch
```

The host that owns the source workspace should create the bundle. For this topology, that host is the coordinator. Before packaging files, the coordinator should verify that the submitted child YAML exactly matches the authoritative stored DAG, then resolve the effective `working_dir` using the task's base configuration and runtime or retry parameters.

This limitation is currently documented in `specs/035-file-dependencies.md` and `skills/dagu/references/file-dependencies.md`, but it prevents file dependencies from composing with distributed named-child routing added for shared-nothing workers.

Related work:

- #2581 / #2601 / #2609: file dependency workspace bundles
- #2591 / #2594: nested worker-to-worker named DAG routing

Current workarounds are to keep the dependency-bearing child local, define it as an inline multi-document child so it inherits the parent workspace, pre-sync its files to workers, or package the logic as a Dagu Action.

bh-schmidt:
## Summary

Enable separately stored named child DAGs with file dependencies to run across shared-nothing distributed workers.

## Changes

- Defer workspace preparation when a worker has a dependency-bearing DAG definition but no source workspace
- Prepare named child workspace bundles from the coordinator's current stored DAG definition and source files
- Reject dispatch when a remotely fetched DAG definition no longer matches the coordinator's stored definition
- Resolve explicit and retry parameters before packaging dependencies from parameterized working directories
- Exercise nested named DAG dependencies across three distinct shared-nothing workers
- Make distributed integration workers use coordinator DAG fallback instead of direct access to the coordinator's DAG repository
- Document coordinator-owned snapshots and failure behavior for named child dependencies

## Evidences

Before:
image

After:
image

## Related Issues

Closes #2656

## Checklist

---
## Summary by cubic
Fixes dispatched named child DAGs with file dependencies failing in shared-nothing distributed runs when workers can't access the child's source workspace.

Before, a worker with a dependency-bearing DAG definition but no source workspace couldn't prepare the run workspace, so dispatch failed. Now the coordinator packages an authoritative workspace bundle from its stored DAG and source files and attaches it to the dispatched task, and workers defer their own packaging when no source is available. If the fetched definition no longer matches the coordinator's stored DAG, dispatch is rejected and the caller must reload and retry.

**Bug Fixes**
- Coordinator builds and stores named-child workspace bundles at dispatch time.
- Workers skip local workspace preparation when the dependency-bearing definition has no source workspace; a configured `working_dir` alone doesn't count as a source.
- Explicit and retry parameters are resolved before packaging dependencies from parameterized working directories.
- Nested named child DAGs with file dependencies run across multiple shared-nothing workers.
- Documents coordinator-owned snapshots and failure behavior for named child dependencies.

<sup>Written for commit 98131d9454829c2ed5ac823143b410f2f6d602dd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2657?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Distributed tasks can now include named file dependencies, including dependencies in nested child workflows.
  - File dependencies are packaged from the authoritative workflow source before remote execution.
  - Remote workers can add and resolve file dependencies without requiring a local source workspace.

- **Bug Fixes**
  - Improved handling when workflow sources are unavailable during dispatch.
  - Added validation to detect changed workflow definitions before execution and provide actionable retry guidance.
  - Preserved dependency resolution across retries and runtime parameter variations.

- **Documentation**
  - Updated file-dependency guidance and specifications to reflect distributed execution behavior.

### DIFF
--- a/internal/service/coordinator/client.go
+++ b/internal/service/coordinator/client.go
@@ -350,6 +350,10 @@ func (cli *clientImpl) prepareTaskWorkspace(ctx context.Context, task *dispatch.
 	if err != nil {
 		return fmt.Errorf("load DAG for file dependency snapshot: %w", err)
 	}
+	// A configured working_dir does not establish that this host owns the DAG source.
+	if task.SourceFile == "" && task.SourceWorkDir == "" && runtimeexec.HasDAGFileDependencies(dag) {
+		return runtimeexec.ErrDAGWorkspaceSourceUnavailable
+	}
 	desc, archivePath, err := runtimeexec.PrepareDAGWorkspaceFile(ctx, dag, cli.config.WorkspaceBundleDir)
 	if err != nil {
 		return err
--- a/internal/service/coordinator/client.go
+++ b/internal/service/coordinator/client.go
@@ -241,7 +241,9 @@ func (cli *clientImpl) Dispatch(ctx context.Context, req dispatch.DispatchReques
 		return fmt.Errorf("dispatch task is nil")
 	}
 	if err := cli.prepareTaskWorkspace(ctx, task); err != nil {
-		return err
+		if !errors.Is(err, runtimeexec.ErrDAGWorkspaceSourceUnavailable) {
+			return err
+		}
 	}
 	protoTask, err := convert.DispatchTaskToProto(task)
 	if err != nil {
--- a/internal/service/coordinator/handler.go
+++ b/internal/service/coordinator/handler.go
@@ -33,6 +33,7 @@ import (
 	"github.com/dagucloud/dagu/v2/internal/proto/convert"
 	"github.com/dagucloud/dagu/v2/internal/queue"
 	"github.com/dagucloud/dagu/v2/internal/runtime"
+	runtimeexec "github.com/dagucloud/dagu/v2/internal/runtime/executor"
 	"github.com/dagucloud/dagu/v2/internal/runtime/workspacebundle"
 	secretpkg "github.com/dagucloud/dagu/v2/internal/secret"
 	"github.com/dagucloud/dagu/v2/internal/spec"
@@ -162,6 +163,7 @@ type Handler struct {
 	logDir                    string                             // For log storage
 	artifactDir               string                             // For artifact storage
 	stateStore                dagrun.StateStore                  // For persistent DAG state shared across DAG runs
+	workspaceBundleDir        string                             // Root for immutable task workspace bundles and staging
 	workspaceBundleStore      *workspacebundle.Store             // For immutable task workspace bundles
 	dispatchTaskStore         dispatch.DispatchTaskStore         // Shared distributed dispatch queue
 	dispatchAdmissionStore    dispatch.DispatchAdmissionStore    // Shared distributed admission state
@@ -299,6 +301,7 @@ func NewHandler(cfg HandlerConfig) *Handler {
 		logDir:                    cfg.LogDir,
 		artifactDir:               cfg.ArtifactDir,
 		stateStore:                cfg.StateStore,
+		workspaceBundleDir:        cfg.WorkspaceBundleDir,
 		workspaceBundleStore:      bundleStore,
 		dispatchTaskStore:         cfg.DispatchTaskStore,
 		dispatchAdmissionStore:    dispatchAdmissionStore,
@@ -513,6 +516,9 @@ func (h *Handler) Dispatch(ctx context.Context, req *coordinatorv1.DispatchReque
 		if err := h.ensureWaitingWorkerAvailability(req.Task.WorkerSelector, req.Task.TargetWorkerId); err != nil {
 			return nil, status.Error(dispatchErrorCode(err), err.Error())
 		}
+		if err := h.prepareDispatchTaskWorkspace(ctx, req.Task); err != nil {
+			return nil, err
+		}
 
 		var prepared *preparedDispatchAttempt
 		if h.dagRunRepository != nil {
@@ -547,6 +553,10 @@ func (h *Handler) Dispatch(ctx context.Context, req *coordinatorv1.DispatchReque
 	}
 	if !anyWorkerMatches(healthyWorkers, req.Task.WorkerSelector, req.Task.TargetWorkerId) {
 		return nil, status.Error(codes.FailedPrecondition, errNoMatchingWorkers.Error())
+	}
+	if err := h.prepareDispatchTaskWorkspace(ctx, req.Task); err != nil {
+		h.releaseAdmissionToken(ctx, admissionToken)
+		return nil, err
 	}
 
 	prepared, err := h.prepareAttemptForDispatch(ctx, req.Task)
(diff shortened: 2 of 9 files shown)


internal/service/coordinator/handler.go
```diff
@@ -536,6 +546,97 @@ func (h *Handler) Dispatch(ctx context.Context, req *coordinatorv1.DispatchReque
...
+	return nil
+}
+
+func loadDispatchWorkspaceDAG(ctx context.Context, task *coordinatorv1.Task, definition []byte, sourceFile string) (*ir.DAG, error) {
+	loadOpts := []spec.LoadOption{spec.WithName(task.Target)}
+	if task.BaseConfig != "" {
+		loadOpts = append(loadOpts, spec.WithBaseConfigContent([]byte(task.BaseConfig)))
+	}
+	if task.Params != "" {
+		loadOpts = append(loadOpts, spec.WithParams(task.Params))
```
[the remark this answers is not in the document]
bh-schmidt: This PR intentionally packages the workspace on the coordinator because, in the shared-nothing topology from #2656, the coordinator owns the named DAG source and dependency files. Here `working_dir` identifies the source root while packaging; the worker executes from the extracted per-run workspace and does not need the coordinator path.

The security finding depends on a trust policy that does not currently appear to be defined. `working_dir` supports absolute paths and runtime parameters, while nested dispatch sends only the final `task.Params`. The coordinator cannot infer whether that value was authorized by the original user or modified by the dispatching worker. Restricting it to the DAG directory or default parameter would break supported external working directories.

Could a maintainer confirm the intended contract?

1. Are connected workers trusted to submit nested dispatch parameters?
2. Is the coordinator-local `working_dir` intentionally the source root for named-child dependency snapshots?
3. If workers are untrusted, should Dagu introduce explicit coordinator-side allowed source roots, such as `file_dependencies.allowed_source_roots`, rather than infer containment from the DAG directory?

The requested workspace-bundle read authorization affects the existing content-addressed bundle RPC and appears broader than this bug fix. I can track that separately if required. 

@yohamta0 could you please check this?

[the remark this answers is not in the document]
yohamta0: Dagu is designed for trusted environments. Executing arbitrary workflow commands means running untrusted workers is already unsafe. Supporting absolute and parameterized working_dir values is intentional, so coordinator-side allowed roots are out of scope for this PR. If Dagu’s trust model changes in the future, bundle read authorization can be addressed separately.

internal/service/coordinator/client.go
```diff
@@ -241,7 +241,9 @@ func (cli *clientImpl) Dispatch(ctx context.Context, req dispatch.DispatchReques
 		return fmt.Errorf("dispatch task is nil")
 	}
 	if err := cli.prepareTaskWorkspace(ctx, task); err != nil {
-		return err
+		if !errors.Is(err, runtimeexec.ErrDAGWorkspaceSourceUnavailable) {
```
yohamta0, line 244: This fallback only works when `working_dir` is empty. `ErrDAGWorkspaceSourceUnavailable` is returned only when both the source file and working directory are empty. A remotely fetched named DAG with `working_dir: ${params.WORKSPACE}` therefore tries to package that path on the worker. If the path is absent, dispatch fails with `stat workspace root` before reaching the coordinator; if it exists, the worker snapshots its local files. Please defer source-less dependency DAGs to the coordinator regardless of `working_dir`.

bh-schmidt: Thanks, I fixed this adding this validation in the `prepareTaskWorkspace`:

```go
	if task.SourceFile == "" && task.SourceWorkDir == "" && runtimeexec.HasDAGFileDependencies(dag) {
		return runtimeexec.ErrDAGWorkspaceSourceUnavailable
	}
```

Could you please check if that is what you expected?

yohamta0: Yes, this addresses it. The guard now defers source-less dependency DAGs before local packaging, including when working_dir is set. I reran the coordinator unit tests and shared-nothing integration test; both pass. Thanks for the fix.

yohamta0 (approved):
LGTM ! 🚀🚀🚀 Thanks for adding support for file dependencies in nested distributed DAGs!

PR #2658, opened 2026-08-29
test: cover value resolution escaping, coercion, and notice classification

arky:
Add missing test in conformance/spec003_value_resolution 

## Summary

  conformance/spec003_value_resolution had zero black-box coverage despite
  Spec 003 being marked Implemented and being the foundational spec every
  other value-resolution spec depends on. Add coverage for:

  - the odd/even backslash escape rule proven across all four reference
    namespaces (consts, params, env, steps), not just one
  - string-insertion coercion for booleans/integers, plus the "shortest
    round-trippable" number rule (trailing zeros and exponent notation
    both normalize to plain minimal decimal text)
  - the defect-vs-runtime-only unresolved-reference split: an unknown
    const and a step-output reference missing its authored dependency
    are both defects shown by `dagu validate` without --show-unresolved;
    a required param missing at validate time is runtime-only and stays
    hidden until --show-unresolved is passed

  All cases pass against the current binary — this is regression coverage
  for already-correct behavior, not a claim of new functionality.

## Checklist

- [x ] Code follows the project style guidelines
- [ x] Self-review of the code has been performed
- [ x] Tests have been added or updated as needed
- [ x] Changes have been tested locally

## Summary by CodeRabbit

* **Tests**
  * Added conformance coverage for escaping behavior across constants, parameters, environment values, and step outputs.
  * Added validation for boolean, integer, decimal, trailing-zero, and exponent-form value insertion.
  * Added checks for defect and runtime-only notice classification, including missing dependencies, unknown constants, and unresolved parameters.

### DIFF
--- a/conformance/spec003_value_resolution/testdata/escape_matrix_consts.yaml
+++ b/conformance/spec003_value_resolution/testdata/escape_matrix_consts.yaml
@@ -1,0 +1,10 @@
+consts:
+  - name: prod
+working_dir: .
+steps:
+  - id: check
+    run: |
+      printf '%s\n' '${consts.name}' > matrix.txt
+      printf '%s\n' '\${consts.name}' >> matrix.txt
+      printf '%s\n' '\\${consts.name}' >> matrix.txt
+      printf '%s\n' '\\\${consts.name}' >> matrix.txt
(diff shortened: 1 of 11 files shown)


yohamta0 (approved):
LGTM, thank you very much for filling those test gap!

PR #2660, opened 2026-08-29
test: cover step run array-form validation and shell selection

arky:
## Summary
conformance/spec013_step_run had zero black-box coverage despite Spec013 depending specs 014/015 build on. 
  
## Changes
Add green regression coverage for the parts already correctly implemented:

  - malformed array-form `run` entries (multi-key mapping items, nested
    mapping/array values) are rejected at validate time, and the
    single-key scalar mapping form is correctly accepted and reconstructed
  - shell selection precedence (step with.shell > root shell >
    DAGU_DEFAULT_SHELL > platform fallback) holds end to end

Two rules from the spec were investigated and found not implemented as specified — `shell: direct` rejection for `run` steps, and  `with.shell_args` without `with.shell` — and are intentionally left untested here rather than covered with red tests.

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added conformance coverage for valid and invalid array-form `run` entries.
  * Added validation checks for malformed mappings and nested arrays.
  * Added coverage for shell selection precedence across step, root, environment, and platform defaults.
  * Added scenarios confirming valid commands execute and produce expected output.

### DIFF
--- a/conformance/spec013_step_run/step_run_test.go
+++ b/conformance/spec013_step_run/step_run_test.go
@@ -1,0 +1,101 @@
+// Copyright (C) 2026 Yota Hamada
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package spec013_step_run_test
+
+import (
+	"runtime"
+	"testing"
+
+	"github.com/dagucloud/dagu/v2/conformance/harness"
+)
+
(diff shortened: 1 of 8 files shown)


yohamta0 (approved):
LGTM !

PR #2661, opened 2026-08-29
test: cover the agent DAG decision loop end to end

arky:
## Summary
conformance/spec032_agent only proved static YAML shape validation for type: agent DAGs; the actual LLM-driven decision loop had zero black-box coverage.
 
## Changes
Add a fake OpenAI-compatible chat-completions server (driven through the real "local" provider, not a Dagu-internal mock) and six scenarios: the basic action -> set_task_status loop, tool-error tolerance for a bad task name, the 5-run repetition limit, stalling on two consecutive no-tool-call replies, partially-succeeded terminal status, and the full ask_user suspend/resume bridge into human tasks.

The ask_user scenario starts its own scheduler process for the resume half of the flow; give it its own port via DAGU_SCHEDULER_PORT + harness.FreePort instead of the scheduler's hardcoded default (8090), so it can't collide with concurrent scheduler-starting tests elsewhere in the suite.
  
## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added conformance coverage for agent runtime decision-making, including tool execution, errors, repetition limits, stalling, partial failures, and task completion.
  * Added coverage for suspending and resuming workflows when user input is requested.
  * Added realistic local-model test fixtures for successful, partially failed, and interactive agent runs.

### DIFF
--- a/conformance/spec032_agent/fake_llm_test.go
+++ b/conformance/spec032_agent/fake_llm_test.go
@@ -1,0 +1,153 @@
+// Copyright (C) 2026 Yota Hamada
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+package spec032_agent_test
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+)
+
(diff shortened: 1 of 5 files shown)


yohamta0 (approved):
LGTM, thank you for adding tests for agent type DAG!

PR #2673, opened 2026-08-30
fix(api): bound the queued-count scan so deep queues stay servable

### LINKED ISSUE #2670: perf: unbounded per-item status reads in queue count make the queues page time out on large queues
## Summary

`ListQueues` computes each queue's `queuedCount` by walking **every** item in **every** queue and doing two filesystem operations per item, with no cap and no cache. On a queue that is merely large — not stuck — this makes both `GET /queues` and the queues SSE topic time out reliably.

The items list on the same page is already correctly bounded. It is only the count that is not.

## Where

`internal/service/frontend/api/v1/queues.go` (on `main`, ef17333):

- `countVisibleQueuedItems` (line 382) pages with `queueCursorScanBatch = 64` but loops until `!page.HasMore` — there is no limit.
- For **each** item it calls `fetchDAGRunSummary` → `dagRunRepository.FindAttempt` + `attempt.ReadStatus`, i.e. a directory lookup plus a status file read.
- All of that is only to evaluate `summary.Status != api.StatusRunning` (line 399) and increment an `int`.
- It is called from `collectQueues` at lines 268 and 290, so the cost is paid for every queue on every request.

Contrast with `listVisibleQueuedItems` (line 323), which is bounded correctly: it stops at `limit`, returns a cursor, and reads status only for the items it actually returns.

So the item list is genuinely first-page-only, while the count badge next to it forces a full scan of the entire queue.

## Impact

`GetQueuesListData` serves the queues SSE topic through the same `collectQueues` path, so the REST request and the live topic hit the identical scan and fail together — the page does not degrade, it just stops loading.

Observed in production on v2.16.0: distributed mode, ~47 DAGs, several on `* * * * *`. A backlog built up on one queue, and from that point the queues page timed out on every load. The timeout then hides the very backlog causing it, so there is no way to watch it drain from the UI.

Worth stressing that nothing is broken when this happens — the queue drains fine and runs succeed. Queue *depth* alone is enough to make the page unusable.

## Two possible directions

I am happy to send a PR, but the choice here seems yours to make rather than mine:

**A. Bound the scan.** Count up to a cap and report e.g. `500+`. No change to *what* is counted, but the number becomes approximate — and it may want an API field to signal that it was capped, which is a schema change.

**B. Drop the per-item read.** `collectQueues` already has `localRunningIDs` in memory (lines 232-254, from `ListAllAlive`) plus the distributed running set. Testing membership against those removes the I/O entirely.

The caveat on B is that it is not an exact substitution: the current filter reads *status* from each run's file, whereas `localRunningIDs` is *process liveness*. Those can disagree — a run still marked `Running` in its status file whose process is gone, for instance. Whether that difference matters for a count badge is a judgement call about the intended semantics.

Happy to implement whichever you prefer, or a combination (fast path via the known running set, plus a cap as a backstop).

four-bytes-robby:
Fixes #2670.

## Problem

`countVisibleQueuedItems` walked **every** item in **every** queue and called `fetchDAGRunSummary` on each — `FindAttempt` plus `ReadStatus`, two filesystem operations per item — only to evaluate `Status != Running` and increment an `int`. No cap, no cache, so the cost grew with queue depth on every request.

The item list on the same page was already bounded: `listVisibleQueuedItems` stops at `limit`, returns a cursor, and reads status only for the items it returns. It was just the count that scanned without limit.

`GetQueuesListData` serves the queues SSE topic through the same `collectQueues` path, so the REST endpoint and the live topic ran the identical scan and failed together.

Measured on a production instance with a deep queue:

| request | result |
|---|---|
| `GET /api/v1/queues` | times out at 10s |
| `POST /api/v1/events/stream/topics` | returns after 22s |

The queue was healthy throughout — items were dispatching, runs were succeeding. Depth alone was enough to make the page unservable, and because the request never completes, the UI cannot show the backlog that is causing it.

## Scope

This bounds the **status reads**, not the whole request. `QueueList` calls `queueRecordIDs(ctx, "")`, so deriving the set of queue names still walks every queued item path before any budget applies. Queue-name discovery is a follow-up per review discussion, since it belongs in the store rather than the API layer.

What is bounded here is the expensive half: the per-entry `FindAttempt` + `ReadStatus` pair that made counting O(queue depth) in reads and JSON parses.

## Change

Counting stops at `queuedCountScanCap` (500) for a single queue and at `queuedCountRequestScanCap` (2000) across the request, reporting the total as a lower bound so a truncated scan is never presented as exact:

- `Queue.queuedCountCapped` and `QueuesSummary.totalQueuedCapped` are new **optional** booleans, set only when a cap is reached. Clients that ignore them are unaffected.
- The cap counts entries **processed** rather than entries counted, so running or unreadable entries cannot scan without bound.
- A queue whose scan was capped before reaching a visible entry still surfaces instead of returning 404.
- The UI renders a capped count as `500+`, and treats a capped zero as possibly non-empty so the queued-items feed still loads.

Exact counts below the caps are unchanged and set no flag.

## Notes

- `queuedCountScanCap` is a `var` rather than a `const` purely so tests can lower it instead of enqueuing thousands of runs. Glad to change the approach if you prefer a const plus a build-tag hook or an API field.
- Two tests added: one asserting the count stops at the cap and sets the flag, one asserting an exact count below the cap sets no flag.
- The bulk of the diff is the regenerated embedded spec blob in `api.gen.go`; the hand-relevant part there is 12 lines.
- `api.gen.go` was regenerated with oapi-codegen **v2.5.1** to match the version that produced the committed file. Note `go.mod` pins v2.7.2, so `make api` on a clean tree rewrites that file wholesale — that drift predates this PR and I did not want to bundle it in here.
- `make api-validate` currently fails on `main` with `schema "DAG": extra sibling fields: [description]`, unrelated to this change, so I ran the codegen step directly.

## Verification

- `go build ./...`, `go vet`, `gofmt` clean
- `go test ./internal/service/frontend/api/v1/... -race` passes
- `tsc --noEmit` clean for the UI changes

---
## Summary by cubic
Bounds the queued-count scan so deep queues and many-queue deployments stay servable. Counting previously walked every item with two filesystem reads per item, so a deep queue could time out `GET /api/v1/queues` and stall the queues SSE topic. Counting now stops after scanning 500 entries per queue and 2000 entries per request, reporting the total as a lower bound; the UI renders capped counts as `500+` in the queue card and queue detail page. Fixes #2670.

- The cap counts entries scanned, not visible queued entries, so queues full of running or unreadable entries also stop scanning; a queue capped before reaching any visible entry reports a capped count of zero instead of 404.
- New optional `queuedCountCapped` and `totalQueuedCapped` fields are set only when the cap is reached; exact counts below the cap set no flag.
- The caps are `var`s so tests can lower them; added tests cover capped, exact, running-entry, and request-budget caps.
- The bulk of the diff is the regenerated `api.gen.go`; note that `make api` on a clean tree rewrites that file because `go.mod` pins oapi-codegen v2.7.2 while it was generated with v2.5.1 (drift predates this PR).

<sup>Written for commit 0d94f59895e6a494e4eac09a896c333ac38a02b5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Queue counts now indicate when they are lower bounds due to large queue sizes.
  * Queue cards and detail pages display capped counts with a “+” indicator, such as “500+”.
  * Queue summaries include a corresponding indicator for capped total counts.

* **Bug Fixes**
  * Prevented lengthy scans from delaying queue count responses by applying a scan limit.


four-bytes-robby:
Pushed a fix for the first review finding — it was a real gap, thanks.

**Cap now bounds entries scanned, not entries counted.** The cap only advanced on *visible* entries, so a running (or unreadable) entry still cost a `FindAttempt` + `ReadStatus` while never moving the counter. A queue made up of running entries therefore kept scanning without bound — precisely the case this PR exists to prevent. It now tracks entries processed, and reports `capped` only when entries actually remain unscanned, so a queue ending exactly on the cap is still an exact count.

That surfaced a second gap: with `count == 0` and `capped == true`, the `count > 0` guard skipped creating the queue, so a queue holding more running entries than the cap answered **404** instead of reporting a capped count of zero. A truncated scan proves entries exist, so it now surfaces.

Regression test added covering more-than-cap running entries ahead of the first visible queued entry.

---

On the second finding, about `queueRefreshToken` in `ui/src/pages/queues/queue/index.tsx` — I looked and did not change it, because the premise does not quite hold:

- the token is not derived from `queuedCount` alone; it also serializes `running[]` as `name:dagRunId:status`, so it changes whenever a run starts or finishes
- user-driven dequeues bypass the token entirely — the page calls `reload()` explicitly alongside `mutate()`
- the queue object itself polls on a 3s `refreshInterval`

There is a genuine residual edge: a queue pinned at the cap whose membership changes while the running set happens to be unchanged in the polled snapshot would not retrigger the feed. That is newly reachable because the count can now stick at the cap. It looks narrow and self-correcting on the next running change or user action, and closing it properly would mean exposing the queue store's `Revision` (which already changes on membership change) as a new API field — more scope than this PR needs.

Happy to add that here if you would rather have it fixed now, or to split it out.

internal/service/frontend/api/v1/queues.go
```diff
@@ -398,10 +428,13 @@ func (a *API) countVisibleQueuedItems(ctx context.Context, queueName string) (in
 			summary, err := a.fetchDAGRunSummary(ctx, *dagRunRef)
 			if err == nil && summary.Status != api.StatusRunning {
 				count++
+				if count >= queuedCountScanCap {
+					return count, true, nil
```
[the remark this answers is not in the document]
four-bytes-robby: Pushed a fix for the first review finding — it was a real gap, thanks.

**Cap now bounds entries scanned, not entries counted.** The cap only advanced on *visible* entries, so a running (or unreadable) entry still cost a `FindAttempt` + `ReadStatus` while never moving the counter. A queue made up of running entries therefore kept scanning without bound — precisely the case this PR exists to prevent. It now tracks entries processed, and reports `capped` only when entries actually remain unscanned, so a queue ending exactly on the cap is still an exact count.

That surfaced a second gap: with `count == 0` and `capped == true`, the `count > 0` guard skipped creating the queue, so a queue holding more running entries than the cap answered **404** instead of reporting a capped count of zero. A truncated scan proves entries exist, so it now surfaces.

Regression test added covering more-than-cap running entries ahead of the first visible queued entry.

---

On the second finding, about `queueRefreshToken` in `ui/src/pages/queues/queue/index.tsx` — I looked and did not change it, because the premise does not quite hold:

- the token is not derived from `queuedCount` alone; it also serializes `running[]` as `name:dagRunId:status`, so it changes whenever a run starts or finishes
- user-driven dequeues bypass the token entirely — the page calls `reload()` explicitly alongside `mutate()`
- the queue object itself polls on a 3s `refreshInterval`

There is a genuine residual edge: a queue pinned at the cap whose membership changes while the running set happens to be unchanged in the polled snapshot would not retrigger the feed. That is newly reachable because the count can now stick at the cap. It looks narrow and self-correcting on the next running change or user action, and closing it properly would mean exposing the queue store's `Revision` (which already changes on membership change) as a new API field — more scope than this PR needs.

Happy to add that here if you would rather have it fixed now, or to split it out.

four-bytes-robby:
Pushed a third commit that materially changes what this PR claims, so worth calling out rather than leaving in the diff.

**The per-queue cap did not bound a request.** `collectQueues` counts every queue returned by `QueueList` and gave each its own fresh budget, so the total still grew with the number of queues:

```
(number of queues) x cap x 2 filesystem operations
```

I confirmed this in production rather than in theory: running the previous build, `/api/v1/queues` still timed out. Many queues, each individually below the per-queue cap, summed to an unbounded request. So the original fix here was incomplete.

There is now a `scanBudget` created once per request and shared across every queue it counts. Either bound stops the scan and reports the count as a lower bound, so neither one deep queue nor many shallow ones can make a request unbounded. Queues reached after the budget is spent report zero as a lower bound rather than being silently omitted.

Test added covering many small queues whose combined size exceeds the request budget while no single queue reaches the per-queue cap — it fails against the previous commit.

---

One piece of context that came out of chasing this, since it affects how you may want to weigh this PR: the queue count is not the only consumer of the underlying cost. `/api/v1/dag-runs` was timing out on the same instance, and the shared cause is #2674 — the day index is never written while any run in the day is active, so every request re-parses every status file in the current day.

If #2674 is fixed, the pressure behind this PR drops a lot, and a cap may look like belt-and-braces. I still think bounding a list endpoint is correct on its own terms, but you may reasonably prefer to fix the index first and treat this as secondary. Happy either way.

internal/service/frontend/api/v1/queues.go
```diff
@@ -264,18 +309,24 @@ func (a *API) collectQueues(ctx context.Context, onlyQueue string) (map[string]*
...
-		if count > 0 {
+		// A capped scan proves entries exist even when none seen so far were
+		// visible, so the queue must still surface rather than read as absent.
+		if count > 0 || capped {
 			queue := getOrCreateQueue(queueMap, onlyQueue, a.config)
 			queue.queuedCount = count
+			queue.queuedCountCapped = capped
 		}
 	} else {
 		queueNames, err := a.queueStore.QueueList(ctx)
```
yohamta0, line 332: `QueueList` calls `queueRecordIDs(ctx, "")`, which walks and sorts every queued item path before `countVisibleQueuedItems` gets to apply a budget. On large backlogs, `/queues` and the SSE bootstrap still perform O(N) filesystem work and can trigger the exact timeout this PR aims to solve. Discovering queue names needs a bound or an index as well.

yohamta0 (commented):
First pass on the current three-commit version. This comment is outside the changed hunks:

`ui/src/pages/queues/queue/index.tsx:140`

The backend can return `queuedCount == 0` with `queuedCountCapped == true` when running entries fill the scan window before any queued entries are reached (as seen in `TestGetQueueCapsScanWhenRunningEntriesPrecedeQueuedEntries`). Disabling `useQueuedItemsFeed` here stops the page from fetching those remaining entries and leaves the UI showing "No queued items". Treat a capped count as potentially non-empty here and in the empty-state check at line 442.

```suggestion
    enabled: Boolean(
      queue &&
        ((queue.queuedCount || 0) > 0 || queue.queuedCountCapped === true)
    ),
```

internal/service/frontend/api/v1/queues.go
four-bytes-robby: Tbhe comment was solved usuually but good hint, we should take care of this.

four-bytes-robby: You are right, and I verified it: `QueueList` calls `queueRecordIDs(ctx, "")` with an empty prefix, so `collectIDs` walks every queued item path, builds an ID per file and prefix-checks it, purely to derive the set of queue names. That happens before `countVisibleQueuedItems` can apply any budget, so the request is still O(total queued items) no matter what the count does.

So this PR bounds the status reads but not the discovery. It is a genuine improvement — the per-item `FindAttempt` + `ReadStatus` was the expensive half — but it does not make the endpoint bounded on its own, and I should not claim it does.

One constraint I hit while looking at the fix, which rules out the obvious approach: queue names are not necessarily a single path segment. `ProcGroup()` falls back to the DAG name, and DAG names come from the discovery relative path (`filepath.ToSlash(relPath)`), so a DAG in a subdirectory yields a name containing `/`. `splitQueueRecordID` uses `LastIndex("/")`, consistent with that. A plain top-level `ReadDir` of the collection root would therefore miss nested queues and also return intermediate directories that are not queues.

What seems workable is directory-oriented discovery that prunes: walk directories, and once a directory is found to contain `item_` records, record its name and stop descending. That gives O(queue directories) rather than O(queued items), and stays correct for nested names. Alternatively a maintained queue-name index, as you suggested, which would be cheaper still at the cost of keeping it in sync on enqueue and delete.

Do you have a preference? I am happy either to add the pruned discovery to this PR, or to keep this one to the count and do discovery separately — the second is easier to review, since it touches the store rather than the API layer.

yohamta0: Agreed. Let's move queue-name discovery to a follow-up. This PR bounds status reads rather than the full request, so please update the PR description to match.

yohamta0 (commented):
Second pass. The capped-zero issue from my first review remains. When updating the feed and empty-state predicates, please add a UI test for `queuedCount: 0` with `queuedCountCapped: true`.

yohamta0:
Overall looks good to me. I'll make some follow up commits.

yohamta0 (approved):
LGTM! Thank you very much for finding this queue scan issue!

internal/service/frontend/api/v1/queues.go
four-bytes-robby: Agreed, and done — description updated. It now states plainly that this bounds the status reads rather than the full request, with `QueueList` walking every queued item path to derive queue names called out as the remaining unbounded part, deferred to a follow-up in the store layer.

Also pushed `f47472d01` for the capped-zero issue from your first pass. You were right that it survived; I had reasoned about the API side and missed that the page keys off the same count.

`useQueuedItemsFeed` was disabled and the empty-state branch rendered whenever `queuedCount` was 0, so a queue that returned `0` with `queuedCountCapped: true` — exactly the case in `TestGetQueueCapsScanWhenRunningEntriesPrecedeQueuedEntries` — showed "No queued items" and never fetched the remainder. Both predicates now go through a shared `queueMayHaveQueuedItems`, and `queuedCountCapped` joins the refresh token so a change in the flag alone still retriggers a load.

The UI test you asked for is in `ui/src/features/queues/lib/__tests__/queueCounts.test.ts`, covering `queuedCount: 0` with `queuedCountCapped: true`, the exact-zero case, and a missing count. I put the helper in its own module rather than the page so the test does not pull in React and the router. Full suite: 752 tests passing.

PR #2675, opened 2026-08-30
perf(persis): index the terminal runs of a day instead of requiring the whole day

### LINKED ISSUE #2674: perf: day index is never written while any run in the day is active, forcing a full re-parse per request
## Summary

`dagrunindex.RebuildForDay` persists a day index only when **every** run in that day is terminal:

```go
// Write index only if all runs are terminal and there are enough runs.
if allTerminal && len(entries) >= MinRunsForIndex {
```

The current day therefore never gets an index on any instance that schedules frequently, because there is essentially always at least one active run. Every request touching today re-runs the full rebuild: `findLatestAttempt` plus `os.Stat` plus `parseStatusFile` — a read **and** a JSON parse — for every run in the day directory.

The cost is O(runs today) per request and grows through the day, resetting at midnight.

## Why it bites

One active run forfeits caching for every finished run beside it. On an instance with DAGs on `* * * * *` the day directory holds thousands of runs by evening, nearly all terminal and immutable, and each one is re-read and re-parsed on every `/dag-runs` and `/queues` call.

Observed on v2.16.0: `/api/v1/dag-runs` (cursor-paginated, `limit=100`) timing out at ~11s alongside `/api/v1/queues`, while `/api/v1/dags` — which reads only DAG definitions, not run history — stayed fast. That split is what pointed at the day rebuild rather than at either endpoint.

`ListDAGRuns` already anticipates this, catching `context.DeadlineExceeded` and returning a 504, which suggests the ceiling has been hit before.

## Suggested direction

Let the index be a **partial** snapshot: persist entries for the terminal runs, and on read use those while scanning only the runs the index does not cover.

Sketched against the current code:

- `RebuildForDay`: drop the `allTerminal` gate and write the terminal subset, keeping the `MinRunsForIndex` threshold.
- `validateIndex`: allow a subset. The `len(idx.Entries) != len(runDirs)` equality check is what forces all-or-nothing today; the per-entry `runDirModTime` / `latestStatusSize` / `latestStatusModTime` checks already detect a terminal run that later gained an attempt, so a retried run invalidates its own entry and gets rescanned.
- Load path: walk `runDirs` in order, taking the cached entry where covered and scanning otherwise, which preserves ordering without a merge.

That turns the per-request cost for a day from *N* status reads plus JSON parses into *N* stats plus a parse for only the handful of active runs. Still O(N), but stat-bound instead of parse-bound.

No proto change appears necessary — a subset index is self-describing once validation permits it.

## Notes

`RebuildForDay` is only reached from the day load path and from tests, so the blast radius looks contained, and the existing 27 tests in `dagrunindex_test.go` cover the current semantics well.

I am happy to implement this, but the shape is yours to choose — particularly whether a partial index should be marked as such in the proto rather than inferred, and what `fromIndex` should report when a day is served from a mix of cached and freshly scanned entries. Let me know and I will send a PR.

Related: #2670 (the queues count is a second consumer of the same storage cost).

four-bytes-robby:
Fixes #2674.

## Problem

`RebuildForDay` persisted a day index only when **every** run in the day was terminal:

```go
// Write index only if all runs are terminal and there are enough runs.
if allTerminal && len(entries) >= MinRunsForIndex {
```

On an instance that schedules frequently the current day always has an active run, so its index is never written. Every request touching today re-runs the full rebuild — `findLatestAttempt`, plus a status read **and** a JSON parse, for every run in the day directory. The cost is O(runs today) per request and grows through the day, resetting at midnight.

One active run forfeits caching for every finished run beside it, and on a per-minute schedule that never resolves.

## Change

The index now holds the **terminal subset**. Reads reuse the entries that still match disk and rescan only the runs those do not cover, so the per-request cost for a day becomes a stat per finished run plus a full read for the handful still moving.

Correctness rests on the per-entry checks that already existed:

- only terminal runs are ever written to the index
- a retried run gains an attempt, which changes the run directory mtime, so its entry stops matching and the run is rescanned
- entries whose run directory has gone are dropped the same way

The exported `RebuildForDay` keeps its signature and delegates to an internal variant taking the validated cache. The fully-valid fast path in `TryLoadForDay` is untouched, so a completed day still short-circuits with no scanning at all.

No proto change: a subset index is self-describing once validation permits it.

## On the API shape

You asked in #2674 — implicitly — whether a partial index should be marked as such. I did **not** add a flag, because nothing needs to distinguish the two: a reader validates each entry against disk regardless and rescans whatever is missing, so a partial index and a stale full one are handled by the same path. Happy to add an explicit marker if you would rather have it visible in the file.

`fromIndex` now reports that an index was written, including for a day with active runs. Its only non-test consumer (`dataroot.go`) discards it and gates on `len(indexEntries) == dagRunDirCount`, which still holds because the rebuild returns entries for every run directory.

## Tests

`TestRebuildForDay_MixedStatuses` asserted the previous all-or-nothing behaviour. It now asserts that a day with active runs indexes its finished runs and excludes the active ones.

Added `TestTryLoadForDay_ReusesIndexedTerminalRunsWhileDayIsActive`, which proves the index is genuinely consulted rather than merely written: a terminal run whose status file is made unparseable in place — preserving size and mtime so the entry still validates — is still served from the index, where a rescan would drop it. It fails without this change.

Full `./internal/persis/...` suite passes, including `-race` on `dagrun/...`; `go vet` clean.

## Context

This is the root cause behind #2670 as well. The queues count was the worst-behaved consumer of the same rebuild, but `/api/v1/dag-runs` was affected identically on the instance where this was found. If this lands, the pressure behind #2673 drops considerably.

---
## Summary by cubic
Fixes #2674 by indexing the terminal runs of a day instead of requiring the whole day to be terminal, so a single active run no longer forfeits caching for every finished run beside it. Reads now reuse index entries that still match disk and rescan only the runs they don't cover, cutting the per-request cost from a full rebuild of the day to a stat per finished run plus a full read for the handful still moving.

- Correctness relies on existing per-entry checks: only terminal runs are written, a retried run's directory mtime changes so its entry stops matching, and removed run directories are dropped the same way.
- `RebuildForDay` keeps its exported signature and delegates to an internal variant that takes the validated cache.
- Reads skip rewriting the index when the terminal subset is unchanged; without this, the partial index (which never passes validation) would turn every read into a durable fsync'd write. A newly terminal run or a stale entry still triggers a rewrite.
- `fromIndex` now reports true for days with active runs; its only non-test consumer gates on `len(indexEntries) == dagRunDirCount`, which still holds.
- No proto change, and no explicit partial-index marker: readers already validate each entry against disk, so a partial index and a stale full one are handled by the same path.
- This is also the root cause behind #2670.

<sup>Written for commit 109576c6cfd3f9858cf7d9534b9bee063b42f912. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2675?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance Improvements**
  * Faster loading of daily run data by reusing valid cached index entries instead of rescanning all runs.
  * Terminal runs remain available from the index even while other runs are still active.

* **Reliability**
  * Cached run information is validated against the filesystem before reuse.
  * Improved handling when status files become temporarily unreadable or malformed.

### DIFF
--- a/internal/persis/file/dagrun/dagrunindex/dagrunindex.go
+++ b/internal/persis/file/dagrun/dagrunindex/dagrunindex.go
@@ -111,7 +111,15 @@ func TryLoadForDay(ctx context.Context, dayDir string, dagRunDirs []os.DirEntry)
 			}, nil
 		}
 
-		entries, fromIndex, rebuildErr := RebuildForDay(dayDir, dagRunDirs)
+		// The index does not cover the day exactly, which on an actively
+		// scheduled day is the normal case. Reuse whatever it still describes
+		// correctly and rescan only the remainder.
+		var cached map[string]Entry
+		if readErr == nil {
+			cached = usableIndexEntries(dayDir, idx, runDirs)
+		}
+
+		entries, fromIndex, rebuildErr := rebuildForDay(dayDir, dagRunDirs, cached)
 		if rebuildErr != nil {
 			return dayLoadResult{}, rebuildErr
 		}
@@ -146,15 +154,30 @@ func TryLoadForDay(ctx context.Context, dayDir string, dagRunDirs []os.DirEntry)
 // RebuildForDay scans a day directory, discovers latest attempts, reads statuses,
 // and writes the index if all runs are terminal.
 func RebuildForDay(dayDir string, dagRunDirs []os.DirEntry) ([]Entry, bool, error) {
+	return rebuildForDay(dayDir, dagRunDirs, nil)
+}
+
+// rebuildForDay scans a day directory, reusing the cached entries the caller has
+// already validated against disk and reading status only for the runs those do
+// not cover. Passing a nil cache scans everything.
+func rebuildForDay(dayDir string, dagRunDirs []os.DirEntry, cached map[string]Entry) ([]Entry, bool, error) {
 	runDirs := filterDAGRunDirs(dagRunDirs)
 	if len(runDirs) == 0 {
 		return nil, false, nil
 	}
 
 	entries := make([]Entry, 0, len(runDirs))
-	allTerminal := true
+	terminal := make([]Entry, 0, len(runDirs))
 
 	for _, rd := range runDirs {
+		if cachedEntry, ok := cached[rd.Name()]; ok {
+			// Only terminal runs are ever written to the index, and the caller
+			// has confirmed this entry still matches disk.
+			entries = append(entries, cachedEntry)
+			terminal = append(terminal, cachedEntry)
+			continue
+		}
+
 		runDir := filepath.Join(dayDir, rd.Name())
 		runDirInfo, err := os.Stat(runDir)
 		if err != nil {
@@ -180,10 +203,6 @@ func RebuildForDay(dayDir string, dagRunDirs []os.DirEntry) ([]Entry, bool, erro
 			continue
 		}
 
-		if status.Status.IsActive() {
-			allTerminal = false
-		}
-
 		// Parse dag run ID from directory name.
 		dagRunID := parseDagRunID(rd.Name())
 
@@ -191,7 +210,7 @@ func RebuildForDay(dayDir string, dagRunDirs []os.DirEntry) ([]Entry, bool, erro
 		startedAt := parseTimeToUnix(status.StartedAt)
 		finishedAt := parseTimeToUnix(status.FinishedAt)
 
-		entries = append(entries, Entry{
+		entry := Entry{
 			DagRunDir:            rd.Name(),
 			DagRunID:             dagRunID,
 			LatestAttemptDir:     latestAttemptDir,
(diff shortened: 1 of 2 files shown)


internal/persis/file/dagrun/dagrunindex/dagrunindex.go
```diff
@@ -222,12 +241,21 @@ func RebuildForDay(dayDir string, dagRunDirs []os.DirEntry) ([]Entry, bool, erro
...
 
-	// Write index only if all runs are terminal and there are enough runs.
-	if allTerminal && len(entries) >= MinRunsForIndex {
-		if err := writeIndex(dayDir, entries); err != nil {
+	// Persist the terminal subset. A day that still has active runs keeps an
+	// index for everything already finished, so the next read only has to scan
+	// the runs that are still moving. Requiring the whole day to be terminal
+	// meant a single active run forfeited caching for every finished run beside
+	// it, which never resolves on a frequently scheduled instance.
+	if len(terminal) >= MinRunsForIndex {
```
yohamta0, line 263: With 10 terminal runs and one unchanged running run, the second `TryLoadForDay` rewrites `.dagrun.index` again. A partial index fails `validateIndex` because its entry count differs from `runDirs`, so every request hits this block. Because `writeIndex` syncs the temporary file and day directory, each `/dag-runs` or `/queues` read becomes an O(terminal runs) durable write.

Skip the write when the reusable terminal subset has not changed.

yohamta0 (commented):
First pass. I found one issue in the partial-index write path.

internal/persis/file/dagrun/dagrunindex/dagrunindex.go
four-bytes-robby: Will check. All is verified in production to run quite fine.

four-bytes-robby: Confirmed and fixed — thank you, this was a real regression and a bad one: it traded repeated reads for repeated fsyncs, which on a network-backed volume is the worse of the two.

You were exactly right about the mechanism. A partial index never satisfies `validateIndex`, because its entry count differs from `runDirs`, so every read reached the rebuild path and rewrote the index. `writeIndex` syncs both the temporary file and the day directory, so each `/dag-runs` or `/queues` read became a durable write proportional to the finished runs in the day.

Pushed `109576c6c`: the write is skipped when nothing new has become terminal and the on-disk index holds exactly the entries that were reused. A newly finished run, or an indexed entry that no longer matches disk, still triggers a rewrite so the index keeps up.

One detail worth noting, since my first attempt got it wrong: the skip has to sit **inside** the `MinRunsForIndex` threshold. Placed above it, a day with nothing indexable satisfies `reused == len(terminal) == indexedCount` trivially at zero and wrongly reports that an index exists. Two existing tests caught that.

Two tests added:

- repeated reads over an unchanged day must not alter the index mtime, which is the regression you found
- a run reaching a terminal state must still be added to the index, so the skip cannot silently freeze it

yohamta0 (approved):
LGTM ! Thanks a lot for improving performance!

PR #2738, opened 2026-09-09
tools/aqua: serialize standard registry cache refreshes

### LINKED ISSUE #2727: Bug: GitHub aqua-registry rate limiting slows/stalls concurrent black-box tests
## Describe the bug

When a conformance test runs multiple independent steps that each reference a remote/official action (ffmpeg@v1, github-cli@v1, rclone@v1, etc.) concurrently in the same DAG, each step's cold invocation     independently hits GitHub's aqua-registry API to resolve tool versions. Firing these concurrently exhausts GitHub's anonymous rate limit, producing 403 Forbidden on GET /repos/aquaproj/aqua-registry/releases/latest.
Dagu falls back to a bootstrap registry ref when this happens, but the fallback path is much slower, causing steps that normally take a few seconds to take ~60s+ each.

## To Reproduce
Full test suites for remote-action specs (spec064_ffmpeg, spec065_github_cli, spec066_rclone) intermittently balloon from seconds to minutes when multiple steps provision tools in parallel within one dagu start invocation.
   
## Possible workaround: 

Chain the tool-provisioning steps sequentially (depends: + continue_on: failed set on the failing step itself) instead of leaving them as independent/concurrent siblings, so only the first step pays the cold aqua-registry resolution cost and later steps reuse the warmed cache. This fixed the slowdown in all three affected specs.

## Environment

- Dagu CI

fancive:
## Summary

Prevent concurrent installers sharing a tools directory from duplicating GitHub registry metadata requests on a cold cache.

## Changes

- Use the existing OS-backed file lock and recheck the cache after acquiring it.
- Write cache contents atomically; preserve explicit refresh and fallback behavior.
- Add concurrent installer, cross-process, cancellation, unavailable-cache/lock and process-exit regression coverage.

Validation on macOS with Go 1.27:
- Affected package tests with race detection and repository lint passed (including Windows lint).
- Real CLI cold/warm runs installed and executed jq successfully; the warm run reused the registry cache.
- An earlier patch's full suite had one cancellation-test failure. Baseline and final-patch isolated reruns each passed 10 times; the original failure remains unexplained. The final patch's full suite was not rerun. Three opt-in Aqua network integration tests were skipped.

## Related Issues

Closes #2727

## Checklist

- [x] Documentation has been updated as needed (not applicable to this internal cache change)

---
## Summary by cubic
Fixes a race condition where concurrent installers sharing a tools directory would each fetch the aqua registry metadata on a cold cache, duplicating network requests. Now installers take an OS-backed file lock, recheck the cache after acquiring it, and share a failed fetch with concurrent waiters, so only one process performs the lookup.

- Writes cache contents atomically and preserves explicit refresh and fallback behavior.
- Timestamps failed lookups so waiters, including forced refreshes, reuse a shared failure instead of retrying.
- Keeps the lock file separate from the cache, so the OS releases it on process exit without stale-lock reclamation.
- Adds regression tests for concurrent installers, cross-process sharing, cancellation, unavailable cache/lock, and process-exit scenarios.

Closes #2727.

<sup>Written for commit e5b16238c1d44867c071ea53f3e277315c36d74a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2738?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple installations access the registry cache at the same time.
  * Prevented duplicate registry lookups during concurrent operations.
  * Added recovery when a previous process exits unexpectedly while holding the cache lock.
  * Preserved fallback behavior when the cache or lock is unavailable.
  * Improved safe writing of updated registry cache data.

### DIFF
--- a/internal/tools/aqua/registryref.go
+++ b/internal/tools/aqua/registryref.go
@@ -37,6 +37,11 @@ type latestRegistryRef struct {
 	FetchedAt time.Time `json:"fetchedAt"`
 }
 
+type latestRefCacheEntry struct {
+	latestRegistryRef
+	FailedAt time.Time `json:"failedAt,omitzero"`
+}
+
 type registryRefSource int
 
 const (
@@ -57,7 +62,9 @@ type resolvedRegistryRef struct {
 // previously cached ref of any age is used, then the compiled-in bootstrap
 // ref, so resolution never fails.
 func (i *Installer) resolveStandardRegistryRef(ctx context.Context, opts tools.InstallOptions, forceRefresh bool) resolvedRegistryRef {
+	startedAt := i.now()
 	cachePath := i.latestRefCachePath(opts)
+	cacheLockHeld := false
 	if !forceRefresh {
 		if cached, ok := readLatestRefCache(cachePath, i.now()); ok {
 			return resolvedRegistryRef{cached, registryRefSourceCache}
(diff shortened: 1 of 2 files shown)


yohamta0 (approved):
LGTM after the follow-up fixes. Thanks for tackling this and adding the concurrency coverage! CI and conformance are green on Linux and Windows.

PR #2742, opened 2026-09-09
Recover catchup after scheduler pauses

yohamta0:
## Summary

Recover missed start schedules when the scheduler resumes after sleep or a long
pause. Previously, the cron loop drained old ticks as live runs, bypassing the
catchup window and overlap policy.

## Changes

- Replay missed starts through the existing catchup queue, then resume the
  current minute. Disabled catchup skips historical ticks.
- Preserve pending catchup work when workflow defaults or definitions change.
  Removed schedules and disabled catchup discard their pending entries.
- Interpret unzoned schedules in the configured timezone even when their
  persisted checkpoint is UTC. Explicit schedule timezones retain precedence.
- Keep active runs intact and defer queued catchup using existing overlap rules.
- Merge pending and recovered catchup slots before enforcing the buffer limit,
  retaining the newest 1,000 slots and checkpointing discarded slots.
- Advance the integration test's simulated clock between its two expected
  scheduled slots, and verify their timestamps.

## Validation

- Overflow regression tests fail before the fix and pass afterward. Coverage
  includes all overlap policies, capacity boundaries, repeated recovery, and
  enqueue retries, and persisted discard checkpoints.

- Reproduced wake-gap, pending-work loss, and timezone failures before fixing.
- Focused scheduler tests pass, including repeated recovery, all/latest policies,
  busy workflows, disabled queues, clock rollback, and Asia/Tokyo schedules.
- `make fmt` passes native and Windows lint.
- Full scheduler race tests pass. Conformance passes: 1,103 tests with five
  platform-specific skips on Mac.
- Integration race tests pass: 435 passed, 19 skipped. The frozen-clock fixture
  failure was reproduced before correcting it.
- Runstead's 84 UI/editor tests pass against the patched executable; a Mac
  development package builds with the same engine.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scheduler behavior when system time moves forward or backward, preventing incorrect future dispatches.
  - Ensured missed scheduled runs are recovered reliably after delays or scheduler restarts.
  - Preserved pending catch-up work when schedules are updated.
  - Improved catch-up accuracy for schedules configured in different time zones.
  - Removed stale queued work when scheduling is disabled or no active schedules remain.
  - Improved handling of duplicate and capacity-limited pending scheduled runs.
- **Tests**
  - Added coverage for clock changes, scheduler recovery, time zones, and schedule updates.

### DIFF
--- a/internal/intg/sched_test.go
+++ b/internal/intg/sched_test.go
@@ -52,18 +52,25 @@ steps:
 	require.NoError(t, err)
 
 	var dispatchCount atomic.Int32
-	schedulerInstance.SetDispatchFunc(func(_ context.Context, entry scheduler.DAGEntry, _ string, trigger ir.TriggerType, _ time.Time) error {
+	var dispatchMu sync.Mutex
+	var dispatchTimes []time.Time
+	schedulerInstance.SetDispatchFunc(func(_ context.Context, entry scheduler.DAGEntry, _ string, trigger ir.TriggerType, scheduled time.Time) error {
 		dag := entry.DAG
 		if dag != nil && dag.Name == "cron-test" && trigger == ir.TriggerTypeScheduler {
+			dispatchMu.Lock()
+			dispatchTimes = append(dispatchTimes, scheduled)
+			dispatchMu.Unlock()
 			dispatchCount.Add(1)
 		}
 		return nil
 	})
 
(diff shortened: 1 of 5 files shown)


yohamta0:
@coderabbitai review

yohamta0:
@coderabbitai review

PR #2763, opened 2026-09-13
fix(mcp): serve the dagu://dags collection resource

### LINKED ISSUE #2604: Reading DAGs (with MCP) returns completion message instead of DAG list
When requesting/reading all DAGs, with MCP read_dagu call, the response only returns the status message 
"Dagu read completed" rather than the actual list of DAGs.

Image

The read operation should return an array/list of all configured DAGs alongside or instead of the static completion message.

LouisDeconinck:
## Summary

Fixes #2604 (remaining gap after #2605).

`dagu_read` with `target=dags` already embeds the DAG list in the visible result text since #2605. What is still missing is the DAG collection resource itself: Spec 020 declares `dagu://dags` as a supported resource family, and `dagu_read` in URI mode returns a `resource_link` pointing at it — but `resources/read` on `dagu://dags` returned resource-not-found and `resources/list` did not advertise it. An MCP client following that link or browsing resources had no way to read the DAG list.

## Changes

- `internal/service/mcp/server.go`: register `dagu://dags` as an MCP resource (mirroring the existing `dagu://wiki` collection) and serve it from `readResourceText`, returning the same normalized DAG collection model (`items` + `pagination`, `application/json`) that `dagu_read` returns for `target=dags`. Collection query parameters are validated with the same rules as the tool's URI mode.
- `internal/service/mcp/audit.go`: audit `dagu://dags` collection reads as resource type `dags` instead of `dag_spec`.
- `internal/service/mcp/server_test.go`: regression test covering `resources/list` advertising `dagu://dags` and `resources/read` returning the DAG list, including DAGs from `paths.alt_dags_dir`.

Note: the `dagu://runs` and `dagu://reference` collection URIs from Spec 020 have the same read gap; happy to follow up with the same pattern if useful.

## Testing

- `go test ./internal/service/mcp/...` — pass (new test fails before the fix: `resource "dagu://dags" not found`)
- `go test ./conformance/spec020_mcp_server/ ./conformance/spec021_mcp_read_tool/` — pass
- `go vet`, `gofmt`, `go fix -diff`, `golangci-lint run ./internal/service/mcp/...` — clean

---
## Summary by cubic
Fixes #2604 by serving the `dagu://dags` collection resource. MCP clients following a `resource_link` can now read the DAG list; previously `resources/read` returned resource-not-found and `resources/list` did not advertise it.

**Changes**
- `resources/read` on `dagu://dags` returns the same JSON collection model (`items` + `pagination`) as `dagu_read` with `target=dags`, including DAGs from `paths.alt_dags_dir`; filtered reads with list query params resolve to the same data.
- Collection reads are audited as `dags`; per-DAG `spec` reads remain `dag_spec`.
- The read tool and resource read path now share the same DAG collection lookup.
- Adds regression coverage for the resource being listed, read, and containing DAGs from the alternate DAGs dir.

<sup>Written for commit 9ba018c98a501d6d3efc1c610c54511ce63f2fa9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an MCP resource for browsing the complete DAG collection.
  * DAG listings now include DAGs from primary and alternate directories, along with their specification URIs.
  * Improved resource identification for DAG collections and individual DAG specifications.


yohamta0:
Thanks — this is a good catch and the regression test is solid. I pushed one follow-up commit to the branch rather than asking for another round-trip.

**What it fixes.** `resources/read` resolves a URI by exact map lookup first and by URI template second, and never strips the query (go-sdk `lookupResourceHandler`). So registering `dagu://dags` with `AddResource` alone serves the bare URI but not the filtered form:

| URI | before | after |
| --- | --- | --- |
| `dagu://dags` | ok | ok |
| `dagu://dags?name=x` | `Resource not found` | ok |

That mattered here because `dagu_read` in URI mode emits a `resource_link` that preserves the query (`uriWithQuery`, read_tool.go:556), so the dangling-link case this PR set out to fix still reproduced whenever a filter was used. Spec 020 also documents `dagu://dags?name=nightly&perPage=20` as a supported resource URI.

A side effect was that the `validateReadQuery` call in your new branch was unreachable — `parsed.RawQuery` was always empty there. It is now live and covered.

**The change** is a single `AddResourceTemplate` for `dagu://dags{?page,perPage,name,labels,active,sort,order}`, matching the parameter set `isAllowedReadQueryParam` permits. The generated regex does not reference variable names, so any order, any subset, and zero parameters all match; the exact resource still wins for the bare URI, and `dagu://dags/{name}/spec` sorts first, so nothing is shadowed. Your handler body needed no change.

**One limitation worth knowing:** the SDK's template regex allows only unreserved characters plus `,`, `=` and `%XX` in the query, so a `name` filter containing a raw `/` or space must be percent-encoded by the client. `page`/`perPage`/`active`/`sort`/`order` are digits or fixed words, and labels like `production` or `env=prod` are fine, so in practice only `name` is affected.

Verified: `go test ./internal/service/mcp/...`, conformance spec020/021/022, `golangci-lint` clean. Also checked the full round trip — the `resource_link` the read tool hands out now resolves, and `perPage=99999` is still rejected.

On your note about `dagu://runs` and `dagu://reference`: yes please, a follow-up PR would be welcome. They have the same read gap *and* the same query-routing gap (`dagu://runs/{name}/{id}/logs?tail=100` currently 404s for the same reason, which makes the step-log `validateReadQuery` branch dead too), so the same template pattern applies there.

yohamta0:
Addressed the duplication nit in `9ba018c9`.

`svc.listDAGs(ctx, query)` now sits alongside the other DAG service helpers in `server.go`, and both the resource branch and the read tool's `target=dags` branch call it. That makes the DAG collection consistent with every other family — `getDAGSpec`, `listWikiPages` and `getWikiPage` are each already shared between `readResourceText` and `read_tool.go` — and the resource branch now reads the same as the wiki collection branch directly below it. `requireAPI` stays at the call sites, mirroring `listWikiPages`. Pure refactor, no behavior change, so no new test; `TestDAGsCollectionResource` and `TestReadToolListsAndReadsAltDAGsDir` already cover both paths.

One deliberate exclusion: `dagNameSuggestions` in `suggest.go` is the third caller of `GetDAGsListDataIncludingAltDirs`, but it needs the typed `[]daguapi.DAGFile` off the raw response to build did-you-mean candidates rather than the normalized model, so it keeps its own type switch.

On the other two nits:

- **Conformance coverage** — not adding it here. The premise that conformance-level coverage is the norm for MCP resources doesn't hold: there is exactly one `ReadResource` call in the entire `conformance/` tree (the reference topic in `spec020_mcp_server`), against roughly twelve resource families declared in spec 020. Worth doing as its own PR that covers the family properly rather than pinning one resource because it happens to be the one in flight.
- **`dagu://runs` / `dagu://reference`** — agreed, and the template pattern here is the one to reuse. Left for the follow-up PR.

Verified: `go test ./internal/service/mcp/...`, conformance spec020/021/022, `golangci-lint` clean.

Still open and worth knowing about: the SDK's template regex permits only unreserved characters plus `,`, `=` and `%XX` in a query, so a `name` filter containing a raw `/` or space still won't route and must be percent-encoded by the client.

yohamta0 (approved):
LGTM, thank you very much!

PR #2770, opened 2026-09-14
fix(scheduler): reload inherited base defaults

### LINKED ISSUE #2767: bug: base config never reaches the scheduler — catchup_window has no effect, overlap_policy cannot be inherited
## Describe the bug

Two defects that share one theme: values written in the **base config** (`$DAGU_HOME/base.yaml`) do not reach the scheduler, even though `base.yaml` advertises them as inheritable defaults.

The base config Dagu itself writes for a new install (`internal/persis/file/baseconfig/default_base_config.go`) lists both fields under "Active Defaults":

```yaml
# Behavior when a new run is triggered while one is already active.
# "skip" = discard the new run. "all" = queue all runs. "latest" = keep only the most recent.
overlap_policy: skip

# Lookback window for replaying missed scheduled runs after downtime.
# On scheduler restart, all missed cron intervals within this window are executed (max 1000).
catchup_window: 24h
```

Neither line can take effect.

### 1. `overlap_policy` in the base config is always overridden

`buildOverlapPolicy` maps an unset field to `ir.ParseOverlapPolicy("")`, which returns the **non-zero** value `"skip"` (`internal/ir/overlap_policy.go:16,29-30`; `internal/spec/dag.go:1105-1107`). Base composition is `mergo.Merge(effective, current, WithOverride)` (`internal/spec/dag.go:922-938` → `internal/spec/loader.go:1325-1328`), which only lets a **zero** field inherit, and there is no `mergeTransformer` case for `OverlapPolicy` (`internal/spec/loader.go:1044-1105`).

So every DAG always carries a non-zero `overlap_policy` and always wins over the base. `overlap_policy: all` in `base.yaml` silently resolves to `skip` for every DAG that does not repeat it. `catchup_window` does not have this problem — `buildCatchupWindow` returns `0` when unset (`internal/spec/dag.go:1098-1103`), so it inherits normally on a full load.

### 2. The scheduler never reads the base config at all

The scheduler's registry is built from **metadata-only** loads — `internal/service/scheduler/entryreader.go:470` (`dagRepository.GetMetadata`) and `internal/service/scheduler/dag_file_source.go:57-65` (`spec.Load(..., OnlyMetadata(), ...)`). Under `buildFlagOnlyMetadata`, `loadBaseDefinition` returns `nil` before reading anything (`internal/spec/loader.go:576-580`), so the base config is dropped. The DAG store does pass `WithBaseConfig` / `WithWorkspaceBaseConfigDir` (`internal/persis/file/dag/store.go:255-262`), but `GetMetadata` requests `spec.OnlyMetadata()` (`store.go:355-372`), so those options never apply.

`queue`, `catchup_window`, `overlap_policy`, `skip_if_successful` and `max_active_runs` are all metadata-stage fields (`internal/spec/dag.go:553-566`), and the scheduler acts on them directly:

- catch-up buffers: `if dag.CatchupWindow <= 0 { continue }` — `internal/service/scheduler/tick_planner.go:418-421`
- buffer policy: `NewScheduleBuffer(dag.Name, dag.OverlapPolicy)` — `tick_planner.go:452`
- overlap guards: `IsRunning` → `CountAliveByDAGName(dag.ProcGroup(), dag.Name)` (`scheduler.go:279`) and `IsQueued` → `ListByDAGName(dag.ProcGroup(), dag.Name)` (`scheduler.go:244`), consumed at `tick_planner.go:554`, `:751`, `:918`

Meanwhile the run the scheduler launches is a subprocess `dagu start|enqueue <file>` (`internal/launcher/launcher.go:135-236`), i.e. a **full** load that does apply the base config, and it registers its proc entry under the merged `dag.ProcGroup()` (`internal/intake/local.go:89`, `internal/engine/run.go:637`). The two halves of one tick therefore read different definitions of the same DAG.

## To Reproduce

Two identical setups, one running `dagu scheduler`, differing only in *where* `catchup_window` is declared.

**A — `catchup_window` in the base config**

```yaml
# $DAGU_HOME/base.yaml
catchup_window: 24h
```
```yaml
# $DAGU_HOME/dags/gamma.yaml
schedule: "* * * * *"
steps:
  - command: echo tick
```

**B — the same value in the DAG (control)**

```yaml
# $DAGU_HOME/dags/gamma.yaml
schedule: "* * * * *"
catchup_window: 24h
steps:
  - command: echo tick
```

Start the scheduler in each, stop it, wait past a minute boundary so one occurrence is missed, then start it again.

Full loads do see the base config, for contrast:

```
$DAGU_HOME/base.yaml        ->  queue: pool
$DAGU_HOME/dags/alpha.yaml  ->  steps: [{command: sleep 30}]

$ dagu enqueue dags/alpha.yaml
$ find $DAGU_HOME/data/queue -type d
.../data/queue/pool          # base applied
```

A DAG that writes `queue:` itself lands in its own queue, as expected — so the inheritance works on the execution path and only the scheduler's view is missing it.

## Expected behavior

A field declared in the base config reaches the scheduler exactly as if the DAG had declared it. Concretely: `catchup_window` in `base.yaml` enables catch-up for every DAG that does not set it, and `overlap_policy` in `base.yaml` governs the schedule buffer.

## Actual behavior

**B (control)** replays the missed occurrence:

```
level=INFO msg="Catch-up planned" dag=gamma missedCount=1 replayFrom=2026-09-14T15:42:00+09:00 replayTo=2026-09-14T15:43:27+09:00
level=INFO msg="Catch-up initialization complete" dagCount=1 totalMissedRuns=1
level=INFO msg="Dispatching planned run" dag=gamma scheduleType=Start scheduledTime=2026-09-14T15:43:00+09:00
level=INFO msg="Catchup run enqueued" dag=gamma run-id=catchup-be9d587defa1-20260914T064300
```

**A** does not. There is no catch-up buffer, no `Catch-up planned`, no replay — only the current occurrence:

```
level=INFO msg="Dispatching planned run" dag=gamma scheduleType=Start scheduledTime=2026-09-14T15:43:00+09:00
```

A `schedule:` declared only in `base.yaml` is likewise never seen by the scheduler: the DAG is loaded (`dagCount=1`) but no occurrence ever fires.

For `overlap_policy` the effect needs no runtime at all: an unset field builds to the non-zero `"skip"`, so the base value loses the merge on every load, full or metadata.

## Environment

- Dagu version: 2.16.2 (binary); code references traced on `main` @ `2635b927b`
- OS: macOS 15 (arm64)
- Installation method: binary

## Additional context

Notes toward a fix:

- For (1): either build an unset `overlap_policy` as the empty `ir.OverlapPolicy` and resolve the default where it is used, or add a `mergeTransformer` case so a DAG that did not author the field does not override the base.
- For (2): merging the base config into metadata loads fixes the whole class at once. The DAG store already tracks `base.yaml`'s size and mtime and invalidates both its cache and its index when they change (`internal/persis/file/dag/store.go:267-313`), so the freshness machinery is already in place. If a full merge is too costly for listing, merging only the metadata-stage fields would be enough.
- Whatever shape the fix takes, `ProcGroup()` must resolve identically on both sides of a tick: the scheduler's overlap guards look the DAG up by process group, while the run registers under the group the full load produced. A mismatch there fails open silently — two runs of the same DAG.

## What I could not reproduce

I expected the `ProcGroup()` mismatch above to be observable as overlapping runs: with `queue: pool` in `base.yaml` only, `schedule: "* * * * *"` and a 180-second step, I expected the scheduler to start a second run while the first was alive. It did not — the occurrences at +1m, +2m and +3m were all skipped and the next run started only after the first finished, which is the correct guarded behaviour. Either something else keeps the group consistent on that path, or the guard is not the code path I traced. Worth confirming before treating that particular symptom as real; the two defects above do not depend on it.

yohamta0:
## Summary

Scheduling defaults in global and workspace `base.yaml` were missing from metadata loads. The scheduler could ignore inherited schedules, catchup settings, overlap policies, and queues.

Apply the same scheduling inheritance to metadata and execution loads, and refresh the live scheduler when base configuration changes. Share unchanged base sources across DAG loads while rebuilding each DAG with its own unevaluated manifest copy.

## Changes

- Preserve global → workspace → DAG precedence, including explicit `false`, empty queue/catchup overrides, overlap policies, and concurrency limits.
- Cache base file contents and parsed YAML with bounded capacity. Compare modification time and size on access; concurrent readers share work only for the same file version. Embedded worker configuration stays independent of local files.
- Include base state in metadata cache keys so an earlier in-flight load cannot replace current metadata. Invalidate persisted indexes on the first lookup and after base edits.
- Keep the `EntryReader` interface in the scheduler and move its file implementation and monitoring into the file persistence adapter. The CLI injects the adapter; scheduling consumes definition changes without filesystem knowledge.
- Extend directory monitoring to global and workspace bases. Reuse the existing debounce to reload definitions and notify the planner after base creation, edits, replacement, or deletion. Restore replaced directory and symlink watches, including missing targets, and stop removed polling watches. Update fsnotify to 1.10.1 for its upstream dangling-symlink and directory-watch fixes.
- Keep metadata loading free of runtime evaluation and step-default building. Document automatic refresh and explicit inheritance overrides in the base template.
- Fix test failures found during the requested validation: synchronize on complete child start markers, compile helpers before workflow timeouts, and make live-schedule fixtures explicitly opt out of inherited catchup.

Native watcher limitation: on macOS, fsnotify/kqueue can omit events when a parent-directory symlink is replaced under the same name. That operation requires another observed definition/configuration change or a scheduler restart. Ordinary base-file edits, atomic file replacement, and symlink-target deletion/recreation are covered; this change adds no periodic refresh loop.

The generated `catchup_window: "6h"` now applies to scheduled DAGs that inherit it. Set `catchup_window: ""` on a DAG to disable inherited catchup.

## Validation

- Focused regressions cover metadata/full-load parity, global/workspace/embedded bases, explicit overrides, planner updates, cache reuse and isolation, in-flight updates, timestamp rollback, invalid configuration recovery, and watcher replacement/removal.
- Observed the cache rollback, concurrent base read, missed directory-watch update, and polling-watch shutdown regressions fail before their fixes and pass afterward.
- `make fmt` passes, including native and Windows linting.
- CI passes on `2572f9b3a`: all Linux/Windows test shards, both conformance jobs, all three E2E shards, lint, vulnerability checks, and container scans.
- CodeRabbit reviewed the final commit and found no actionable issues. All review comments and replies have been checked.

## Related Issues

Closes #2767

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added inheritance for scheduling settings from global and workspace defaults, with DAG-level overrides.
  - Global and workspace base configuration changes now reload automatically.
  - DAG discovery now detects file additions, updates, removals, renames, and base-configuration changes.
  - CLI wiring injects the filesystem DAG reader into the scheduler.

- **Bug Fixes**
  - Improved cache invalidation so updated configurations are reflected immediately.
  - Corrected handling of explicit empty values and file changes when timestamps move backward.
  - Metadata-only loads now surface invalid base scheduling configuration errors.


yohamta0:
@coderabbitai review

05eba942c fixes the two conformance failures found locally:

- The parallel abort test now waits for the complete start marker. Shell redirection can create the file before `printf` writes its contents.
- The direct-execution tests build their helper before starting the workflow. A cold Go compile no longer consumes the workflow command timeout.

Both affected packages passed five runs with race detection. The full conformance suite and native/Windows lint checks also pass locally.

The first review's docstring-percentage warning does not identify an undocumented public contract. I kept comments focused on non-obvious behavior, as required by the repository's comment guidelines, rather than adding repetitive descriptions to private builders and tests.

yohamta0:
@coderabbitai review

13d554720 fixes the two scheduler integration failures seen on Linux and Windows. These fixtures test live cron dispatch and schedule edits. They now explicitly set `catchup_window: ""` so the generated six-hour catchup default does not route them into startup replay.

Both failures reproduced locally before this change and passed afterward. Their assertions and timeout budgets are unchanged. Inherited catchup remains covered by `TestInheritedScheduling` and the loader regressions.

The latest commit is ready for CI validation.

yohamta0:
@coderabbitai The conformance fixes are intentionally included. The requested scope covers fixing brittle or flaky tests encountered during validation, in addition to resolving #2767 and getting CI green. Both conformance failures occurred during this PR's validation; commit 05eba942c fixes their causes without increasing timeouts or reducing concurrency.

I have clarified that scope in the PR description. Please assess the scope warning against that requirement. The repository comment guidance discussed above still applies to the docstring-percentage warning.

yohamta0:
@coderabbitai review

Please review the new commits through 2572f9b3a.

- `EntryReader` remains owned by the scheduler. Its existing file implementation moved into the file persistence adapter and is injected by the CLI.
- Unchanged base sources share a bounded cache keyed by file state. Metadata cache keys isolate older in-flight loads.
- Base changes now update the live planner through definition events. Tests cover source reuse, stale-load races, directory/symlink lifecycle, and ordinary DAG edits through a symlinked root.
- Focused regressions and native/Windows linting pass. Full CI is running. The macOS same-name parent-symlink notification limitation is documented in the PR.

The requested scope still includes fixing test-stability failures encountered during validation. The previous replies accepting that scope and the repository's comment guidance still apply.

PR #2784, opened 2026-09-16
feat: move default suspend flags directory under data_dir

### LINKED ISSUE #1861: Move suspend flags under data dir to avoid cross-process misconfiguration
## Summary

DAG suspension is currently stored in `cfg.Paths.SuspendFlagsDir`, which is a separate path from `paths.data_dir`.

Local filesystem persistence is still the intended model. If the server and scheduler observe the same `suspend_flags_dir`, the current behavior is correct.

The problem is that suspension state sits outside the main shared data boundary, so a deployment can share the usual Dagu data storage while still misconfiguring suspend state visibility across processes. In that case, one process can report a DAG as suspended while another scheduler process continues to enqueue scheduled runs.

This issue is therefore about tightening the persistence boundary and defaults, not replacing local filesystem storage.

## Current Behavior

Current source behavior:

- API/UI writes suspend state with `dagStore.ToggleSuspend(...)`
  - `internal/service/frontend/api/v1/dags.go`
  - `internal/persis/filedag/store.go`
- scheduler checks suspension with `dagStore.IsSuspended(...)`
  - `internal/service/scheduler/scheduler.go`
  - `internal/service/scheduler/tick_planner.go`
  - `internal/service/scheduler/queue_processor.go`
- the backing location comes from `cfg.Paths.SuspendFlagsDir`
  - `internal/cmn/config/path.go`

That means:

- if all relevant Dagu processes observe the same `suspend_flags_dir`, suspension works correctly
- if they do not, suspend state can diverge across processes

## Why This Is Easy To Misconfigure

- many persisted stores are derived from or grouped under `data_dir`
- `suspend_flags_dir` is a separate configurable path
- charts/docs already place it on shared storage, but Dagu still allows deployments where suspend state is not aligned with the main shared persistence root

## Expected Behavior

Suspension should remain filesystem-based, but its default persistence location should align with Dagu's main shared data boundary.

More concretely:

- API/UI and scheduler should observe the same suspension state by default
- distributed/shared-process deployments should not need an extra independent suspend-path decision for correctness
- single-node behavior should remain unchanged

## Proposed Direction

- move suspend flags under `data_dir` by default
- keep filesystem-based storage; do not introduce a separate database requirement
- if needed, preserve compatibility with existing explicit `suspend_flags_dir` values during migration
- optionally add validation or warnings when `suspend_flags_dir` is configured outside the main shared persistence root

This would make the configuration model clearer: shared state stays under the shared data dir, and suspend behavior becomes harder to misconfigure across Dagu processes.

LouisDeconinck:
## Summary

Move the default suspend flags directory from `<home>/suspend` to `<data_dir>/suspend` so processes sharing application data also share suspension state. Explicit suspend-path settings remain unchanged and disable legacy fallback and migration. Configured paths outside `data_dir` produce a warning.

## Migration and state changes

- Repository construction preserves existing flags. Reads and DAG listings recognize flags in either the current or legacy default directory.
- The active scheduler copies missing flags after acquiring leadership, with its heartbeat running, before queues, retries, or schedules execute. Existing destination flags and all legacy flags remain intact.
- Suspending writes the current flag before deleting its legacy counterpart. Resuming removes both. Migration and state changes use a shared file lock so startup cannot undo a concurrent resume.
- Migration errors stop scheduler startup and release leadership. SSE watches existing legacy directories so legacy-only changes refresh the UI.

## Upgrade requirements

Server and scheduler processes must be able to access the same legacy flags; reconcile private per-host directories before upgrading. Upgrade all processes together before changing suspension state. Mixed-version writes are unsupported. For rollback, explicitly configure the older version to use the new shared suspend directory.

## Validation

- Config, persistence, scheduler, command, API, and SSE package tests passed.
- Focused migration, concurrency, scheduler, API, and SSE regressions passed with race detection.
- Suspend/resume scheduler integration tests passed.
- `make conformance`: 1,108 tests, 5 platform/tool skips, no failures.
- `make fmt` and `make check` passed, including Windows lint checks.

Closes #1861


yohamta0 (approved):
LGTM! Thanks a ton!

PR #2785, opened 2026-09-16
feat: add retry --bypass-preconditions flag

### LINKED ISSUE #1014: feat: Allow bypassing preconditions for retry
Hi,

When a step fails or (marked as failed), if that step has a time precondition (for example), the retry does not launch that step if the precondition is no longer met.

Is it possible to add an option asking if we want to bypass preconditions?

Regards,
Jeremy

LouisDeconinck:
## Summary

Adds `dagu retry --step <name> --bypass-preconditions` so a step can be retried after its preconditions stop matching, such as when a time window has closed. With `--downstream`, the selected step and its reachable descendants bypass step preconditions for that retry. DAG-level preconditions and lifecycle handlers still apply.

The flag requires `--step` and a local CLI context. Targeted child retries carry the option through local execution and distributed dispatch, including protobuf serialization and worker execution. The workflow definition stays unchanged, and later retries without the flag evaluate preconditions normally.

Retrying a parent step keeps preconditions active inside newly started child DAGs, including their automatic step retries. Bypass reaches a child only through an explicit child retry path.

## Validation

- Reproduced the distributed child retry bug with a worker execution regression before fixing propagation; the regression now passes.
- Covered target-only and downstream bypass, unaffected handlers and unrelated steps, and a subsequent retry without bypass.
- Reproduced automatic parallel-child retries inheriting a parent-only bypass. The regression failed before the scope guard and passes afterward; explicit child retry propagation remains covered.
- Covered protobuf serialization in both directions and persisted dispatch-task claim/reclaim.
- Race-enabled tests for affected CLI, runtime, dispatch, persistence, coordinator subflow, worker, and protobuf conversion packages.
- Full integration, executor, DAG executor, coordinator subflow, and worker race suites, plus the CLI retry tests, passed after the child-scope fix.
- The combined local CLI run timed out in `TestRestartCommand`; that test passed alone, and the remaining CLI suite passed separately.
- `make proto`, `make fmt`, `make check`, and `make conformance` (1,108 tests, 5 platform-specific skips).

## Documentation

Updated the bundled CLI reference and the separate documentation site's retry and child-run guides. The docs build and validation of 829 workflow examples passed.

https://github.com/dagucloud/docs/commit/3f12aaad51fac9e04aad831c94ba44c94ce049f0

Closes #1014


yohamta0 (approved):
LGTM, thank you very much!

PR #2788, opened 2026-09-16
feat(ui): add Artifacts page listing runs with files

yohamta0:
## Summary

Adds a new `/artifacts` page for browsing artifact files produced by DAG runs from the `GET /artifacts` recent-artifacts list API.

## Layout

Two-pane design, reusing the existing run-artifacts UI:

- **Left panel — file list**: artifact files grouped under their DAG run. Runs expand into subdirectories and files (with sizes); the newest run auto-expands and its first file auto-selects. Cursor-based `Load more` at the bottom of the list.
- **Right panel — preview**: markdown (rendered/raw), HTML, text, image, and binary previews with Copy and Download — the exact same preview pane used by the run-details Artifacts tab.
- **Filter bar** above both panes: DAG name, file name (Enter to apply), date range (preset or custom). All API calls carry `remoteNode` and the workspace scope.

## Implementation

- **`ui/src/features/dags/components/artifacts/ArtifactFilePreview.tsx`** (new): the preview pane extracted from `ArtifactsTab` (preview fetch, image blob, copy/download, view modes, all render branches). Handles root and sub-DAG-run endpoints.
- **`ArtifactsTab.tsx`**: now renders the shared `ArtifactFilePreview` in its right pane; tree/selection logic unchanged.
- **`ui/src/pages/artifacts/index.tsx`**: two-pane page (file tree on the left, `ArtifactFilePreview` on the right), filters, re-anchored cursor pagination via `usePaginatedArtifacts`.

## Review fixes (previous commits)

- Cursor pagination re-anchors when the head page moves; no silent SWR refresh (artifacts + dag-runs hooks).
- Date filters initialize from the active preset ("Today" bound) and interpret custom ranges in the configured timezone (`utcOffset(offset, true)`).
- List API errors surface instead of showing an empty state.
- New-tab URLs honor `config.basePath` (artifacts page, `DAGRunTable`, `DAGRunGroupedView`, `DAGRunCard`).

## Verification

- `pnpm vitest run` — 161 files / 892 tests pass (page tests rewritten for the two-pane layout; `ArtifactsTab` tests exercise the extracted pane)
- `pnpm typecheck` — clean
- `pnpm build` — webpack compiles (pre-existing size warnings only)
- lint clean on touched files

Out of scope: SSE live updates for the page (manual refresh + per-page reload), and the same base-path fix for raw `window.open` URLs in the run-details modal/panel and the DAGs page table (shared URL helpers — follow-up).

## Summary by CodeRabbit

* **New Features**
  * Added an Artifacts page for browsing, filtering, and previewing files produced by DAG runs.
  * Added date-range filters, pagination, expandable run trees, keyboard navigation, downloads, and file previews.
  * Added Artifacts to the navigation menu with localized labels and breadcrumbs.

* **Bug Fixes**
  * DAG run links now respect configured base paths when opened in new tabs.
  * Improved pagination behavior when data updates, preventing stale or duplicate results.
  * Corrected timezone handling for custom date-range searches.


yohamta0:
All six CodeRabbit review threads have been addressed and resolved. Note: the bot reviewed only the initial commit (4f828739); most findings were fixed by later commits in this branch. Summary per thread:

1. **Paginated rows lost when the head page refreshes** (artifactListPagination.ts:116) — Fixed in `396a4d2`: the silent `refreshInterval` was removed entirely. The head page only changes through `refresh()` (resets continuation pages first) or a query-key change (also resets), and a head-cursor re-anchor effect drops loaded older pages whenever the head moves, so a stale page can never mix into the merged list or leave an exhausted cursor behind. Covered by `artifactListPagination.test.ts`.

2. **Default date preset not applied before the first request** (index.tsx:92) — Superseded by product decision `84af5df`: the page now defaults to a new "All time" preset with no date bound, and the select displays "All time", so the first request matches the UI. The original mismatch (Today shown while unbounded) no longer exists.

3. **Configured timezone wall-clock lost in date conversion** (index.tsx:118) — Fixed in `396a4d2` with exactly the proposed `.utcOffset(config.tzOffsetInSec / 60, true)`, plus regression tests ("interprets custom dates in the configured timezone") on the artifacts and Executions pages.

4. **Initial request error shown as empty state** (index.tsx:212) — Fixed in `396a4d2`: `error` is destructured and rendered as a destructive banner instead of "No artifacts found", with a regression test.

5. **Row key collision on dagRunId** (index.tsx:502) — Fixed by the two-pane redesign `a278ab1`: runs are keyed by `runKey(item)` = name + dagRunId.

6. **Long table text truncation** (index.tsx:520) — The reviewed table was replaced by a file tree in `a278ab1`; wrapping long paths inside the fixed-width tree would break the layout. The tree truncates names and now exposes the full path via title tooltips (`77b7429`).

CI: Frontend build green, E2E suites running.

yohamta0:
@coderabbitai review

yohamta0:
Second CodeRabbit pass (10:54Z) raised 4 findings — all addressed in `08b72fd`:

1. **Keyboard order vs visible tree order** (artifacts/index.tsx:395) — Fixed: `fileRefs` is now derived by walking the rendered tree (`filesToTreeNodes` → `collectFileLeaves`) and mapping synthetic paths back to real ones, so j/k/arrows move in the exact visible order even when a run's files span multiple directories. Regression test: "navigates files in the rendered tree order across directories".

2. **Arrow keys while browsing the preset select / mode toggle** (artifacts/index.tsx:465) — Fixed: the key handler now ignores events originating inside the filter bar (`filterBarRef.contains(target)`) and any Radix select popper (`[role="listbox"]`). Regression test: "ignores navigation keys from the filter bar".

3. **Tooltip exposed the synthetic internal path** (artifacts/index.tsx:824) — Fixed: `TreeNode` receives the run's tree root prefix and titles render the real artifact path (`out/report.md` instead of `@run:…/out/report.md`). Regression test: "shows the real artifact path in file tooltips".

4. **Import order on the Executions page** (dag-runs/index.tsx:4) — Fixed: `@/lib/dayjs` moved after the external imports.

yohamta0:
@coderabbitai review

yohamta0:
Third CodeRabbit pass (11:11Z) finding addressed in `c73ee44`:

- **Traversal included hidden rows** (artifacts/index.tsx:409) — Fixed: `fileRefs` now only includes leaves whose run root and every directory ancestor are in `expandedPaths`, with the expansion set in the memo dependencies — j/k/arrows move strictly through visible rows and clamp at the edges. Regression test: "does not navigate into collapsed runs until they are expanded".

yohamta0:
@coderabbitai review

PR #2793, opened 2026-09-16
feat(ui): save artifact views for the Artifacts page

### LINKED ISSUE #2792: feat: add saved views to the Artifacts page

The Executions page (`/dag-runs`) and the Workflows page (`/dags`) both let a team save the
current filter set as a shared view, star it into the left sidebar, and mark one as the
workspace default. The Artifacts page (`/artifacts`) has none of that.

Its filter state — DAG name, file name, date range — lives entirely in React state. There
are no URL query parameters, so a filtered Artifacts page cannot be linked, bookmarked,
restored with the back button, or shared with a teammate, and there is nothing for the
sidebar to pin. Operators re-type the same DAG name, file glob, and date range on every
visit.

A fourth saved-view type, `artifact`, reusing the existing `/views` API, the shared
`ViewSelector` component, and the existing sidebar pinning.

**UI** — a `View` dropdown in the Artifacts page title row, identical to the one on the
Executions page: select a shared view, "Save current filters as view…", "Manage views…",
an `Edited` badge with `Update "<name>"` / `Reset changes`, star-to-sidebar, and a shared
default view per workspace scope. Pinned views render as top-level sidebar entries linking
to `/artifacts?view=<id>`.

**URL state** — the Artifacts page gains query-parameter-backed filters, matching the
Executions page:

```
/artifacts?view=<id>
/artifacts?view=<id>&name=etl&fileName=*.csv
/artifacts?name=etl&dateMode=custom&fromDate=...&toDate=...
```

Explicit URL filters take precedence over the selected view, so a deep link stays exact.

**API** — two additive extensions to the existing `/views` model:

- `ViewSpec.fileName` / `View.fileName` (string, `maxLength: 255`) — the Artifacts page's
  file-name/glob filter has no equivalent field today.
- An `all` member on `RunDatePreset`, meaning "no date bound". This is the Artifacts
  page's default. It is accepted only for `artifact` views; run views continue to reject
  it server-side.

```json
POST /views
{
  "name": "Nightly ETL reports",
  "type": "artifact",
  "workspaceScope": "all",
  "dagName": "nightly-etl",
  "fileName": "*.csv",
  "dateMode": "preset",
  "datePreset": "last7days",
  "intervalDays": 1,
  "pinned": true,
  "isDefault": false
}
```

## Describe alternatives you've considered

- **A dedicated `ArtifactDatePreset` enum** instead of adding `all` to `RunDatePreset`.
  Rejected: it duplicates six enum values across OpenAPI, Go, and TypeScript to express
  one extra member, and forces a parallel `artifactDatePreset` property plus a second
  mapping branch in `viewFromSpec`/`toViewResponse`. A shared enum with a type-specific
  validator (`ValidArtifactDatePreset`) gives the same safety for a fraction of the
  surface.
- **Dropping the file-name filter from saved views**, reusing only existing `ViewSpec`
  fields. Rejected: the file-name/glob filter is the Artifacts page's most distinctive
  control, and a view that silently clears it when re-selected is worse than no view.
- **localStorage-only bookmarks**, as a purely client-side feature. Rejected: saved views
  in Dagu are deliberately shared and server-persisted, so a local-only variant would be
  neither shareable nor pinnable for the rest of the team, and would diverge from the two
  pages that already have this.

## Use case

A platform team runs a nightly ETL workflow that writes CSV reports as artifacts. The
on-call engineer currently opens `/artifacts`, types the DAG name, types `*.csv`, and
picks "Last 7 days" — every single morning, and again for anyone they hand off to.

With this change they save that filter set once as "Nightly ETL reports", star it, and it
appears in every teammate's sidebar as a one-click entry. A link to
`/artifacts?view=<id>` can be pasted into an incident channel and lands everyone on the
same filtered list.

## Additional context

This mirrors the run views added in #2774 and the workflow views before them. The
implementation reuses `internal/view`, `internal/persis/store/view.go`, the `/views`
endpoints in `internal/service/frontend/api/v1/views.go`, and
`ui/src/features/views/ViewSelector.tsx` — `sameDefaultScope` already keys on the view
type, so an artifact default cannot clear a run default.

Backward compatibility: both API additions are optional/additive. Existing kanban,
workflow, and run views are unaffected, and a stored view without `file_name` reads back
as an empty filter.

yohamta0:
## Summary

Adds saved views ("bookmarks") to the Artifacts page, matching the Executions and
Workflows pages: save the current filters as a shared view, star it into the left
sidebar, and set a shared default per workspace scope.

The Artifacts page kept every filter in component state, so a filtered list could not be
linked, restored, or pinned. Its DAG name, file name, and date filters are now backed by
URL query parameters and session state, which is what makes `/artifacts?view=<id>` work
from the sidebar.

## Changes

- Add a `TypeArtifact` saved-view type to `internal/view`, with its own `Normalize`
  defaults and `Validate` branch
- Add `fileName` to `ViewSpec`/`View` (the Artifacts page's file glob has no existing
  field) and an `all` member to `RunDatePreset` meaning "no date bound"
- Keep `ValidRunDatePreset` unchanged so a run view still rejects `all`; artifact views
  use the new `ValidArtifactDatePreset`/`ValidArtifactDateMode`, which also reject the
  `specific` date mode the Artifacts page does not have
- Map the artifact fields through `viewFromSpec`, `toViewResponse`, and `UpdateView`'s
  nil-preserve block, so a partial `PUT` cannot wipe a view's file filter
- Add `ui/src/features/artifacts/lib/artifactViews.ts`, mirroring `runViews.ts`
- Back the Artifacts page filters with URL params (`name`, `fileName`, `dateMode`,
  `preset`, `fromDate`, `toDate`, `view`) and `SearchStateContext`; explicit URL filters
  take precedence over the selected view
- Render the shared `ViewSelector` with a new `artifact` kind, and its strings in all
  three locales
- Render pinned artifact views as sidebar entries and stop the Executions group claiming
  the active state while one is open

## Related Issues

Closes #2792

## Checklist

## Testing

```
go test ./internal/view/... ./internal/service/frontend/api/v1/...   # ok
golangci-lint run ./internal/view/... ./internal/service/frontend/api/v1/...  # 0 issues
cd ui && pnpm test        # 161 files, 918 tests passing
cd ui && pnpm typecheck   # clean
```

New coverage: artifact normalize/validate/storage round trip and a guard that
`ValidRunDatePreset` still rejects `all`; artifact view create plus a partial-update
regression test for `fileName`; seven Artifacts page cases (default view, `?view=<id>`,
URL override, `?view=all`, fresh preset dates, save-and-apply, edited badge); plus
`ViewSelector` and sidebar cases.

Also verified against a running server: created an artifact view over two artifact-
producing DAG-runs and confirmed in a browser that the default view applies on a bare
`/artifacts`, the sidebar entry links to `/artifacts?view=<id>`, an explicit `?name=`
overrides the view, "All artifacts" clears it, the Edited badge appears, and the
Executions group does not highlight. The API rejects a run view with `datePreset: all`
and an artifact view with `dateMode: specific`, both with 400.

## Notes for reviewers

`api/v1/api.gen.go` and `ui/src/api/v1/schema.ts` are regenerated output from `make api`
and `pnpm gen:api`; they are committed alongside the `api.yaml` change that caused them.

---
## Summary by cubic
Adds saved artifact views to the Artifacts page and closes #2792. Previously, artifact filters lived only in component state; they now support shared URLs, session restoration, starred sidebar links, and workspace defaults.

- Adds an `artifact` view type with DAG name, file name, workspace, and date filters; `all` removes date bounds and remains invalid for run views.
- URL filters override selected views, shared links ignore the reader's session filters, and explicit custom ranges treat missing bounds as empty.
- Emptying a date field now clears its bound instead of being ignored, and omitted fields survive partial updates.
- Preset ranges are recalculated on every restore, now for run views too, so sessions kept across date boundaries query the current window.

<sup>Written for commit 6683448d0babfca6e6045662014c7ebaddddedc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added saved views for the Artifacts page.
  - Save, update, pin, set defaults for, select, and delete artifact views.
  - Filter artifacts by DAG name, file name, workspace, and date range.
  - Added “All artifacts” view with an unbounded date option.
  - Persist and restore artifact view selections through URLs and workspace state.
  - Added pinned artifact views to sidebar navigation.
  - Added localized labels and messages for artifact views.
- **Bug Fixes**
  - Preserved existing artifact view settings when omitted during updates.


yohamta0:
Both review findings reproduced and fixed in 995a66654.

**1. Shared links inherited the reader's stored filters** — confirmed. `/artifacts?name=reporter&dateMode=preset&preset=all` returned `fileName=**/*.csv` from the reader's own session. The restore effect merged `stored` into `base` unconditionally, so any filter the URL did not name fell back to whatever that reader happened to have.

Fixed: `base` starts from defaults whenever the URL carries filters, and keeps the session fallback only for a bare URL.

This needed a second change to be safe. `handleDatePresetChange` rewrites the URL but only wrote the date params, so under the stricter restore a name/file filter that had come from session state would be silently dropped on the next pass. It now carries the applied text filters too. There is a regression test for exactly that combination — it passes with the old `base` merge and fails with the new one unless the handler is also fixed.

For the record, this logic is identical in `pages/dag-runs/index.tsx:656`, so the Executions page has the same shareability gap. I left it alone as out of scope for this PR, but it is worth a follow-up.

**2. Cleared end date did not persist** — confirmed, and the cause runs one layer deeper than the handler.

`buildArtifactViewSpec` sent `toDate: undefined`, which `JSON.stringify` drops, and `UpdateView`'s nil-preserve branch then restored the old bound. The nil-preserve itself is the intended patch-merge contract, so the fix is client-side: both dates are now sent explicitly, with `''` for cleared.

Verified end to end against a running server. Saved a custom-range view with `2026-09-01T00:00 → 2026-09-30T23:59`, switched to the Last 7 days preset (no end bound) and back to Custom, then hit Update. `toDate` is now cleared server-side; previously it came back as `2026-09-30T23:59`.

Worth flagging: typing to clear a date still cannot reach this path at all. `DateRangePicker.handleInputChange` only calls `onChange(parsed)` when `parsed` is truthy, so emptying the visible field never propagates. That component is shared with the Executions page, so I did not change it here — happy to fix it separately if you want clearing by typing to work.

**Tests added**, each verified to fail without its fix:
- `artifactViews.test.ts` — the spec builder sends `''` rather than omitting a cleared date
- `pages/artifacts` — a filter-carrying URL ignores stored session state; a bare URL still restores it; changing only the preset keeps restored filters
- `views_test.go` — an explicit empty date clears the bound server-side

`ui/src/test/setup.ts` gained Pointer Capture stubs alongside the existing `ResizeObserver`/`scrollIntoView` ones; jsdom does not implement them and Radix's Select needs them to open.

Full suite green: 926 UI tests, Go package tests, typecheck and lint clean.

yohamta0:
Both still reproduced. Fixed in 8b41c2411.

**1. Stale custom range** — confirmed. The date block only read the bounds when at least one date param was present:

```
dateModeParam !== 'preset' && (params.has('fromDate') || params.has('toDate'))
```

So `?view=<id>&name=reporter&dateMode=custom` skipped it entirely and fell through to the view's saved range. My repro came back with `fromDate=1788188400` (2026-09-01), exactly as you described.

It is reachable without hand-writing a URL: pick the All time preset, toggle Custom (both bounds undefined), type a name, press Enter. `handleSearch` writes `dateMode=custom` and deletes both date params.

Fixed: an explicit custom mode owns its range, so an absent bound means no bound. A mode-less URL carrying concrete dates still gets the legacy custom treatment. Empty-string params now normalize to `undefined` so they compare equal to a view with no bounds and do not trip the Edited badge.

**2. Partial PUT clearing dagName and pinned** — confirmed, and it was my inconsistency. The repro logged `dagName=<nil> fileName=0x... pinned=<nil>`: adding the `fileName` nil-preserve in the previous round left two fields from the same filter bar disagreeing about what an omitted field means.

Fixed by completing the patch-merge: `dagName`, `labels`, and `pinned` now survive an omitted field alongside everything else.

Worth calling out explicitly, since `UpdateView` is shared: this changes partial-PUT behavior for kanban, workflow, and run views too. No UI-visible change, because every spec builder sends these fields on every update, and the full views package suite passes unchanged. If you would rather keep that blast radius at zero I can gate it on `TypeArtifact`, but the uniform patch-merge looked like the better contract.

Verified against a running server: a rename-only PUT now keeps `dagName: 'reporter'` and `fromDate: '2026-09-01T00:00'`, and in the browser `?view=<id>&name=reporter&dateMode=custom` leaves both date fields empty while `&fromDate=2026-08-01T00:00` keeps that one bound and no end bound.

Tests added, each verified to fail against the previous commit: two page cases for explicit custom ranges, and `TestViewsAPI_UpdatePreservesOmittedArtifactFilters`. Suite green at 928 UI tests, Go package tests, typecheck and lint clean.

Still open from the last round and not addressed here: `DateRangePicker.handleInputChange` only calls `onChange(parsed)` when `parsed` is truthy, so clearing a date by typing never propagates. It is shared with the Executions page, so I left it alone.

yohamta0:
Follow-up on the last two findings, plus the same bug on the Executions page.

**P2 (stale restored presets) — real, fixed in 554fec992.** Reproduced: with session state holding a 3-day-old concrete range under preset `today`, a bare `/artifacts` queried `fromDate=1789311600` while the control still read "Today"; expected today's start, `1789570800`.

Resolution was gated on `dateModeParam === 'preset'` — the URL naming the mode. A range restored from session state or a saved view skipped it and kept whatever dates were computed when the preset was last picked. It now resolves whenever the restored mode is relative, wherever the filters came from.

**P3 (unchanged custom views show Edited) — not reachable as described.** The stated mechanism is API `''` versus URL-restored `undefined`, but the API never emits `''`: `toViewResponse` uses `ptrOf`, which drops zero values, so an empty bound is omitted and arrives as `undefined`. I tested the realistic case, a view with an absent `toDate`, and there is no Edited badge. It only reproduces against a fabricated response carrying `''`.

The normalization inconsistency behind it was real though — `artifactsFilterSetFromView` used `?? undefined` (passes `''` through) while the URL path used `|| undefined`. Harmless today, a trap if serialization ever changes, so I aligned them and covered it with a test.

**Same stale-preset bug on the Executions page — fixed in 6683448d0.** `pages/dag-runs/index.tsx` gated resolution identically, so `/dag-runs` had it too. I confirmed it with a focused test before touching anything: `fromDate=1789311600` against an expected `1789570800`, exactly as on the Artifacts page.

This is outside the original scope of this PR, which is artifact views — flagging it explicitly so you can judge. A legacy URL carrying concrete dates already resolves to a custom range, so those dates are still honored, and all 26 existing Executions tests pass unchanged. Happy to split it into its own PR if you would rather keep this one scoped.

Three regression tests added, each verified to fail against the previous commit. Suite green at 934 UI tests, typecheck and lint clean.

PR #2809, opened 2026-09-18
fix(runtime): wait for durable status before continuing past a node

### LINKED ISSUE #2817: bug: progress handoff does not make a node's status durable before dependents run
## Describe the bug

A progress send only proves the agent has *received* a node, never that it has *persisted* it.

`Runner.report` publishes a node state change on an unbuffered `progressCh`. The send returns as soon as the agent's progress goroutine picks the node up. The status append and its `fsync` happen afterwards, concurrently with the node's dependents already starting.

`execNode` already relies on the stronger guarantee it never had:

```go
if progressCh != nil && node.Step().SubDAG != nil {
    // Send an additional progress notification after the executor is set up
    // so that SubRuns are persisted to storage before the subDAG starts running.
```

That extra notification exists specifically so a sub-DAG's `SubRuns` reach storage before the child starts. The unbuffered handoff does not provide it.

## To Reproduce

Timing-dependent; there is no deterministic YAML reproduction. The window is:

1. Run a DAG with a step that declares a `subDAG`.
2. `execNode` sends the post-setup progress update carrying the child's `SubRuns`.
3. The agent goroutine receives it. The sender returns immediately and the child run starts.
4. The parent process dies before `Attempt.Write` → `writer.Write` → `file.Sync()` completes.

## Expected behavior

A node's state is durable before execution continues past it. In particular, the parent's attempt record names the child run before that child is allowed to start.

## Actual behavior

The parent's persisted status can omit a child run that has already started, because the write covering it had not been synced when the child was launched.

The gap is not limited to sub-DAGs: every node transition is published the same way, so dependents can begin before the predecessor's state is durable. Sub-DAGs are only where the code already asserted otherwise.

## Environment

- Dagu version: `main` (all versions carrying the current `progressCh` handoff)
- OS: platform-independent
- Installation method: n/a — source-level defect

## Additional context

Found by @RawssemMoussaoui-rtx while reading `execNode`, rather than observed in the field. Filed retroactively so the fix has a tracking ticket.

Secondary defect in the same path: `Attempt.Write` returns `ErrWriteFailed` bare, discarding the underlying encode/flush/sync error, so a caller cannot tell which layer failed.

Fixed by #2809.

RawssemMoussaoui-rtx:
Closes #2817

## Problem

A progress send only proved the agent had *received* a node, not that it had *persisted* it. The unbuffered handoff on `progressCh` returns as soon as the consumer picks the node up; the status append and its `fsync` happen afterwards, concurrently with the node's dependents already starting.

`execNode` already relied on the stronger guarantee it never had:

```go
if progressCh != nil && node.Step().SubDAG != nil {
    // Send an additional progress notification after the executor is set up
    // so that SubRuns are persisted to storage before the subDAG starts running.
```

## Solution

Carry an acknowledgement channel on each progress update. `Runner.report` sends the update and waits for the receiver to report the outcome of persisting the status snapshot covering it, so a node's state is durable before execution continues past it.

- **`ProgressUpdate`** (`internal/runtime/progress.go`) — `{Node, ack chan error}`. `Ack` is a no-op on an update with no ack channel, so an update sent by a path that does not wait can never block, and future senders stay safe by construction.
- **Sender** — `Runner.report` is now the single send helper; every previously raw `progressCh <- node` routes through it. The wait is abandoned once `ctx` is done.
- **Receiver** — the agent's progress goroutine acks with the result of `recordCurrentStatus`, which returns the error from `Attempt.Write` (and therefore from `writer.Write` → `file.Sync()`).

### Behaviour

| Case | Result |
|---|---|
| Local run | Sender waits for the append + `fsync` covering its change |
| Distributed run (`statusPusher != nil`) | Sender released before the push; status goes over the network, not to a local durability boundary |
| Dry run | Sender released at once; nothing is persisted |
| Write fails | Logged; the run outcome is unchanged. The terminal write at the end of `Run` already joins its error into `lastErr` |
| Terminal status still subject to change | Stays suppressed — see below |

### Terminal status is still withheld

`shouldDelayTerminalStatus` holds a terminal status back while a notification, artifact finalization, or scheduler-log streaming is outstanding. That is not cosmetic: `artifactFinalizer.Finalize` can turn a `Succeeded` run into `Failed`, so publishing early means publishing an outcome the final write then has to retract. A withheld write acks `nil` — the status it holds back is written, and synced, by the final `writeStatus`.

### Event handler lock

The acknowledgement travels back through `HandlerNode`, which takes `handlerMu.RLock()`. The handler loop used to hold that same read lock across the send, so the handler nodes are now resolved up front and the lock released before any of them runs.

### `attempt.go`

`ErrWriteFailed` was returned bare, discarding the underlying encode/flush/sync error. It now wraps the cause.

## Testing

- `internal/runtime/runner_internal_test.go` — `report` does not return before the update is acked; an ack error leaves the run outcome alone; a cancelled context releases an abandoned waiter; a nil channel never blocks.
- `internal/runtime/agent/status_pusher_test.go` — a terminal status is withheld while a notification is outstanding, and published once it is not. Confirmed failing against the previous revision of this branch.
- `go test -race ./internal/runtime/... ./internal/persis/file/dagrun/... ./internal/intg/...` — pass.
- `make conformance` — 1135 tests, 5 platform skips, no failures.
- `make check` — 0 issues, including `GOOS=windows`.

## Note on the previous revision

This replaces the monotonic watermark tracker. That design released waiters by a shared high-water mark and, on any write failure, entered a terminal failed state that turned every later wait into an error and overwrote the run's real error. It also bypassed `shouldDelayTerminalStatus` whenever a request was pending — which, inside the consumer loop, is always — so `TestBaseConfig_SMTPRetry` failed on both Linux and Windows.


yohamta0:
Thanks @RawssemMoussaoui-rtx — good catch.

This was never reported by anyone; you found it by noticing that the comment in `execNode` claimed a guarantee the unbuffered handoff never actually provided. Filed retroactively as #2817 so the fix has a tracking ticket, and linked it from the PR body.

yohamta0 (approved):
LGTM, thank you very much for finding the bug and the fix!

PR #2825, opened 2026-09-20
fix(mailer): generate RFC-compliant MIME messages

### LINKED ISSUE #2803: bug: mail.send produces non-RFC-compliant MIME (missing Date/Message-Id/MIME-Version), triggers spam filters
## Describe the bug

`mail.send` builds MIME messages that are missing the `Date`, `Message-Id`, and `MIME-Version` headers, and always wraps the body in `multipart/mixed` with a static, non-random boundary (`==simple-boundary-dagu-mailer`) even when there are no attachments. In our setup this causes legitimate mail to be hard-rejected as spam (`554 5.7.0 spam`) by the receiving mail provider's spam filter (Rspamd), while byte-identical content sent as a properly-formed message (correct `Date`/`Message-Id`/`MIME-Version`, plain `text/html` instead of an attachment-less `multipart/mixed`) is delivered without issue.

## To Reproduce

1. Create a DAG with a step using `action: mail.send`, no `attachments`:
   ```yaml
   - id: send_email
     action: mail.send
     with:
       from: "Daily Digest <sender@example.com>"
       to: recipient@example.com
       subject: "Test"
       message: "<!DOCTYPE html><html>...</html>"
   ```
2. Run the workflow.
3. Inspect the raw headers of the message as received on the wire (the step log only shows the resolved `with:` parameters, not the constructed MIME message, so this requires capturing the message at the receiving/relaying mail server).
4. If the receiving server runs a spam filter sensitive to missing standard headers (e.g. Rspamd), the message is rejected during the SMTP transaction.

## Expected behavior

- `mail.send` sets standard headers on every outgoing message: `Date`, `Message-Id` (e.g. `<random>@<hostname>`), and `MIME-Version: 1.0`.
- When there are no `attachments`, the message is sent as a single `text/plain`/`text/html` part matching the content, not wrapped in `multipart/mixed`. `multipart/mixed` is only used when `attachments` are actually present.
- If a multipart message is used, the boundary string is randomly generated per message rather than a static/shared literal.
- The resulting message is RFC 5322/2045-compliant and doesn't get penalized by standard spam-filter heuristics for missing headers.

## Actual behavior

Raw headers of the message dagu actually sent (captured from a delivery-failure report at the receiving mail provider):

```
To: recipient@example.com
From: Daily Digest <sender@example.com>
Subject: Daily Digest - Thursday, September 17, 2026 - 14 articles
Content-Type: multipart/mixed;
  boundary="==simple-boundary-dagu-mailer"
```

Missing: `Date:`, `Message-Id:`, `MIME-Version:` (the last one required per RFC 2045 whenever a `Content-Type`/MIME structure is used). The body is a single `multipart/mixed` part containing base64-encoded `text/html` — `multipart/mixed` is used even though there's exactly one part and no attachment; normally `text/html` directly (or `multipart/alternative` for a text+HTML combo) would be used.

The receiving mail server (using Rspamd as its spam filter) hard-rejects the message during the SMTP transaction:

```
554 5.7.0 Reject, id=343130-19 - spam
```

We isolated the cause via a series of controlled tests, sending byte-identical HTML content through a manual SMTP session (`swaks`) over the same relay path:

| Test | Result |
|---|---|
| Real digest HTML, real links, security-related keywords (CVE/exploit/vulnerability), via `swaks` with normal `Date`/`Message-Id`/`Content-Type: text/html` | Delivered |
| Same, but with `Message-Id:` header explicitly forced empty | Delivered |
| Exact byte-identical HTML that dagu's own `mail.send` had produced in the same run, sent via `swaks` as a plain `text/html` message with normal `Date`/`Message-Id`/`MIME-Version` | Delivered |
| Same content, sent by dagu's `mail.send` (missing `Date`/`Message-Id`/`MIME-Version`, wrapped in attachment-less `multipart/mixed` with static boundary) | Rejected as spam (554 5.7.0) |

This rules out content, links, keywords, message size, and sender domain as the cause of the rejection — the only remaining variable is how `mail.send` itself constructs the MIME message.

## Environment

- Dagu version: 2.16.6
- OS: Docker container (Linux)
- Go version: unknown
- Installation method: Docker

## DAG configuration

```yaml
- id: send_email
  depends: [email_body]
  action: mail.send
  with:
    from: Daily Digest <sender@example.com>
    to: recipient@example.com
    subject: "${email_body.output.result.subject}"
    message: "${email_body.output.result.html}"
```

## Screenshots

N/A

## Additional context

- No `attachments` were used in this step; there was no reason for the message to be `multipart/mixed` in the first place.
- The boundary string `==simple-boundary-dagu-mailer` appears to be a static literal rather than randomly generated per message, which is itself unusual for a MIME boundary and could independently contribute to spam heuristics.
- Workaround in use: replaced the `mail.send` step with a small custom script that opens a raw SMTP connection and constructs a compliant single-part `text/html` message with proper `Date`/`Message-Id`/`MIME-Version` headers.

lzyruc:
## Summary

- add `Date`, `Message-ID`, and `MIME-Version` headers to outgoing messages
- send messages without readable attachments as single-part `text/html`
- use `multipart/mixed` only when attachments are present
- generate a unique MIME boundary for each multipart message
- wrap base64-encoded content at 76 characters per line
- preserve the existing HTML body behavior

## Tests

- `go test -p 1 -parallel 1 ./internal/cmn/mailer -count=1`
- `go test -p 1 -parallel 1 ./internal/runtime/builtin/mail -count=1`
- `go vet ./internal/cmn/mailer ./internal/runtime/builtin/mail`

Fixes #2803

---
## Summary by cubic
Fixes #2803 by generating RFC-compliant MIME messages so email clients no longer misinterpret outgoing mail. Previously every message used `multipart/mixed` with a fixed boundary and lacked `Date`, `Message-ID`, and `MIME-Version`; now all messages include these headers, messages without readable attachments are single-part `text/html`, multipart messages use unique boundaries, and base64-encoded content is wrapped at 76 characters per line. `Message-ID` domains now derive from the sender address to avoid negative spam-filter scoring, falling back to `dagu.local` when unusable. SMTP sending now reports email composition failures clearly.

<sup>Written for commit f5cdaf301f7c197c4c7b4f19fe2ac793ddc68a54. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2825?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email formatting for messages with attachments, including correct MIME metadata and encoding.
  * Messages without usable attachments are now sent as standard single-part HTML emails.
  * Improved handling of unreadable or empty attachment files.
  * Generated messages now use unique identifiers and multipart boundaries.
  * SMTP sending now reports email composition failures clearly.


yohamta0 (approved):
LGTM, thank you very much for improving the mail feature!

PR #2829, opened 2026-09-21
fix: reject unusable continue_on.output patterns and report out-of-range numbers

### LINKED ISSUE #2820: num: is silently a literal match in continue_on.output
Split out of the review on #2816.

`expected` reaches the pattern matcher from five YAML surfaces. #2816 wired `num:` into four of them:

| Surface | `num:` |
|---|---|
| DAG `preconditions[].expected` | yes |
| step `preconditions[].expected` | yes |
| router route pattern | yes |
| `repeat_policy.expected` | yes (validated at build time as of #2816) |
| `continue_on.output` | **no** |

`continue_on.output` goes through `MatchPatternScanner` (`internal/runtime/node.go:1367`), and `stringutil.MatchPattern` deliberately does not know `num:` — it scans log lines, where an ordering comparison has no meaning. That was the right call.

The problem is the failure mode. A pattern is either `re:`-prefixed or a literal substring, so `continue_on: {output: ["num:>=5"]}` is silently a substring search for the text `num:>=5`. No error, no warning. It just never matches.

`continue_on.output` gets no build-time validation at all today — an uncompilable `re:` regexp there is also a silent runtime skip (`matcher.go:97-102` logs and drops the pattern, and `LogContainsPattern` returns `false` with `err == nil`).

Two things worth deciding together:

1. Reject a `num:`-prefixed `continue_on.output` pattern at build time, since it can never do what the author meant.
2. Validate `re:` patterns there too, for the same reason #2816 started validating them in router routes.

### LINKED ISSUE #2821: Numeric comparison accepts Go float literal forms
Split out of the review on #2816. Cosmetic, but the surface should be pinned.

`num:` parses both the pattern operand and the actual value with `strconv.ParseFloat`, which accepts more than the spec contemplates. Verified:

| Input | Result |
|---|---|
| `1_000.5` | 1000.5, no error — underscore separators |
| `0x1p-2` | 0.25, no error — hex float literal |
| `1e-400` | 0, no error — silently underflows |
| `1e400` | +Inf with `ErrRange` |

Spec 023 just says the operator is "followed by a number", so Go literal syntax leaks in on both sides. Harmless in practice but undocumented and untested.

Separately, `1e400` gets the wrong diagnostic. `parseFiniteFloat` (`internal/cmn/stringutil/numeric.go`) returns early on any `ParseFloat` error and reports `"1e400" is not a number`, when in fact it parsed fine and is out of range. `errors.Is(err, strconv.ErrRange)` distinguishes the two.

Suggested: add tests pinning the accepted forms, one sentence in spec 023 naming Go float syntax, and a distinct out-of-range message.

yohamta0:
Two follow-ups from the review on #2816, both cheap to close before the next tag.

## `continue_on.output` (#2820)

`expected` reaches the pattern matcher from five YAML surfaces. #2816 wired `num:` into four of them. The fifth, `continue_on.output`, scans log lines and knows only `re:` and literal text, so `continue_on: {output: ["num:>=5"]}` was a substring search for the text `num:>=5` — it never matched, and reported nothing. An uncompilable `re:` pattern behaved the same way: logged and dropped at runtime, after the step had already failed.

Both are now rejected when the workflow loads:

```
field 'continue_on.output': pattern "num:>=5" is invalid: numeric comparison is not
supported in log patterns; compare numbers with a precondition or router route, or
use "re:" to match the text literally (value: num:>=5)
```

Log matching stays line-based and gains no numeric operator. A comparison reads a whole value, and `LogContainsPattern` answers yes or no, so the not-a-number case that fails a precondition would have to mean "no match" here — the silent gating `num:` exists to prevent. Supporting it later remains possible and would break nothing, since a rejected workflow never ran. Text beginning with `num:` is still matchable through `re:num:>=5`.

The one call site in `buildStepContinueOn` covers the DAG-level `defaults.continue_on` too, since defaults are merged into the step before the field builders run.

## Out-of-range numbers (#2821)

`num:>=1e400` parses fine and overflows to infinity, but `parseFiniteFloat` discarded `strconv`'s error and reported `"1e400" is not a number`, sending the author after a typo instead of a bound. It now says `"1e400" is out of range`. Both the build-time operand and the runtime actual value get the better message.

The forms Go's float parser accepts beyond plain decimals — `1_000.5`, `0x1p-2`, and `1e-400` underflowing to zero — now have tests pinning them, and spec 023 names the syntax instead of saying only "a number".

## Testing

- `go test ./internal/spec/... ./internal/cmn/stringutil/...`
- `go test ./internal/runtime/ -run 'TestNodeLogContainsPattern|TestEvaluateConditions|Numeric'`
- `make conformance` for `spec023_preconditions` and `spec040_router_route`
- `make fmt`
- Manual `dagu validate` for each of the four cases above

No YAML in `examples/`, `conformance/`, `internal/`, or the docs site has a `continue_on.output` entry with a `re:` or `num:` prefix, so nothing that loads today starts failing. The runtime tolerance for an uncompilable regexp stays as a net for already-persisted runs.

A docs update for `writing-workflows/continue-on.md` is ready in the docs repo and will be pushed once this merges.

Closes #2820
Closes #2821

---
## Summary by cubic
Fixes two follow-ups from the #2816 review: unusable `continue_on.output` patterns now fail at load time, and out-of-range numbers report a useful error.

**Bug Fixes**
- Rejects `num:` patterns in `continue_on.output` at load time instead of matching them as literal text and never reporting anything; uncompilable `re:` patterns are also rejected instead of dropped at runtime.
- Reports values like `1e400` as "out of range" instead of "not a number", so authors look for a bound rather than a typo.
- Updates the JSON schema so the array form of `continue_on.output` documents the numeric restriction too.

**Tests**
- Pins Go float literal forms accepted in numeric comparisons (`1_000.5`, `0x1p-2`, underflow to zero) and covers the DAG-level `defaults.continue_on` path.

<sup>Written for commit 7a2d613d92d7a2253d9f6f3ddc8c96497aac6e3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - `continue_on.output` patterns now provide clearer validation errors for unsupported numeric comparisons and invalid regular expressions.
  - Numeric parsing now distinguishes values that exceed the supported floating-point range from non-numeric input.

- **Enhancements**
  - Numeric comparisons support additional Go floating-point formats, including digit separators and hexadecimal exponents.
  - Log output patterns continue to support literal and regular-expression matching.

- **Documentation**
  - Updated guidance clarifies supported numeric formats and range limitations.


yohamta0:
Addressed the schema finding: the array-item form of `continue_on.output` carries its own description, and it now states the `num:` restriction alongside the `re:` guidance (7a2d613d92). Verified the file still parses and `internal/cmn/schema` tests pass.

The `defaults.continue_on` block is left alone deliberately — none of its subfields carry descriptions, so adding one only to `output` would be inconsistent there.

PR #2833, opened 2026-09-22
fix: persist foreach item executions as child runs

### LINKED ISSUE #2832: bug: foreach items have no child DAG run history
## Describe the bug

Inline `foreach` items execute successfully, but their body executions are not persisted as child DAG runs. The parent exposes aggregate item results, while the existing child-run API has no saved child executions to inspect. This prevents drill-down into each item's steps, status, and logs, including nested loops.

## To reproduce

Run this workflow and inspect its completed run:

```yaml
type: graph
steps:
  - id: process
    foreach:
      items: [alpha, beta]
      max_concurrent: 2
      steps:
        - id: show
          run: echo '${foreach.item}'
```

```sh
dagu start foreach-inspection.yaml
```

Request `GET /api/v1/dag-runs/foreach-inspection/{runId}/sub-dag-runs` for that execution.

## Expected behavior

Each iteration has a persisted child execution linked to its parent and root run. Existing child-run endpoints can return its recorded definition, step statuses, and separate logs. Nested iterations remain navigable, and completed executions remain inspectable after restart.

## Actual behavior

The parent completes and reports two successful items, but has no `subRuns` references and the child-run list is empty. The foreach executor invokes the runtime directly without recording child attempts.

## Environment

- Dagu: 2.17.0; regression reproduced on `main` at `f501e3fefb2e52f9430d56b8ac7a055ec97ce320`
- macOS, arm64
- Source build

yohamta0:
## Summary

Inline `foreach` items complete without saved child attempts, leaving the child-run API empty. Persist each item through the existing run-state store and publish its child reference so clients can inspect its recorded steps, status, definition, and logs.

## Changes

- Record queued, running, and terminal item states with parent/root links and separate attempt logs.
- Preserve nested item scope, collection errors, cancellation, and repeat history.
- Publish child references during execution through `SubRunProvider`; reuse the existing child-run API.
- Consume the prepared parent attempt once instead of reusing it for children.
- Add runtime, persistence, and API regressions, and document recorded item executions.

## Related Issues

Closes #2832

## Validation

The nested-history regression fails on the base commit: two successful iterations leave zero child references.

- `make fmt` — native and Windows lint passes.
- `make test TEST_TARGET='./internal/runtime ./internal/runtime/builtin/foreach ./internal/persis ./internal/service/frontend/api/v1'` — 1,749 tests pass with race detection.
- `make conformance CONFORMANCE_TEST_TARGET=./conformance/spec018_parallel_foreach BUILD_VERSION=dev` — 24 tests pass.
- The issue's reproduction workflow passes `dagu validate`.

Commands ran with Go 1.27.0 on macOS arm64.

## Checklist

---
## Summary by cubic
Fixes foreach items not persisting as child runs. Previously, inline items executed without saved attempts, leaving the child-run API empty. Now each item is recorded through the run-state store and child references are published so clients can inspect steps, status, definitions, and logs.

- Records queued, running, and terminal states for each item with parent/root links and separate attempt logs.
- Publishes child references during execution via `SubRunProvider`, reusing the existing child-run API.
- Consumes the prepared parent attempt once; child attempts create their own history.
- Preserves nested item scope, collection errors, cancellation, and repeat history.
- Adds regressions for nested loops, cancellation, failed bodies, and API inspection.

<sup>Written for commit 00e6181f5436d6a3107b020f97237979f22eb377. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dagucloud/dagu/pull/2833?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Foreach item executions are now recorded as individual child runs with item details, parent relationships, statuses, and separate logs.
  - Child runs can be inspected through existing run, specification, and step-log views, including nested and repeated foreach executions.
  - Queued, running, completed, failed, and cancelled item states are reflected in run history.

- **Bug Fixes**
  - Prevented prepared run attempts from being incorrectly reused by subsequent child runs.
  - Improved persistence of failure and cancellation details for foreach items.


yohamta0:
Closing without merging.

Review found two blocking defects and one design problem that together outweigh the benefit:

1. **Data loss.** Each item calls `BeginAttempt` with a synthetic child DAG named after the *step*, and `runStateStore.BeginAttempt` runs history retention keyed on `req.DAG.Name`. A foreach step whose name matches an existing DAG prunes that DAG's run history. Reproduced: a workflow with `hist_retention_runs: 1` and a step named `victim` deleted 3 of 4 stored runs of the unrelated DAG `victim`. `main` leaves them untouched.

2. **Regression.** `foreach` under `handler_on:` fails the whole run with `foreach body specification not found for onSuccess`, because the body spec is located by searching the parent YAML's top-level `steps:` only. That works on `main`. The same lookup also breaks for unnamed array-form foreach steps, nested step groups, and `steps`/`handler_on` inherited from base config.

3. **Cost.** 500 items goes from 1.5s / 2 data files to 35.3s / 1002. That is exactly the platform price of a child run — `parallel` with 500 children on `main` costs 34.9s / 1002 — but it removes the reason to choose `foreach` over `parallel` in the first place, which is the distinction spec 018 draws.

Secondary: the child run ID includes the parent attempt ID, so a run-level retry mints new child runs and orphans the previous ones under `sub/`, leaving the `Retry` lookup unreachable.

